PR #9, opened 2023-02-08
SQLite support via a MySQL parser

adamziel:
## Description

Implements SQLite support using a MySQL lexer instead of regular expressions to:

* Greatly improve security
* Solve most WordPress core unit test failures 
* Make the plugin more maintenable

See https://github.com/aaemnnosttv/wp-sqlite-db/issues/55 to learn more about the problems with the current implementation.

## Status:

* WordPress works with no failing queries and escaping issues
* WooCommerce works
* Most unit tests pass: `Tests: 14853, Assertions: 47545, Errors: 15, Failures: 37, Warnings: 34, Skipped: 60, Risky: 2`. The remaining failures [will have to be fixed in WordPress core](https://github.com/WordPress/sqlite-database-integration/pull/9#issuecomment-1437018625).

## Remaining tasks:

- [x] Fixing the lexer – porting went wrong – `token->keyword` is always null
- [x] Removing the parser dependency
- [x] Fixing the remaining WordPress test failures (by supporting any remaining query types)
- [x] For each fixed failure in ^, add a test case in the translator to enable fast regression spotting
- [ ] Porting Lexer PHPUnit tests
- [ ] API cleanup
- [ ] Add docstrings
- [ ] Merge!

cc @aristath


adamziel:
I pushed basic support for more query types

The remaining ones I spotted:

```
SHOW FULL COLUMNS FROM `testprefix_options`
SHOW INDEX FROM wptests_commentmeta;
SHOW INDEXES FROM wptests_test_truncated_index WHERE Key_name='a_key';
SHOW TABLES LIKE 'wptests\\_dbdelta\\_create\\_test'
SHOW VARIABLES WHERE Variable_name='collation_connection'
START TRANSACTION;
TRUNCATE TABLE testprefix_options
```

The loose code living in the loops would ideally be brought back into the main Translator class.

I'm still trying to figure out an internal API for moving around the token stream – let's keep iterating.

adamziel:
This PR now supports installing, navigating, and logging into WordPress!

The code could use a good cleanup. If this PR was a piece of art, it would be Pollock's painting. I'd like it to be tidy and nicely documented. I don't think of any part of it as final – let's tweak, improve, or completely rewrite it until it's clean and easy to grasp. 

Also, the test suite boots but many tests still fail – I'm sure some queries aren't correctly translated. Also, the integration with $wpdb needs a solid revision, the current one is just a stopgap to get things going.

adamziel:
Test run report:

Tests: 14853, Assertions: 41375, Errors: 201, Failures: 1869, Warnings: 17, Skipped: 81, Risky: 1.

adamziel:
Unit test report after the recent branch updates:

```
Tests: 14853, Assertions: 46841, Errors: 104, Failures: 333, Warnings: 19, Skipped: 67, Risky: 9
```

**YAY! That's ~1500 passed tests more with just a few simple tweaks!** 

Note there's some issue with the Lexer class and I had to manually swap out this:

```php
$tokens     = ( new WP_SQLite_Lexer( $query ) )->list->tokens;
```

For this:

```php
$tokens = \PhpMyAdmin\SqlParser\Lexer::getTokens( $query )->tokens;
```

Ideally, this branch would also port the Lexer unit tests from the PHPUnit repo.

adamziel:
I added a `WP_SQLite_Crosscheck_DB` class that runs all the SQL queries against both SQLite and MySQL and looks for any differences in `num_rows`, `errors`, `$wpdb->query()` return value etc. This is a good starting point for getting all the unit tests to pass. For example, right now it stops at the following error:

```
MySQL query:
SELECT * FROM wptests_users WHERE ID = '2' LIMIT 1

SQLite queries:
SELECT * FROM wptests_users WHERE ID = :param0 LIMIT 1

error:
  MySQL: ''
  SQLite: ''

retval:
  MySQL: 0
  SQLite: 1

num_rows:
  MySQL: 0
  SQLite: 1

insert_id:
  MySQL: 3
  SQLite: 2

rows_affected:
  MySQL: 0
  SQLite: 0
```

Ultimately we'll want to have a PHPUnit test suite not just for `WP_SQLite_Translator` but also for the `WP_SQLite` class to ensure common scenarios like selecting, inserting, deleting etc work the same in SQLite as in MySQL.

adamziel:
The discrepancy I reported above means we can only cross-check one test case at a time by using the `--filter` CLI switch :-(

Why?

To recap:

```
insert_id:
  MySQL: 3
  SQLite: 2
```

In MySQL, AUTO_INCREMENT columns don't reuse IDs assigned in rollback transactions
In SQLite, AUTOINCREMENT columns do reuse IDs assigned in rollback transactions

The first significant ROLLBACK will throw off the rest of the test suite.

adamziel:
Getting closer!
```
Tests: 15318, Assertions: 49108, Errors: 40, Failures: 247, Warnings: 62, Skipped: 60, Risky: 4
```

adamziel:
### This is it!

Here's a PHPunit report from the latest WordPress trunk:

```
Tests: 14853, Assertions: 47545, Errors: 15, Failures: 37, Warnings: 34, Skipped: 60, Risky: 2.
```

Fixing the remaining failures will require patching WordPress core.

### A rundown of the failures

#### Tests_DB 

**Tests_DB test private wpdb methods:**

```php
    class WpdbExposedMethodsForTesting extends WP_SQLite_DB {}
    self::$_wpdb = new WpdbExposedMethodsForTesting();
```

I doesn't make a lot of sense when an SQLite `wpdb` alternative is used.

**Tests_DB::test_replace_value_too_long_for_field**
**Tests_DB::test_process_fields_value_too_long_for_field**

Tests for truncation of long strings inserted into varchar fields. However, SQLite only has TEXT fields, no varchar() fields. The text values are not going to be truncated. It doesn't seem like a useful test in SQLite context.

#### Tests_Comment

**test_wp_update_comment** fails because it counts updated rows and founds 1 when it expects 0. This is happening because in SQLite, all the rows matched by an `UPDATE` query are reported as affected, even if the actual data did not change. In MySQL, a row is reported as updated only if the actual data changes:

```
14) Tests_Comment::test_wp_update_comment
Failed asserting that 1 is identical to 0.
```

**test_comment_field_lengths** fails because it relies on MySQL truncation of too long strings. In SQLite, all text data is stored using the `text` datatype which means no truncation occurs:

```
15) Tests_Comment::test_comment_field_lengths
Failed asserting that 65535 is identical to 65536.
```


#### Tests_Comment_WpComment

**Tests_Comment_WpComment::test_get_instance_should_succeed_for_float_that_is_equal_to_post_id** fails because it compares ints and floats which always yields false in SQLite. It's similar to:

```sql
SELECT * FROM wp_posts WHERE id=42.0 
-- The actual ID is an integer 42
```

The failure:

```
13) Tests_Post_wpPost::test_get_instance_should_succeed_for_float_that_is_equal_to_post_id
Attempt to read property "ID" on bool
```

#### Tests_Menu_Walker_Nav_Menu

This failure is unrelated to SQLite changes. The test depends on a global state and won't work without the `--process-isolation` CLI option:

```
27) Tests_Menu_Walker_Nav_Menu::test_start_el_with_empty_attributes with data set #0 ('', '')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'<li id="menu-item-833" class="menu-item-833"><a>Post title 0002501</a>'
+'<li class="menu-item-833"><a>Post title 0002501</a>'
```

The `--process-isolation` flag cannot be used, however, as it triggers the following error:

```
Serialization of 'pdo' is not allowed
```

### Tests_Theme_ThemeDir

**Tests_Theme_ThemeDir::test_broken_themes**

I'm not sure why it broke, but here's the error:

```
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
         'Title' => 'Child and Parent Theme'
         'Description' => 'The theme defines itself as its parent theme. Please check the <code>Template</code> header.'
     )
+    'tt1-blocks' => Array &3 (
+        'Name' => 'tt1-blocks'
+        'Title' => 'tt1-blocks'
+        'Description' => 'Stylesheet is missing.'
+    )
 )
```

wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
```diff
@@ -224,126 +224,99 @@ public function dateformat( $date, $format ) {
 	/**
 	 * Method to extract the month value from the date.
 	 *
-	 * @param string representing the date formatted as 0000-00-00.
+	 * @see https://www.php.net/manual/en/datetime.format.php
```
adamziel, line 227: This shims a MySQL function using SQLite datatype, the argument type is not specific to PHP so I don't think linking to PHP manual here is helpful. I intentionally included database-related information in the method comment and PHP-related information in a second comment inside the function body – this way a developer may learn both about the function inputs and the translation process.

adamziel: Ditto for all other functions in this class. 

Would you rather debug this code?

```php
	/**
	 * Method to extract the year value from the date.
	 *
	 * @see https://www.php.net/manual/en/datetime.format.php
	 *
	 * @return string representing the number of the year.
	 */
	public function year( $field ) {
		return intval( gmdate( 'Y', strtotime( $field ) ) );
	}
```

Or this code?

```php
	/**
	 * Method to extract the year value from the date.
	 *
	 * @param string representing the date formatted as 0000-00-00.
	 *
	 * @return string representing the number of the year.
	 */
	public function year($field)
	{
		/*
		 * From https://www.php.net/manual/en/datetime.format.php:
		 * 
		 * Y - A full numeric representation of a year, 4 digits.
		 */
		return intval(date('Y', strtotime($field)));
	}
```

The latter offers all the assumption about the inputs, translation, and outputs. The former only gives me limited information about the expected output.

wp-includes/sqlite/class-wp-sqlite-pdo-engine.php
```diff
@@ -49,6 +50,11 @@ class WP_SQLite_PDO_Engine extends PDO { // phpcs:ignore
 	 */
 	public $pre_ordered_results = null;
 
+	/**
+	 * Class variable to store the last query.
+	 *
+	 * @var string
+	 */
 	public $last_translation;
```
adamziel, line 59: This one seems unused so we could remove it completely

```suggestion
```

wp-includes/sqlite/class-wp-sqlite-pdo-engine.php
```diff
@@ -169,8 +175,18 @@ class WP_SQLite_PDO_Engine extends PDO { // phpcs:ignore
...
+	 *
+	 * @var WP_SQLite_Translator
+	 */
 	protected $translator;
 
+	/**
+	 * An array of the last found rows.
+	 *
+	 * @var array
+	 */
```
adamziel, line 189:
```suggestion
	/**
	 * The number of rows found by the last SELECT query.
	 *
	 * @var int
	 */
```

wp-includes/sqlite/class-wp-sqlite-pdo-engine.php
```diff
@@ -6,6 +6,7 @@
  * @since 1.0.0
  */
 
+// Require the translator class.
 require_once dirname( dirname( __DIR__ ) ) . '/lexer-explorations/class-wp-sqlite-translator.php';
```
adamziel: Should we move all the `require_once` statements to `db.php`?

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -1582,15 +1582,15 @@ private function get_primary_keys( $table_name ) {
 	 * @return array
 	 */
 	private function get_keys( $table_name, $only_unique = false ) {
-		$q       = $this->sqlite->query( 'SELECT * FROM pragma_index_list("' . $table_name . '") as l;' );
-		$indices = $q->fetchAll();
+		$query   = $this->sqlite->query( 'SELECT * FROM pragma_index_list("' . $table_name . '") as l;' );
```
adamziel, line 1574: I'm now thinking all instances of concatenating SQL with PHP variables should be turned into prepared statements

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -169,17 +169,6 @@ public function __construct( $pdo, $table_prefix = 'wp_' ) {
...
 	}
 
-	/**
-	 * The regexp function for SQLite.
-	 *
-	 * @param string $pattern  The pattern.
-	 * @param string $subject  The subject.
-	 * @return bool
-	 */
-	public function sqlite_regexp( $pattern, $subject ) {
```
adamziel: good spot!

wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
aristath: Fair point 👍 
Restored the comments.

wp-includes/sqlite/class-wp-sqlite-pdo-engine.php
aristath: It's used in the `WP_SQLite_Crosscheck_DB` class, I assume we'll want to remove that class completely as well?

wp-includes/sqlite/class-wp-sqlite-pdo-engine.php
aristath: Fixed 👍

wp-includes/sqlite/class-wp-sqlite-pdo-engine.php
adamziel: Ah good spot! I'd keep it in the plugin – it's super useful for debugging discrepancies between MySQL and SQLite outputs, I don't think I'd get this far without it.

wp-includes/sqlite/class-wp-sqlite-pdo-engine.php
aristath: Done

wp-includes/sqlite/class-wp-sqlite-db.php
```diff
@@ -249,12 +287,12 @@ public function query( $query ) {
 		}
 
 		$this->last_error = $this->dbh->get_error_message();
-		if ( $this->last_error && ( ! defined( 'WP_INSTALLING' ) || ! WP_INSTALLING ) ) {
+		if ( $this->last_error ) { // && ( ! defined( 'WP_INSTALLING' ) || ! WP_INSTALLING ) ) {
```
adamziel, line 290: I suspect they used these checks to prevent the errors coming up during WP installation from halting the process. There are no errors now so we can remove them:

```suggestion
		if ( $this->last_error ) {
```

wp-includes/sqlite/class-wp-sqlite-db.php
```diff
@@ -249,12 +287,12 @@ public function query( $query ) {
...
 			$this->print_error( $this->last_error );
 			return false;
 		}
 
-		if ( preg_match( '/^\\s*(create|alter|truncate|drop|optimize)\\s*/i', $query ) ) {
+		if ( preg_match( '/^\\s*(set|create|alter|truncate|drop|optimize)\\s*/i', $query ) ) {
```
adamziel, line 295: Ideally these checks would be handled in the engine class – it already sets all this data based on a query type.

wp-includes/sqlite/class-wp-sqlite-pdo-engine.php
```diff
@@ -385,97 +378,148 @@ private function prepare_directory() {
...
+					try {
+						$stmt        = $this->pdo->prepare( $query->sql );
+						$last_retval = $stmt->execute( $query->params );
+					} catch ( PDOException $error ) {
+						if ( $error->getCode() !== self::SQLITE_BUSY ) {
+							throw $error;
+						}
+					}
+				} while ( $error );
+			}
```
adamziel, line 435: This entire block should be inside of a transaction. There might be one active already, though, so `beginTransaction()` and friends should support nested transactions via `SAVEPOINT` and `RELEASE`

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -0,0 +1,2283 @@
...
+
+		// Skip over "DUPLICATE", "KEY", and "UPDATE".
+		$this->rewriter->skip();
+		$this->rewriter->skip();
+		$this->rewriter->skip();
+
+		// Add the CONFLICT keyword.
+		$this->rewriter->add( new WP_SQLite_Token( 'CONFLICT', WP_SQLite_Token::TYPE_KEYWORD ) );
+
+		// Add "( <primary key> ) DO UPDATE SET ".
```
adamziel, line 1524:
```suggestion
		// Add "( <columns list> ) DO UPDATE SET ".
```

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -0,0 +1,2283 @@
...
+		$this->rewriter->consume();
+		$subject = strtolower( $this->rewriter->consume()->token );
+		if ( 'table' !== $subject ) {
+			throw new Exception( 'Unknown subject: ' . $subject );
+		}
+
+		$table_name = $this->normalize_column_name( $this->rewriter->consume()->token );
+		$queries    = array();
+		do {
+			// This loop may be executed multiple times if there are multiple operations in the ALTER query.
```
adamziel, line 1617: In core, all multiline comments need to use the `/*` notation

Issue #16, opened 2023-03-01
Flush the object cache when switching databases.


OllieJones:
It's necessary to call wp_cache_flush() when switching databases -- when creating or deleting wp-content/db.php .

(With this change, it seems this plugin and the SQLite Object Cache plugin are compatible.)

PR #17, opened 2023-03-01
#16 Flush the object cache when switching databases.

OllieJones:
When activating or deactivating the plugin, it's necessary to flush the object cache as late as possible before switching databases.

This change uses PHP_INT_MAX for hook priorities to make sure the hooked functions are called very late.

### DIFF
--- a/deactivate.php
+++ b/deactivate.php
@@ -22,7 +22,12 @@ function sqlite_plugin_remove_db_file() {
 
 	// Init the filesystem if needed, then delete custom drop-in.
 	if ( $wp_filesystem || WP_Filesystem() ) {
+		// Flush any persistent cache.
+		wp_cache_flush();
+		// Delete the drop-in.
 		$wp_filesystem->delete( WP_CONTENT_DIR . '/db.php' );
+		// Flush the cache again to mitigate a possible race condition.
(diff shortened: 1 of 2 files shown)


deactivate.php
```diff
@@ -47,12 +52,17 @@ function() {
 				$value = maybe_unserialize( $row->option_value );
 				if ( is_array( $value ) ) {
 					$value_flipped = array_flip( $value );
-					unset( $value_flipped[ str_replace( WP_PLUGIN_DIR . '/', '', SQLITE_MAIN_FILE ) ] );
+					$items         = array_reverse( explode( DIRECTORY_SEPARATOR, SQLITE_MAIN_FILE ) );
+					$item          = $items[1] . DIRECTORY_SEPARATOR . $items[0];
+					unset( $value_flipped[ $item ] );
```
aristath, line 57: Why this change? 🤔

OllieJones: This works correctly in environments where plugins are symlinked into `wp-content/plugins`. (And, of course, in environments where they're simply installed there.)

aristath: Ah OK! Makes sense now. Thank you for the insights! 👍

Issue #29, opened 2023-03-27
Don't use {SQLITE_IMPLEMENTATION_FOLDER_PATH}


adamziel:
Right now db.copy requires a `str_replace` to swap `{SQLITE_IMPLEMENTATION_FOLDER_PATH}` with the right path. Could it just use `WP_CONTENT` etc and not require any string replacements? That way it would only need to be copied over to `wp-content`.

cc @aristath

aristath:
If we use `WP_CONTENT`, then we'll need to hardcode the foldername of the plugin. In many cases that won't be the right path... For example I currently have 3 versions of the plugin installed on my localhost with different folder-names, and the `db.php` file correctly references the one I activated. There's also a lot of people who use symlinks, mu-plugins etc.
The `SQLITE_IMPLEMENTATION_FOLDER_PATH` replacement was the simplest method I could think of at the time to ensure we cover all scenarios.
I'm not at all against simplifying that implementation, but we should take into account some non-standard scenarios as well 😉

adamziel:
> I currently have 3 versions of the plugin installed 

Oh, that's interesting! What's your use-case?

aristath:
I have 1 on trunk, the 2nd one is for experiments & performance testing (it's easier in some cases 'cause I have changes both in trunk and the "experiments" one which prohibit just switching branches), and the 3rd one is production from w.org 😄

adamziel:
That seems pretty unique to you as `sqlite-database-integration` contributor, no? In which case you could use a custom `db.php` file. I don't expect plugin users to do that.

The reason I even opened this issue is that installing the `db.php` in WordPress Playground and any related work could be as simple as `cp db.php wp-content`, but it's not. It needs reasoning about paths and string replacements – it adds steps and makes the process harder to explain.

aristath:
> That seems pretty unique to you as `sqlite-database-integration` contributor, no? In which case you could use a custom `db.php` file. I don't expect plugin users to have the same setup, do you?

Definitely! But what if they add the plugin in `mu-plugins`, or use symlinks? Specifically using symlinks is a very common scenario in a lot of workflows I've seen people use.

adamziel:
What makes this plugin different from all the other plugins that are not worried about symlinks?

PR #30, opened 2023-03-31
Allow directly copying the db.php file

aristath:
See details in #29

### DIFF
--- a/db.copy
+++ b/db.copy
@@ -13,17 +13,15 @@
 
 define( 'SQLITE_DB_DROPIN_VERSION', '1.8.0' );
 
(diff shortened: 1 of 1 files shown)


aristath:
Hmmm actually this won't work... `WP_PLUGIN_DIR` is not defined when `db.php` gets loaded in Core 🤔

adamziel:
@aristath yes, I just ran into that the other day. Perhaps we could use `__DIR__` instead?

aristath:
@adamziel I just pushed an update if you ant to try this 👍

adamziel (approved):
This is awesome and worked for me, thank you @aristath !

Issue #37, opened 2023-05-19
Critical error on site health screen


mwt:
I get a critical error on the site health screen:

```
2023/05/19 04:09:39 [error] 273564#273564: *3 FastCGI sent in stderr: "PHP message: PHP Fatal
error:  Uncaught TypeError: property_exists():
Argument #1 ($object_or_class) must be of type
object|string, array given in ./html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php:2427
Stack trace:
#0 ./html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(2427): property_exists()
#1 [internal function]: WP_SQLite_Translator->{closure}()
#2 ./html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(2424): array_filter()
#3 ./html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(2384): WP_SQLite_Translator->strip_sqlite_system_tables()
#4 ./html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(1381): WP_SQLite_Translator->get_information_schema_" while reading response header from upstream
```

This is on Debian Testing/Bookworm php8.2-fpm with nginx.

chriskthomas:
My trace is a little longer

```
An error of type E_ERROR was caused in line 2427 of the file /srv/chriskthomas.com/html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php. Error message: Uncaught TypeError: property_exists(): Argument #1 ($object_or_class) must be of type object|string, array given in /srv/chriskthomas.com/html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php:2427
Stack trace:
#0 /srv/chriskthomas.com/html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(2427): property_exists()
#1 [internal function]: WP_SQLite_Translator->{closure}()
#2 /srv/chriskthomas.com/html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(2424): array_filter()
[...]

PR #40, opened 2023-06-12
Update class-wp-sqlite-translator.php

lucazdj:
Fixing message error in diagnostics page

### DIFF
--- a/wp-includes/sqlite/class-wp-sqlite-translator.php
+++ b/wp-includes/sqlite/class-wp-sqlite-translator.php
@@ -2424,7 +2424,7 @@ class WP_SQLite_Translator {
 			array_filter(
 				$tables,
 				function ( $table ) {
-					$table_name = property_exists( $table, 'Name' ) ? $table->Name : $table->table_name; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+					$table_name = is_object( $table ) && property_exists( $table, 'Name' ) ? $table->Name : $table->table_name; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
 					return ! array_key_exists( $table_name, $this->sqlite_system_tables );
 				},
 				ARRAY_FILTER_USE_BOTH


eliot-akira:
I believe this pull request is meant to solve #37, where a fatal error is thrown due to this line.

```diff
-$table_name = property_exists( $table, 'Name' ) ? $table->Name : $table->table_name; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+$table_name = is_object( $table ) && property_exists( $table, 'Name' ) ? $table->Name : $table->table_name; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```

However, adding the check `is_object( $table )` is not enough to solve the issue completely because the error message is:

```
Uncaught TypeError: property_exists():
Argument #1 ($object_or_class) must be of type object|string, array given
```

It means the variable `$table` is an array for some reason. In the changed line, if the variable is not an object, the condition tries to access `$table->table_name` which will throw another type error.

I think the real solution to find out why `$table` is not an object as expected, somewhere earlier in the error stack trace posted in that issue https://github.com/WordPress/sqlite-database-integration/issues/37#issuecomment-1577810048.

Otherwise, as a temporary solution, that `array_filter` callback can return true for that exception at the top of the function.

```php
if ( is_object( $table ) ) return true;
```

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -2424,7 +2424,7 @@ private function strip_sqlite_system_tables( $tables ) {
 			array_filter(
 				$tables,
 				function ( $table ) {
-					$table_name = property_exists( $table, 'Name' ) ? $table->Name : $table->table_name; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+					$table_name = is_object( $table ) && property_exists( $table, 'Name' ) ? $table->Name : $table->table_name; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```
aristath, line 2427: Hmmm I don't think this would actually fix the issue...
if it's not an object, it will try to get `$table->table_name` which will again error.

Perhaps we should just add an additional condition _above_ this line?
Something like this:
```php
if ( ! is_object( $table ) ) {
	return false;
}
```

aristath:
Closing this one as a fix has already been committed for it in https://github.com/WordPress/sqlite-database-integration/commit/0753b7cc519a84d116b0d632b030d29428048e30 a couple of weeks ago 👍

PR #61, opened 2023-12-09
WordPress nightly compat: Bump the db_version to 8.0 

adamziel:
WordPress minimum MySQL version was recently bumped to 5.5, which means this plugin no longer works with the trunk WordPress version. Let's bump the reported db version to 8.0 to keep the version check happy for a longer while.

The version bump was done here:

https://github.com/WordPress/wordpress-develop/commit/e1265a3732eedaf6bf12f87b080e19ebd65a08a6

CleanShot 2023-12-09 at 22 40 34@2x

cc @aristath

### DIFF
--- a/wp-includes/sqlite/class-wp-sqlite-db.php
+++ b/wp-includes/sqlite/class-wp-sqlite-db.php
@@ -349,7 +349,7 @@ class WP_SQLite_DB extends wpdb {
 	 * @see wpdb::db_version()
 	 */
 	public function db_version() {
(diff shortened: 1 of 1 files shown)


adamziel:
I'll go ahead and merge to make the main branch compatible with the latest WordPress. @aristath, does this change look sound? If so, would you please release the new plugin version to WordPress plugin directory?

aristath:
@adamziel done! Released v2.1.3 👍

PR #66, opened 2023-12-28
Allow using `TRUNCATE` without `TABLE`

### LINKED ISSUE #51: Support TRUNCATE without TABLE
From the [docs](https://dev.mysql.com/doc/refman/8.0/en/truncate-table.html): `TRUNCATE [TABLE] tbl_name`

You can do `TRUNCATE wp_posts` but also `TRUNCATE TABLE wp_posts`.

Right now, the former is not supported and results in a database error (`General error: 1 incomplete input`).

aristath:
fixes #51

### DIFF
--- a/wp-includes/sqlite/class-wp-sqlite-translator.php
+++ b/wp-includes/sqlite/class-wp-sqlite-translator.php
@@ -176,7 +176,7 @@ class WP_SQLite_Translator {
 	 * @var array
 	 */
 	private $table_name = array();
-	
+
 	/**
 	 * The type of the executed query (SELECT, INSERT, etc).
 	 *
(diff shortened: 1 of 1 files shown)


wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -1464,7 +1464,9 @@ private function execute_select() {
 	 */
 	private function execute_truncate() {
 		$this->rewriter->skip(); // TRUNCATE.
-		$this->rewriter->skip(); // TABLE.
+		if ( 'TABLE' === strtoupper( $this->rewriter->peek()->value ) ) {
+			$this->rewriter->skip(); // TABLE.
+		}
```
aristath, line 1469: This will now skip only if the value is `TABLE`

PR #71, opened 2024-01-08
Update plugin author in main plugin name

swissspidy:
"The WordPress Team" is used by many other feature projects already


mukeshpanchal27:
The readme: https://github.com/WordPress/sqlite-database-integration/blob/main/readme.txt#L12 still shows `from the WordPress Performance Team.` Should we update that as well?

felixarntz (approved):

mukeshpanchal27 (approved):

PR #77, opened 2024-03-20
Add support for SHOW TABLE STATUS LIKE queries

### LINKED ISSUE #75: Support SHOW TABLE STATUS
Any query that depends on getting data from information_schema is not returning anything; queries that use 'SHOW TABLE STATUS' also broken, and I noticed some normal queries that use group are returning no results.

More context in https://github.com/WordPress/wordpress-playground/issues/663

adamziel:
Adds support for the following queries:

* `SHOW TABLE STATUS FROM dbschema`
* `SHOW TABLE STATUS IN dbschema`
* `SHOW TABLE STATUS IN dbschema LIKE ""`
* `SHOW TABLE STATUS LIKE ""`

Also, brings the keys of the returned record in line with [MySQL spec](https://dev.mysql.com/doc/refman/8.0/en/extended-show.html). The values are mostly fake and supplying the correct ones will require more work.

## Follow-up work

Support `SHOW TABLE STATUS WHERE`

Closes https://github.com/WordPress/sqlite-database-integration/issues/75

cc @brandonpayton

### DIFF
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -195,6 +195,119 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		);
 	}
 
+	public function testShowTableStatusFrom()
+	{
+		// Created in setUp() function
+		$this->assertQuery("DROP TABLE _options");
+		$this->assertQuery("DROP TABLE _dates");
+
+		$this->assertQuery(
+			"CREATE TABLE _tmp_table (
+				ID INTEGER PRIMARY KEY AUTO_INCREMENT NOT NULL,
+				option_name TEXT NOT NULL default '',
+				option_value TEXT NOT NULL default ''
+			);"
+		);
+
+		$this->assertQuery(
+			"SHOW TABLE STATUS FROM 'mydb';"
+		);
+
+		$this->assertCount(
+			1,
+			$this->engine->get_query_results()
+		);
+	}
+
+	public function testShowTableStatusIn()
+	{
+		// Created in setUp() function
+		$this->assertQuery("DROP TABLE _options");
+		$this->assertQuery("DROP TABLE _dates");
+
+		$this->assertQuery(
+			"CREATE TABLE _tmp_table (
+				ID INTEGER PRIMARY KEY AUTO_INCREMENT NOT NULL,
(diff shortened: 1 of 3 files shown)


tests/bootstrap.php
```diff
@@ -7,6 +7,17 @@
 require_once __DIR__ . '/../wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php';
 require_once __DIR__ . '/../wp-includes/sqlite/class-wp-sqlite-translator.php';
 
+/**
+ * Polyfills for WordPress functions
+ */
+
+function do_action( $tag, ...$args ) {
```
adamziel, line 14: Unit tests crashed for me without these, I'm not sure how to run them.

tests/WP_SQLite_Translator_Tests.php
```diff
@@ -195,6 +195,119 @@ public function testShowTablesLike() {
...
+				option_name TEXT NOT NULL default '',
+				option_value TEXT NOT NULL default ''
+			);"
+		);
+
+		$this->assertQuery(
+			"SHOW TABLE STATUS LIKE '_tmp_table1';"
+		);
+		$this->assertCount(
+			1,
```
brandonpayton, line 302: Maybe it would be good to test LIKE's wildcard aspect like:
```suggestion
			"SHOW TABLE STATUS LIKE '_tmp_table%';"
		);
		$this->assertCount(
			2,
```

I worried about wildcard functionality and tried the above test, and it passes.

tests/bootstrap.php
aristath: I think it's safe to keep them in this case...

aristath:
Updated this PR with the latest changes from `main` because there were conflicts that needed to be resolved 👍

adamziel:
`SHOW TABLE STATUS WHERE` is not supported yet, but `SHOW TABLE STATUS LIKE` is which is a win – I'll go ahead and merge.

bgrgicak:
This PR broke PHP 7.1 and 7.2 support in Playground. https://playground.wordpress.net/?php=7.1

bgrgicak:
Locally this crashes the site and it breaks e2e tests.

bgrgicak:
https://github.com/WordPress/sqlite-database-integration/pull/92

Issue #27, opened 2023-03-24
Support UPDATE LIMIT


adamziel:
WordPress cron fails because it cannot execute this query:

`UPDATE wp_posts SET post_password = '971f813efffc0aa3250b', post_modified_gmt = '2023-03-24 15:41:06', post_modified = '2023-03-24 15:41:06' WHERE post_type = 'scheduled-action' AND post_status = 'pending' AND post_password = '' AND post_date_gmt <= '2023-03-24 15:41:06' ORDER BY menu_order ASC, post_date_gmt ASC, ID ASC LIMIT 25`

The problem is with the trailing `LIMIT 25`. In SQLite, it should be split into a SELECT query and an UPDATE query – just like `DELETE from a,b` is currently handled.

MocioF:
I think there is a problem with LIMIT on DELETE queries too:
`DELETE FROM mytable WHERE mydatefield < DATE_SUB( NOW(), INTERVAL 10 DAY) LIMIT 1000` does trigger a sintax error, while
`DELETE FROM mytable WHERE mydatefield < DATE_SUB( NOW(), INTERVAL 10 DAY)`, works

 Error message was: SQLSTATE[HY000]: General error: 1 near "LIMIT": syntax error.

kozer:
I'm currently come up with this issue while I was working at https://github.com/Automattic/dotcom-forge/issues/6479. Can you please assign it to me?

adamziel:
Capturing notes from my conversation with @kozer

## GitHub Comment: SQLite UPDATE with ORDER BY and LIMIT

**Issue:** WooCommerce encounters an error due to SQLite not supporting `UPDATE ... ORDER BY` statements.

**Solution:**

This comment outlines how to potentially modify queries using the `sqlite-database-integration` plugin to address the limitation.

**Background:**

- WordPress cron fails because it cannot execute `UPDATE` queries with a trailing `LIMIT` clause.
- SQLite requires separating the logic into two queries: a `SELECT` to fetch IDs and a subsequent `UPDATE` using those IDs in a `WHERE` clause.

**Implementation Approach:**

1. Leverage the `WP_SQLite_Translator` class within the plugin.
2. Track the table name and `WHERE` clause during query parsing.
3. Upon encountering `LIMIT`:
    - Option A: Construct a `SELECT` query with the same `WHERE` clause and `LIMIT`, followed by an `UPDATE` with `WHERE ID IN (SELECT results)`.
[...]

PR #93, opened 2024-04-11
Support UPDATE LIMIT

kozer:
Fixes https://github.com/WordPress/sqlite-database-integration/issues/27

### Proposal
`UPDATE` statement in SQLite, doesn't support `LIMIT` and `ORDER BY` statements.

The current implementation was throwing an error when this case occurred. ( For example when a user tries to add Woo plugin ).

image

This pr aims to fix that.

### DIFF
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -196,6 +196,45 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		$this->assertEquals( '2001-05-12 10:00:40', $result2[0]->option_value );
 	}
 
+	public function testUpdateWithoutWhereButWithSubSelect() {
+		$this->assertQuery(
+			"INSERT INTO _options (option_name, option_value) VALUES ('User 0000019', 'second');"
+		);
+		$this->assertQuery(
+			"INSERT INTO _dates (option_name, option_value) VALUES ('first', '2003-05-27 10:08:48');"
+		);		
+		$this->assertQuery(
+			"INSERT INTO _dates (option_name, option_value) VALUES ('second', '2003-05-27 10:08:48');"
+		);
+		$return = $this->assertQuery(
+			"UPDATE _dates SET option_value = (SELECT option_value from _options WHERE option_name = 'User 0000019')"
+		);
+		$this->assertSame( 2, $return, 'UPDATE query did not return 2 when two row were changed' );
+		
+		$result1 = $this->engine->query( "SELECT option_value FROM _dates WHERE option_name='first'" );
+		$result2 = $this->engine->query( "SELECT option_value FROM _dates WHERE option_name='second'" );
+		$this->assertEquals( 'second', $result1[0]->option_value );
+		$this->assertEquals( 'second', $result2[0]->option_value );
+	}
+
+	public function testUpdateWithoutWhereButWithLimit() {
+		$this->assertQuery(
+			"INSERT INTO _dates (option_name, option_value) VALUES ('first', '2003-05-27 10:08:48');"
+		);		
+		$this->assertQuery(
+			"INSERT INTO _dates (option_name, option_value) VALUES ('second', '2003-05-27 10:08:48');"
+		);
+		$return = $this->assertQuery(
+			"UPDATE _dates SET option_value = 'second' LIMIT 1"
+		);
+		$this->assertSame( 2, $return, 'UPDATE query did not return 2 when two row were changed' );
+		
+		$result1 = $this->engine->query( "SELECT option_value FROM _dates WHERE option_name='first'" );
+		$result2 = $this->engine->query( "SELECT option_value FROM _dates WHERE option_name='second'" );
+		$this->assertEquals( '2003-05-27 10:08:48', $result1[0]->option_value );
+		$this->assertEquals( 'second', $result2[0]->option_value );
+	}
+
 	public function testCastAsBinary() {
 		$this->assertQuery(
 			// Use a confusing alias to make sure it replaces only the correct token
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -133,6 +133,9 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		$this->assertQuery(
 			"INSERT INTO _dates (option_name, option_value) VALUES ('first', '2003-05-27 00:00:45');"
 		);
+		$this->assertQuery(
+			"INSERT INTO _dates (option_name, option_value) VALUES ('second', '2003-05-28 00:00:45');"
+		);
 
 		$this->assertQuery(
 			"UPDATE _dates SET option_value = '2001-05-27 10:08:48' WHERE option_name = 'first' ORDER BY option_name LIMIT 1;"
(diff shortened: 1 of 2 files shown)


tests/WP_SQLite_Translator_Tests.php
```diff
@@ -129,6 +129,36 @@ public function testInsertDateNow() {
 		$this->assertEquals( gmdate( 'Y' ), $results[0]->y );
 	}
 
+	public function testUpdateWithLimit() {
+		$this->assertQuery(
+			"INSERT INTO _dates (option_name, option_value) VALUES ('first', '2003-05-27 00:00:45');"
```
adamziel, line 134: Also insert a few values and confirm only the first row was updated.

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -1580,13 +1607,35 @@ private function execute_update() {
...
+
 		$this->rewriter->consume_all();
 
 		$updated_query = $this->rewriter->get_updated_query();
 		$this->execute_sqlite_query( $updated_query, $params );
 		$this->set_result_from_affected_rows();
 	}
 
+	private function prepare_update_for_limit_or_order() {
+			$this->rewriter->add( new WP_SQLite_Token( ' ', WP_SQLite_Token::TYPE_WHITESPACE ) );
```
adamziel, line 1623: I think there's a function to add many tokens at the same time. `add_many` perhaps?

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -1550,14 +1550,41 @@ private function execute_describe() {
...
 				break;
 			}
 
+			if ( $token->value === 'WHERE' && ( $limit || $order_by ) ) {
+				$this->remember_last_reserved_keyword( $token );
+				$this->rewriter->consume();
+				$this->prepare_update_for_limit_or_order();
+			}
+
+			if ( $token->value === ';' && $token->type === WP_SQLite_Token::TYPE_DELIMITER && ( $limit || $order_by ) ) {
```
adamziel, line 1584: Why is this condition needed? Query processing should never advance past the `;` delimiter

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -1550,14 +1550,41 @@ private function execute_describe() {
 	 */
```
adamziel, line 1550: Let's add this for completeness:

```suggestion
	 * Supported syntax:
	 *
	 * UPDATE [LOW_PRIORITY] [IGNORE] table_reference
	 *    SET assignment_list
	 *    [WHERE where_condition]
	 *    [ORDER BY ...]
	 *    [LIMIT row_count]
	 *
	 * @see https://dev.mysql.com/doc/refman/8.0/en/update.html
	 */
```

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -1550,14 +1550,41 @@ private function execute_describe() {
...
 	private function execute_update() {
 		$this->rewriter->consume(); // Update.
-
-		$params = array();
+		$limit    = $this->rewriter->peek(
```
adamziel, line 1553: Technically this will also kick in here:

```sql
UPDATE table SET value = 1 WHERE value in (SELECT id from another_table);
```

This isn't a big issue, but since getting it exactly right would require a complex subquery analyzer, why don't we ditch the entire "add a custom SELECT conditionally" and just always transform the query to its `WHERE rowid in` form?

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -1550,14 +1550,41 @@ private function execute_describe() {
...
+			)
+		);
+		$params   = array();
 		while ( true ) {
 			$token = $this->rewriter->peek();
 			if ( ! $token ) {
 				break;
 			}
 
+			if ( $token->value === 'WHERE' && ( $limit || $order_by ) ) {
```
adamziel, line 1578: Let's add a block comment to explain how this is needed to support LIMIT and ORDER BY clauses.

adamziel:
Thank you @kozer! I left a few minor comments but this looks good otherwise. It would also be lovely to add the same support for the DELETE queries if you wouldn't mind? The `LIMIT` in those also keeps coming up.

wp-includes/sqlite/class-wp-sqlite-translator.php
kozer: I didn't add it initially, but while I was writing the tests made me to add this statement.
What was happening is the following:
1. We have the query `SELECT * FROM test limit 10;`
2. While we are in the loop we add the where, so the above becomes: `SELECT * FROM test WHERE rowid IN ( SELECT rowid LIMIT 10;
3. After the loop I add the closing parenthesis, so the above becomes:
4. SELECT * FROM test WHERE rowid IN ( SELECT rowid LIMIT 10;)

Which is wrong. This is why I skip the `;`.

kozer:
> It would also be lovely to add the same support for the DELETE queries if you wouldn't mind? The LIMIT in those also keeps coming up.

Can we proceed with this one, as we consider it a blocker for Studio release, and open a new ticket for the DELETE statement?
Feel free to assign it to me.
Thank you @adamziel !!

wp-includes/sqlite/class-wp-sqlite-translator.php
kozer: In my mind, this sounds more complex than adding the SELECT statement conditionally, only to those two cases we now that they are not supported. Can you elaborate more?
Also, is it ok to consider this an enhancement and create another issue for that?

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -1547,17 +1547,64 @@ private function execute_describe() {
...
+			 */
+			if ( $token->value === 'WHERE' && ( $limit || $order_by ) ) {
+				$this->remember_last_reserved_keyword( $token );
+				$this->rewriter->consume();
+				$this->prepare_update_for_limit_or_order();
+			}
+			/*
+			* In case we rewrite the query, we need to skip the semicolon.
+			* This is because the semicolon becomes part of the nested SELECT statement, and it breaks the query.
+			*/
```
mukeshpanchal27, line 1603:
```suggestion
			/*
			 * In case we rewrite the query, we need to skip the semicolon.
			 * This is because the semicolon becomes part of the nested SELECT statement, and it breaks the query.
			 */
```

wp-includes/sqlite/class-wp-sqlite-translator.php
adamziel: Aha, gotcha! What if the query is just `SELECT * FROM test limit 10` without the final `;`, though?

adamziel: Maybe `;` should not be consumed and just `break` from the loop, and then you would add the closing `)` after the loop?

Also remember to make sure `UPDATE` query without `WHERE` keep working.

kozer: > Also remember to make sure UPDATE queries without WHERE keep working.
I added some extra tests to ensure that.
I also added some nested select statements, to test if it executes the query as expected.
Finally, I break when I encounter a `;`.

adamziel:
@kozer nice, getting closer! I went ahead and pushed two failing test cases – let's make sure they work

adamziel:
I had an idea how to fix those two test cases so I went ahead and did that

PR #94, opened 2024-04-11
Fix default null values and NULL value update edge case

bgrgicak:
This PR allows SQL to set `DEFAULT NULL` which currently isn't supported because the code is built as `DEFAULT ` (empty value instead of _NULL_).

It also adds support for updating a field to `NULL` that is `NOT NULL` and doesn't have a `DEFAULT` value.
The result of the update should be an empty value (_''_ for text and _0_ for numbers).

## Testing instructions

- confirm that all tests pass

### DIFF
--- a/wp-includes/sqlite/class-wp-sqlite-translator.php
+++ b/wp-includes/sqlite/class-wp-sqlite-translator.php
@@ -1128,10 +1128,12 @@ class WP_SQLite_Translator {
 		 * This mode allows the use of `NULL` when NOT NULL is set on a column that falls back to DEFAULT.
 		 * SQLite does not support this behavior, so we need to add the `ON CONFLICT REPLACE` clause to the column definition.
 		 */
-		if (null !== $field->default && $field->not_null) {
+		if ($field->not_null) {
 			$definition .= ' ON CONFLICT REPLACE';
 		}
-		if ( null !== $field->default ) {
+		if (null === $field->default) {
+			$definition .= ' DEFAULT NULL';
+		} else if (false !== $field->default) {
 			$definition .= ' DEFAULT ' . $field->default;
 		}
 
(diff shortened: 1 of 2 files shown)


wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -1128,11 +1128,19 @@ private function make_sqlite_field_definition( $field ) {
...
+		if (null === $field->default) {
+			$definition .= ' DEFAULT NULL';
+		} else if (false !== $field->default) {
 			$definition .= ' DEFAULT ' . $field->default;
+		} else if ($field->not_null) {
+			if ('text' === $field->sqlite_data_type) {
+				$definition .= ' DEFAULT \'\'';
+			} else {
+				$definition .= ' DEFAULT 0';
+			}
```
bgrgicak, line 1143: @adamziel what do you think about this solution? 
Instead of rewriting update queries with COALESCE we could always set a default.

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -1128,11 +1128,19 @@ private function make_sqlite_field_definition( $field ) {
...
-		if ( null !== $field->default ) {
+		if (null === $field->default) {
+			$definition .= ' DEFAULT NULL';
+		} else if (false !== $field->default) {
 			$definition .= ' DEFAULT ' . $field->default;
+		} else if ($field->not_null) {
+			if ('text' === $field->sqlite_data_type) {
+				$definition .= ' DEFAULT \'\'';
+			} else {
+				$definition .= ' DEFAULT 0';
```
bgrgicak, line 1142: I'm not sure if we should include this.

wp-includes/sqlite/class-wp-sqlite-translator.php
adamziel: This will allow both inserting NULL values and updating existing rows to NULL values. MySQL rejects inserting NULL values and only allows updating. I think it's fine for now, let's just create a new issue to track rejecting inserts in the future.

bgrgicak: https://github.com/WordPress/sqlite-database-integration/issues/97

bgrgicak:
@aristath @adamziel could we release a new version?

aristath:
Released v2.1.9 with this patch 👍

PR #102, opened 2024-04-21
feat: support SQLite WAL mode

soulteary:
Hello maintainer 👋

This plug-in project is very good. Although SQLite is suitable for small websites, for small websites with intensive writing, perhaps turning on WAL can further improve the user experience.

And it has almost no side effects.

### DIFF
--- a/wp-includes/sqlite/class-wp-sqlite-translator.php
+++ b/wp-includes/sqlite/class-wp-sqlite-translator.php
@@ -420,6 +420,11 @@ class WP_SQLite_Translator {
 			$this->pdo->query( 'PRAGMA foreign_keys = ON' );
 		}
 		$this->pdo->query( 'PRAGMA encoding="UTF-8";' );
+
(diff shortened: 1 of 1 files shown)


adamziel:
Looping in @JanJakes

JanJakes:
This is already supported in both the old and the new driver via `SQLITE_JOURNAL_MODE`.

See:
https://github.com/WordPress/sqlite-database-integration/blob/4cfe524b228577e9d1b682249e92a3da4505eee0/wp-includes/sqlite/class-wp-sqlite-translator.php#L425-L428

And:
https://github.com/WordPress/sqlite-database-integration/blob/4cfe524b228577e9d1b682249e92a3da4505eee0/wp-includes/sqlite/class-wp-sqlite-db.php#L331

https://github.com/WordPress/sqlite-database-integration/blob/4cfe524b228577e9d1b682249e92a3da4505eee0/wp-includes/sqlite-ast/class-wp-sqlite-connection.php#L96-L101

PR #109, opened 2024-04-30
Tolerate selecting MySQL system variables

### LINKED ISSUE #104: Variable references like @@GLOBAL.sql_mode cause syntax errors
An example query is:
```sql
SELECT @@session.max_allowed_packet
```

Within WordPress Playground this is leading to an error message like:
```
SQLSTATE[HY000]: General error: 1 unrecognized token: "@"
```

Related to https://github.com/WordPress/wordpress-playground/issues/1272

brandonpayton:
The purpose of this PR is to add more support for selecting MySQL system variables. Fixes #104.

Related to:
https://github.com/WordPress/wordpress-playground/issues/1272 - "UpdraftPlus plugins doesn't fully work in Playground."

We'll start with failing tests for the set used by the UpdraftPlus plugin, but the ideal is to support as many MySQL variables as possible so there are no conflicts when using SQLite (and WordPress Playground).

### DIFF
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -2252,8 +2252,8 @@ QUERY
 	 * @dataProvider mysqlVariablesToTest
 	 */
 	public function testSelectVariable( $variable_name ) {
+		// Make sure the query does not error
 		$this->assertQuery( "SELECT $variable_name;" );
-		// TODO: Assert about results once we provide them
 	}
 
 	public static function mysqlVariablesToTest() {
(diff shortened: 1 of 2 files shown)


brandonpayton:
For a first pass, I am currently considering expanding [this existing hack](https://github.com/WordPress/sqlite-database-integration/pull/109/commits/21b0ad2066acf20d927009dbd55abe330fe085dd) to replace queries for these variables with dummy SELECTs.

Doing this does appear to address the original error report here:
https://github.com/WordPress/wordpress-playground/issues/1272

After that, in order to properly support these variables, we should review the list of all variables and decide which make sense to support and which make sense to replace with dummy GETs and SETs.

cc @adamziel

PR #113, opened 2024-05-17
Prevent adding a comma if key isn't in the first position

bgrgicak:
[WordPress Playground observed an issu](https://github.com/WordPress/wordpress-playground/issues/731)e with insert queries that use `ON DUPLICATE KEY` and don't have the KEY as the first value of the insert.

After investigating it it turns out that the condition for adding commas didn't work well in case the KEY value is in a "random" place.

This PR ensures the comma is added for all items except for the last one. 

## Testing instructions

- Ensure tests pass


adamziel:
Good spot!

PR #117, opened 2024-06-05
Bugfixes in SHOW CREATE TABLE and SHOW TABLES

adamziel:
Fixes a series of crashes encountered when

* SHOW TABLES now returns a flat list, not an array of objects – I'm not sure here. It makes `$tables = $wpdb->get_results("SHOW TABLES LIKE '{$prefix}%'", ARRAY_N);` work as expected [here](https://github.com/WordPress/playground-tools/blob/128123e84e25cf0421770f6885c6f26aaf0b05dc/packages/playground/src/playground-db.php#L90-L91), but is the right thing to return, or is this more about correct handling of the fetch mode, like ARRAY_N?
* `SHOW tables` isn't all uppercase
* There's no table tho SHOW CREATE: `SHOW CREATE TABLE _no_such_table;`
* The table identifier passed to SHOW CREATE is quoted:

```
'SHOW CREATETABLE `_tmp_table`;'
```

This PR makes the [Sandbox site plugin](https://wordpress.org/plugins/playground/) work in Playground

cc @bgrgicak @brandonpayton @wojtekn

### DIFF
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -253,6 +253,13 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		$this->assertEquals( 1, $result[0]->output );
 	}
 
+	public function testShowCreateTableNotFound() {
+		$this->assertQuery(
+			'SHOW CREATE TABLE _no_such_table;'
+		);
+		$results = $this->engine->get_query_results();
+		$this->assertCount(0, $results);
+	}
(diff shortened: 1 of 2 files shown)


adamziel:
This solves a real problem so I'll go ahead and merge, but let's revisit in case there's more to `SHOW TABLES` and returning a flat list. Also, GitHub says all checks are still in progress when I can see they passed:

CleanShot 2024-06-06 at 01 24 38@2x

PR #120, opened 2024-06-13
Add REGEXP_REPLACE support (including unit test)

### LINKED ISSUE #47: Support regex_replace
[Torsten Landsiedel](https://twitter.com/zodiac1978) said:

> I would like to know if regexp_replace would be available in Playground
> Fixing the native search is my use:
> https://wordpress.org/plugins/ignore-block-name-in-search/
> But Regex_Replace is only available since MySQL 8.0.4+ and MariaDB 10.0.5+ and my check prevents the activation in Playground.

Let's support regex_replace

Zodiac1978:
Fixes #47

### DIFF
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -128,11 +128,6 @@ class WP_SQLite_Translator_Tests extends TestCase {
 
 		$this->assertQuery( "SELECT * FROM _options WHERE REGEXP_REPLACE(option_name, '(-ignore|-remove)', '') = 'test'" );
 		$this->assertCount( 2, $this->engine->get_query_results() );
-
-		$result1 = $this->engine->query( "SELECT option_value FROM _options WHERE option_name='test-ignore';" );
-		$result2 = $this->engine->query( "SELECT option_value FROM _options WHERE option_name='test-remove';" );
-		$this->assertEquals( '1', $result1[0]->option_value );
-		$this->assertEquals( '2', $result2[0]->option_value );
 	}
 
 	public function testInsertDateNow() {
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -119,16 +119,17 @@ class WP_SQLite_Translator_Tests extends TestCase {
 	}
 
 	public function testRegexpReplace() {
+		/* Testing if an actual replacment works correctly */
 		$this->assertQuery(
 			"INSERT INTO _options (option_name, option_value) VALUES ('test-ignore', '1');"
 		);
 		$this->assertQuery(
 			"INSERT INTO _options (option_name, option_value) VALUES ('test-remove', '2');"
 		);
-
 		$this->assertQuery( "SELECT * FROM _options WHERE REGEXP_REPLACE(option_name, '(-ignore|-remove)', '') = 'test'" );
 		$this->assertCount( 2, $this->engine->get_query_results() );
 
+		/* If one of the required parameters is null, the return value is null, copying the MYSQL/MariaDB behavior */
 		$this->assertQuery( "SELECT REGEXP_REPLACE( null, 'a', 'x') as result" );
 		$results = $this->engine->get_query_results();
 		$this->assertEquals( null, $results[0]->result );
@@ -138,6 +139,11 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		$this->assertEquals( null, $results[0]->result );
 
 		$this->assertQuery( "SELECT REGEXP_REPLACE( 'abc', 'a', null) as result" );
+		$results = $this->engine->get_query_results();
+		$this->assertEquals( null, $results[0]->result );
+
+		/* Providing an empty pattern should produce an error - but we changed that to null to avoid breaking things */
+		$this->assertQuery( "SELECT REGEXP_REPLACE( 'abc', '', 'x') as result" );
 		$results = $this->engine->get_query_results();
 		$this->assertEquals( null, $results[0]->result );
 	}
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -128,6 +128,11 @@ class WP_SQLite_Translator_Tests extends TestCase {
 
 		$this->assertQuery( "SELECT * FROM _options WHERE REGEXP_REPLACE(option_name, '(-ignore|-remove)', '') = 'test'" );
 		$this->assertCount( 2, $this->engine->get_query_results() );
+
+		$result1 = $this->engine->query( "SELECT option_value FROM _options WHERE option_name='test-ignore';" );
+		$result2 = $this->engine->query( "SELECT option_value FROM _options WHERE option_name='test-remove';" );
+		$this->assertEquals( '1', $result1[0]->option_value );
+		$this->assertEquals( '2', $result2[0]->option_value );
 	}
 
 	public function testInsertDateNow() {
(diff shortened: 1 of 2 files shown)


tests/WP_SQLite_Translator_Tests.php
```diff
@@ -118,6 +118,23 @@ public static function regexpOperators() {
...
+			"INSERT INTO _options (option_name, option_value) VALUES ('test-remove', '2');"
+		);
+
+		$this->assertQuery( "SELECT * FROM _options WHERE REGEXP_REPLACE(option_name, '(-ignore|-remove)', '') = 'test'" );
+		$this->assertCount( 2, $this->engine->get_query_results() );
+
+		$result1 = $this->engine->query( "SELECT option_value FROM _options WHERE option_name='test-ignore';" );
+		$result2 = $this->engine->query( "SELECT option_value FROM _options WHERE option_name='test-remove';" );
+		$this->assertEquals( '1', $result1[0]->option_value );
+		$this->assertEquals( '2', $result2[0]->option_value );
```
Zodiac1978, line 135:
```suggestion
```

Zodiac1978:
Happy to get your review @bgrgicak ! :)

Zodiac1978:
@adamziel I have now additionally added checks for if any of the required parameters are null.

Copying the result from MySQL:
> If expr, pat, or repl is NULL, the return value is NULL. 
https://dev.mysql.com/doc/refman/8.4/en/regexp.html#function_regexp-replace

tests/WP_SQLite_Translator_Tests.php
```diff
@@ -118,6 +118,30 @@ public static function regexpOperators() {
...
+			"INSERT INTO _options (option_name, option_value) VALUES ('test-ignore', '1');"
+		);
+		$this->assertQuery(
+			"INSERT INTO _options (option_name, option_value) VALUES ('test-remove', '2');"
+		);
+
+		$this->assertQuery( "SELECT * FROM _options WHERE REGEXP_REPLACE(option_name, '(-ignore|-remove)', '') = 'test'" );
+		$this->assertCount( 2, $this->engine->get_query_results() );
+
+		$this->assertQuery( "SELECT REGEXP_REPLACE( null, 'a', 'x') as result" );
```
bgrgicak, line 132: Let's add some comments to explain these edge cases

tests/WP_SQLite_Translator_Tests.php
```diff
@@ -118,6 +118,30 @@ public static function regexpOperators() {
...
+		$this->assertEquals( null, $results[0]->result );
+
+		$this->assertQuery( "SELECT REGEXP_REPLACE( 'abc', null, 'x') as result" );
+		$results = $this->engine->get_query_results();
+		$this->assertEquals( null, $results[0]->result );
+
+		$this->assertQuery( "SELECT REGEXP_REPLACE( 'abc', 'a', null) as result" );
+		$results = $this->engine->get_query_results();
+		$this->assertEquals( null, $results[0]->result );
+	}
```
bgrgicak, line 143: Can we add a test for empty string patterns?

bgrgicak (approved):
Thank you!

adamziel (approved):
Lgtm!

adamziel:
Actually, let’s also test for empty string pattern, not just a null one

wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
```diff
@@ -493,6 +494,54 @@ public function regexp( $pattern, $field ) {
...
+
+		/* Return null if one of the required parameter is null */
+		if ( is_null( $field ) || is_null( $pattern ) || is_null( $replacement ) ) {
+			return null;
+		}
+
+		/* Return null if the pattern is empty - this changes MySQL/MariaDB behavior! */
+		if ( empty( $pattern ) ) {
+			return null;
+		}
```
adamziel, line 530: Oh! Wha is the MySQL/MariaDB behavior? to just use an empty pattern? If so, let's do that.

wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
```diff
@@ -493,6 +494,54 @@ public function regexp( $pattern, $field ) {
...
+		 * byte codes.
+		 *
+		 * The REGEXP function can't be easily made to accept two
+		 * parameters, so we'll have to use a hack to get around this.
+		 *
+		 * If the first character of the pattern is a null byte, we'll
+		 * remove it and make the comparison case-sensitive. This should
+		 * be reasonably safe since PHP does not allow null bytes in
+		 * regular expressions anyway.
+		 */
```
adamziel, line 520: Does regexp_replace support the `BINARY` keyword? I couldn't find anything. If it doesn't, let's remove this comment and the special casing for the null byte.

wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
```diff
@@ -493,6 +494,54 @@ public function regexp( $pattern, $field ) {
...
+	/**
+	 * Method to emulate MySQL REGEXP_REPLACE() function.
+	 *
+	 * @param string|array $pattern     Regular expression to search for (or array of strings).
+	 * @param string|array $replacement The string or an array with strings to replace.
+	 * @param string|array $field       Haystack.
+	 *
+	 * @return Array if the field parameter is an array, or a string otherwise.
+	 */
+	public function regexp_replace( $field, $pattern, $replacement ) {
```
adamziel, line 506: I just checked and in MySQL REGEXP_REPLACE() also [takes three optional parameters](https://dev.mysql.com/doc/refman/8.4/en/regexp.html#function_regexp-replace):

> pos: The position in expr at which to start the search. If omitted, the default is 1.
> occurrence: Which occurrence of a match to replace. If omitted, the default is 0 (which means “replace all occurrences”).
> match_type: A string that specifies how to perform matching. The meaning is as described for [REGEXP_LIKE()](https://dev.mysql.com/doc/refman/8.4/en/regexp.html#function_regexp-like).

Would you be up for adding a support for those? If not, that's fine, but let's at add them to the function signature and fail with an informative warning (`REGEXP_REPLACE() don't support non-default values for $pos (fourth argument), .... if you need it then you can contribute here: ....`) if they're set to non-default values.

adamziel:
I just had a moment to review this in a more focused setting and I left a few more notes. It is pretty close!

wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
Zodiac1978: It is throwing an error (because without pattern it can't work at all):
> Error in the SQL query (3685): Illegal argument to a regular expression. 

@bgrgicak was suggesting to return null instead to avoid breaking things, I assume.

wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
Zodiac1978: > Would you be up for adding a support for those?

I will try to do that!

wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
Zodiac1978: Isn't the binary-mode something global?
See: https://mariadb.com/docs/server/ref/mdb/cli/mariadb/binary-mode/

wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
bgrgicak: Null seemed like a better response from an error. @Zodiac1978 let's update the comment to make it clear what _this changes MySQL/MariaDB behavior_ means.

bgrgicak:
@JanJakes is this PR still relevant?

JanJakes:
@bgrgicak I think it is; we only need to update the tests as well for the new driver. I can take a look at this.

JanJakes:
Closing this in favor of https://github.com/WordPress/sqlite-database-integration/pull/368.

Issue #122, opened 2024-07-17
Decimal column type not stored correctly in _mysql_data_types_cache


jeroenpf:
For decimal columns, the format of the type + modifier is `decimal(x,y)` where `x` is the length and `y` is the number of decimal places. The `mysql_type` column currently contains values like `decimal(x,`.  This causes issues when trying to get a create statement:

```
CREATE TABLE wp_wc_order_operational_data (
	`id` bigint(20) unsigned PRIMARY KEY AUTO_INCREMENT NOT NULL,
	...
	`date_completed_gmt` datetime DEFAULT NULL,
	`shipping_tax_amount` decimal(26, DEFAULT NULL,
	`shipping_total_amount` decimal(26, DEFAULT NULL,
	`discount_tax_amount` decimal(26, DEFAULT NULL,
	`discount_total_amount` decimal(26, DEFAULT NULL,
	`recorded_sales` tinyint(1) DEFAULT NULL,
	KEY wp_wc_order_operational_data__order_key (order_key),
	UNIQUE KEY wp_wc_order_operational_data__order_id (order_id)
);
```

This can be fixed in  the `WP_SQLite_Translator::skip_mysql_data_type` method. Here we consume exactly 3 tokens, which works only when there is a single value between the brackets `(x)`. Instead, we should consume/skip tokens until we find the closing `)` of the type modifier.

PR #126, opened 2024-07-17
Support multi-value column type modifiers for `_mysql_data_types_cache`

jeroenpf:
Fixes: #122 

This fix will skip over tokens until it finds the closing bracket of the type modifier `)` instead of consuming exactly 3. This should account for data types that support more than one value as a modifier, such as: enum, float, decimal, double.

Additionally, I've added support for enums. There are plugins that use enum fields and it makes sense to convert those to text in SQLite.

## Testing instructions
Confirm the tests pass

### DIFF
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -704,6 +704,43 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		$this->assertEquals( 1, $result );
 	}
 
+	public function testCreateTableWithDecimalColumn() {
+		$result = $this->assertQuery(
+			'CREATE TABLE wptests_users (
+				ID bigint(20) unsigned NOT NULL auto_increment,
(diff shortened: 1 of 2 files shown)


katinthehatsite:
Thanks for adding the test - are there any other steps that I could manually run on my end to confirm that the issue is fixed?

jeroenpf:
@katinthehatsite 

The unit tests should cover this change but if you want to do it manually you could try:

Create a site in Studio, replace the sqlite database integration plugin with a clone of this repo and branch and then install a plugin that uses enum and/or decimal fields. The woocommerce plugin will add tables with decimals and the [redirection plugin](https://wordpress.org/plugins/redirection/) will create tables with enums.

After you installed those plugins, you can inspect the `_mysql_data_types_cache` table and ensure that the columns that have decimal or enum types are properly represented in that table.

After that you can try and run a query using `$wpdb->query('SHOW CREATE table`);` and see if it outputs a valid MySQL-compatible create statement.

bgrgicak (approved):

Issue #123, opened 2024-07-17
`Show create` incorrectly handles composite PKs


jeroenpf:
When running `SHOW CREATE table` the resulting create statement does not correctly handle tables that have a composite PK (a PK with more than one column).

Here is an example of the original (MySQL) table definition where you can see that there is a PK on `object_id` and `term_taxonomy_id`:

```
CREATE TABLE `wp_term_relationships` (
 `object_id` bigint(20) unsigned NOT NULL DEFAULT 0,
 `term_taxonomy_id` bigint(20) unsigned NOT NULL DEFAULT 0,
 `term_order` int(11) NOT NULL DEFAULT 0,
 PRIMARY KEY (`object_id`,`term_taxonomy_id`),
 KEY `term_taxonomy_id` (`term_taxonomy_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
```
In SQLite, the table retains the composite PK but when running the `SHOW CREATE` table query, we get the following output:

```
CREATE TABLE wp_term_relationships (
	`object_id` bigint(20) unsigned PRIMARY KEY AUTO_INCREMENT NOT NULL,
	`term_taxonomy_id` bigint(20) unsigned NOT NULL,
	`term_order` int(11) NOT NULL,
	KEY wp_term_relationships__term_taxonomy_id (term_taxonomy_id),
	UNIQUE KEY sqlite_autoindex_wp_term_relationships_1 (object_id, term_taxonomy_id)
);
```

As you can see, the primary key is now only on the column `object_id` making this table unusable.

PR #127, opened 2024-07-17
Address PK issues for show create table queries

jeroenpf:
Fixes: #123 

In this PR i propose a refactor of the handling of `show create table` queries.

The main issue was that it is possible for a primary key to be composite (more than one column). The existing implementation only supports a single PK which causes the `CREATE TABLE` statement to be invalid. An example of such a table is the `wp_term_relationships` found in WordPress:

```mysql
-- original MySQL statement
CREATE TABLE `wp_term_relationships` (
 `object_id` bigint(20) unsigned NOT NULL DEFAULT 0,
 `term_taxonomy_id` bigint(20) unsigned NOT NULL DEFAULT 0,
 `term_order` int(11) NOT NULL DEFAULT 0,
 PRIMARY KEY (`object_id`,`term_taxonomy_id`),
 KEY `term_taxonomy_id` (`term_taxonomy_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;

-- statement produced by SQLite database integration
CREATE TABLE wp_term_relationships (
	`object_id` bigint(20) unsigned PRIMARY KEY AUTO_INCREMENT NOT NULL,
	`term_taxonomy_id` bigint(20) unsigned NOT NULL,
	`term_order` int(11) NOT NULL,
	KEY wp_term_relationships__term_taxonomy_id (term_taxonomy_id),
	UNIQUE KEY sqlite_autoindex_wp_term_relationships_1 (object_id, term_taxonomy_id)
);
```
SQLite also creates a unique key for any primary key (single or composite) which is something that is not necessary for MySQL and should be skipped from the final create statement.

## In this PR

- Refactor the code to separate things into smaller methods.
- Add the PK definition before the keys and remove it from the column definitions.
- Use backticks around table names, column names and key identifiers. This is to prevent issues when names of identifiers are using reserved keywords, ensure case sensitivity and allow them to contain special chars or spaces.
- Improve detection of AUTO_INCREMENT column. Now we falsely mark columns as AUTO_INCREMENT just because they happen to be the primary key and be of type integer.
- Fix creating an invalid UNIQUE KEY for the PK. When we find that the unique key is identical to the primary key, we can skip the unique key as it is redundant.

The final query output for `show create wp_term_relationships` looks as follows:

```mysql
DROP TABLE IF EXISTS `wp_term_relationships`;
CREATE TABLE `wp_term_relationships` (
	`object_id` bigint(20) unsigned NOT NULL DEFAULT 0,
	`term_taxonomy_id` bigint(20) unsigned NOT NULL DEFAULT 0,
	`term_order` int(11) NOT NULL DEFAULT 0,
	PRIMARY KEY (`object_id`, `term_taxonomy_id`),
	KEY `wp_term_relationships__idx_term_taxonomy_id_x` (`term_taxonomy_id`)
);
```

## Testing

- Run the unit tests
- Create a site with SQLite and then run `$wpdb->query('show create wp_term_relationships')` - The result should be a valid MySQL compatible query as shown above.

### DIFF
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -396,6 +396,16 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		);
 	}
 
+	public function testShowCreateTableWithColumnKeys() {
+		$this->assertQuery(
+			"CREATE TABLE _tmp_table (
+	`ID` bigint PRIMARY KEY AUTO_INCREMENT NOT NULL,
+	`option_name` varchar(255) DEFAULT '',
+	`option_value` text NOT NULL DEFAULT '',
+	KEY _tmp_table__composite (option_name, option_value),
+	UNIQUE KEY _tmp_table__option_name (option_name) );" );
+	}
+
 	public function testSelectIndexHintForce() {
 		$this->assertQuery( "INSERT INTO _options (option_name) VALUES ('first');" );
 		$result = $this->assertQuery(
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -279,7 +279,7 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		# TODO: Should we fix mismatch with original `option_value` text NOT NULL,` without default?
 		$this->assertEquals(
 			"CREATE TABLE `_tmp_table` (
-	`ID` bigint NOT NULL DEFAULT 0 AUTO_INCREMENT,
+	`ID` bigint NOT NULL AUTO_INCREMENT,
 	`option_name` varchar(255) DEFAULT '',
 	`option_value` text NOT NULL DEFAULT '',
 	PRIMARY KEY (`ID`),
@@ -308,7 +308,7 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		# TODO: Should we fix mismatch with original `option_value` text NOT NULL,` without default?
 		$this->assertEquals(
 			"CREATE TABLE `_tmp_table` (
-	`ID` bigint NOT NULL DEFAULT 0 AUTO_INCREMENT,
+	`ID` bigint NOT NULL AUTO_INCREMENT,
 	`option_name` varchar(255) DEFAULT '',
 	`option_value` text NOT NULL DEFAULT '',
 	PRIMARY KEY (`ID`),
@@ -361,7 +361,7 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		$results = $this->engine->get_query_results();
 		$this->assertEquals(
 			'CREATE TABLE `_tmp_table` (
-	`ID` bigint NOT NULL DEFAULT 0 AUTO_INCREMENT,
+	`ID` bigint NOT NULL AUTO_INCREMENT,
 	`option_name` smallint NOT NULL DEFAULT 14,
 	`option_value` text NOT NULL DEFAULT \'\',
 	PRIMARY KEY (`ID`),
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -278,12 +278,13 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		$results = $this->engine->get_query_results();
 		# TODO: Should we fix mismatch with original `option_value` text NOT NULL,` without default?
 		$this->assertEquals(
-			"CREATE TABLE _tmp_table (
-	`ID` bigint PRIMARY KEY AUTO_INCREMENT NOT NULL,
+			"CREATE TABLE `_tmp_table` (
+	`ID` bigint NOT NULL DEFAULT 0 AUTO_INCREMENT,
 	`option_name` varchar(255) DEFAULT '',
 	`option_value` text NOT NULL DEFAULT '',
-	KEY _tmp_table__composite (option_name, option_value),
-	UNIQUE KEY _tmp_table__option_name (option_name)
+	PRIMARY KEY (`ID`),
+	KEY `_tmp_table__composite` (`option_name`, `option_value`),
+	UNIQUE KEY `_tmp_table__option_name` (`option_name`)
 );",
 			$results[0]->{'Create Table'}
 		);
@@ -306,12 +307,13 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		$results = $this->engine->get_query_results();
 		# TODO: Should we fix mismatch with original `option_value` text NOT NULL,` without default?
 		$this->assertEquals(
-			"CREATE TABLE _tmp_table (
-	`ID` bigint PRIMARY KEY AUTO_INCREMENT NOT NULL,
+			"CREATE TABLE `_tmp_table` (
+	`ID` bigint NOT NULL DEFAULT 0 AUTO_INCREMENT,
 	`option_name` varchar(255) DEFAULT '',
 	`option_value` text NOT NULL DEFAULT '',
-	KEY _tmp_table__composite (option_name, option_value),
-	UNIQUE KEY _tmp_table__option_name (option_name)
+	PRIMARY KEY (`ID`),
+	KEY `_tmp_table__composite` (`option_name`, `option_value`),
+	UNIQUE KEY `_tmp_table__option_name` (`option_name`)
 );",
 			$results[0]->{'Create Table'}
 		);
@@ -329,8 +331,8 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		);
 		$results = $this->engine->get_query_results();
 		$this->assertEquals(
-			'CREATE TABLE _tmp_table (
-	`ID` bigint NOT NULL
+			'CREATE TABLE `_tmp_table` (
+	`ID` bigint NOT NULL DEFAULT 0
 );',
 			$results[0]->{'Create Table'}
 		);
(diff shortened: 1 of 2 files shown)


wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3488,6 +3441,170 @@ private function execute_show() {
...
+		foreach ( $columns as $column ) {
+			$definition   = array();
+			$definition[] = '`' . $column->name . '`';
+			$definition[] = $this->get_cached_mysql_data_type( $table_name, $column->name ) ?? $column->name;
+
+			if ( '1' === $column->notnull ) {
+				$definition[] = 'NOT NULL';
+			}
+
+			if ( '' !== $column->dflt_value ) {
```
jeroenpf, line 3508: It would probably be better to skip the default if the current column is a PK and/or Auto increment column.

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3488,6 +3441,170 @@ private function execute_show() {
...
+				$definition[] = 'DEFAULT ' . $column->dflt_value; // quotes?
```
bgrgicak: Could we please expand this comment, it's unclear what it means.

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3488,6 +3441,170 @@ private function execute_show() {
...
+		$pks = array();
+		foreach ( $columns as $column ) {
+			if ( '0' !== $column->pk ) {
+				$pks[] = $column->name;
+			}
+		}
+
+		foreach ( $this->get_keys( $table_name ) as $key ) {
+			// If the PK columns are the same as the unique key columns, skip the key.
+			// This is because the PK is already unique in MySQL.
```
bgrgicak, line 3542: Nice, these comments help a lot.

tests/WP_SQLite_Translator_Tests.php
```diff
@@ -278,12 +278,13 @@ public function testShowCreateTable1() {
 		$results = $this->engine->get_query_results();
 		# TODO: Should we fix mismatch with original `option_value` text NOT NULL,` without default?
 		$this->assertEquals(
-			"CREATE TABLE _tmp_table (
-	`ID` bigint PRIMARY KEY AUTO_INCREMENT NOT NULL,
```
bgrgicak, line 282: Why do we need to remove PRIMARY KEY from tests?

I tried reverting tests and they failed when `PRIMARY KEY`. Is this expected?

wp-includes/sqlite/class-wp-sqlite-translator.php
bgrgicak: What would be the benefit of skipping the default?

tests/WP_SQLite_Translator_Tests.php
jeroenpf: The reason for this is because the change i made moves the primary key constraint away from the column definition. The primary key constraint can now be found amongst the other key definitions. In this case:

```mysql
CREATE TABLE `_tmp_table` (
	`ID` bigint NOT NULL AUTO_INCREMENT,
	`option_name` varchar(255) DEFAULT '',
	`option_value` text NOT NULL DEFAULT '',
	PRIMARY KEY (`ID`), -- Primary key has been moved here.
	KEY `_tmp_table__composite` (`option_name`, `option_value`),
	UNIQUE KEY `_tmp_table__option_name` (`option_name`)
```

You could add PRIMARY KEY on the column level if the table has a single column PK but for consistency, I now always move it below column definitions. This also prevents us from having to write separate logic for single and composite PKs.

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3488,6 +3441,170 @@ private function execute_show() {
 		}
 	}
 
+	private function generate_create_statement() {
+		$table_name = $this->rewriter->consume()->value;
+		$columns    = $this->get_table_columns( $table_name );
```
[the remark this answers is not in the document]
jeroenpf: Thanks for bringing up this point, it made me realize another important detail. I think the two methods are not doing the same. `get_columns_from` is returning a result specifically formatted for the `show columns from` query. However, we could reuse `get_table_columns` inside the `get_columns_from` method.

One problem I now realize I overlooked and further justifies the need for a separate method that returns the raw data is that we need to sort the PK columns. In the `PRAGMA table_info()` query, the position of each PK is returned. The position is not necessarily identical to the order in which they appear. The order of a composite PK is important for indexing and default ordering purposes.

E.g. for the purpose of illustration i created a PK for which the column order is not the same as the order in which the columns are listed. `pragma table_info()` returns:

| cid | name | type | notnull | dflt\_value | pk |
| :--- | :--- | :--- | :--- | :--- | :--- |
| 0 | term\_taxonomy\_id | INTEGER | 1 | 0 | 2 |
| 1 | object\_id | INTEGER | 1 | 0 | 1 |
| 2 | term\_order | INTEGER | 1 | 0 | 0 |


With that, I propose:

1. We use `get_table_columns` inside the `get_columns_from`.
2. I adjust the comment for the `get_columns_from` to clearly indicate its purpose. (which is to return output formatted as `SHOW COLUMNS FROM` MySQL queries).
3. I implement sorting of the PKs so that they appear in the correct order in the final CREATE statement.

What are your thoughts on this?

tests/WP_SQLite_Translator_Tests.php
bgrgicak: But why are the old tests breaking? I'm concerned that this PR will break existing support in favor of adding the new feature.

Is there a way for us to have both your new tests and the old test working?

wp-includes/sqlite/class-wp-sqlite-translator.php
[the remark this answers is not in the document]
bgrgicak: This sounds like a good solution to me, but I'm still wondering if it would be better to update `get_columns_from` to do what you need instead of adding another way of parsing the data from `PRAGMA table_info()`.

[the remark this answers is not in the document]
jeroenpf: The way I see it is that we have two distinct desired outputs. One is for `show columns from` and the other `show create table`. Both use the raw data provided by the pragma query under the hood but the goals and outputs are different. `get_columns_from` adds no value over using the raw data directly in this case and i think its best to separate these concerns. Finally, now that we established that we also need to consider the order of the columns in the PK, we can't rely on `get_columns_from` as its output is meant to directly resemble `show columns from` output as MySQL would provide. It does not convey the order of columns in the PK.

This is the output of `get_columns_from`
| Field | Type         | Null | Key | Default | Extra          |
|-------|--------------|------|-----|---------|----------------|
| id    | int(11)      | NO   | PRI | NULL    | auto_increment |
| name  | varchar(255) | YES  |     | NULL    |                |
| date  | date         | YES  |     | NULL    |                |

Do you see another way that this would make sense?

tests/WP_SQLite_Translator_Tests.php
jeroenpf: If i understand your point correctly, you only reverted the tests and then ran them against the new code and then noticed that the tests did not work?

Also, if i understand correct, you want the output of the create table statement to be the same as it was before?

So that means:

1. Not using backticks for keywords, column and table names.
2. Placing the PRIMARY KEY clause next to the column, but only if there is a single column PK
3. Keeping the UNIQUE KEY that SQLIte adds for PKs (because it does not support uniqueness of PKs without it).

Is this correct?

wp-includes/sqlite/class-wp-sqlite-translator.php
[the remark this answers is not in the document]
bgrgicak: > The way I see it is that we have two distinct desired outputs. One is for show columns from and the other show create table

That sounds good. Let's just make their purpose clear with comments.

tests/WP_SQLite_Translator_Tests.php
bgrgicak: > If i understand your point correctly, you only reverted the tests and then ran them against the new code and then noticed that the tests did not work?

Yes.

> Also, if i understand correct, you want the output of the create table statement to be the same as it was before?

No.

I tested this now by adding back an old test to your PR and it passed, so we are good.
Would it make sense to add this test?

```
	public function testShowCreateTableWithColumnKeys() {
		$this->assertQuery(
			"CREATE TABLE _tmp_table (
				`ID` bigint PRIMARY KEY AUTO_INCREMENT NOT NULL,
				`option_name` varchar(255) DEFAULT '',
				`option_value` text NOT NULL DEFAULT '',
				KEY _tmp_table__composite (option_name, option_value),
				UNIQUE KEY _tmp_table__option_name (option_name)
			);"
		);
	}
```

bgrgicak (approved):
@jeroenpf the PR looks good to me. Thanks for working on this.

I just left one question about tests after we answer that I'm happy to merge the PR.

tests/WP_SQLite_Translator_Tests.php
jeroenpf: Thanks for clarifying and providing an additional unit test. It makes sense now.

I've added the test so we can be sure that create statements with PKs at the column level still works as expected.

jeroenpf:
Thanks @bgrgicak for the review. I've addressed your latest comment so we should be good to go.

PR #129, opened 2024-07-18
Support the current_timestamp() function

akirk:
[The function `CURRENT_TIMESTAMP()` is a synonym for NOW()](https://dev.mysql.com/doc/refman/8.0/en/functions.html#function_current-timestamp) which we already support.

cc @costasovo


akirk:
Ok, I think I need some help here. I think the problem is that [`CURRENT_TIMESTAMP` can also be a keyword](https://github.com/WordPress/sqlite-database-integration/blob/d24fff8e58b8dc232c72d8cac78a4c8ac283dd86/wp-includes/sqlite/class-wp-sqlite-lexer.php#L1269) in a field definition.

JanJakes:
@akirk I think the issue is actually that in SQLite, the `current_timestamp` function [can't be used with partentheses](https://www.sqlitetutorial.net/sqlite-date-functions/sqlite-current_timestamp/) — `SELECT CURRENT_TIMESTAMP` works, `SELECT CURRENT_TIMESTAMP()` doesn't.

In a bigger picture:
1. `CURRENT_TIMESTAMP` is supported by SQLite, but only without parentheses, `NOW()` is not supported.
2. `CURRENT_TIMESTAMP` is SQL standard, `NOW()` is not.
3. In the current implementation, `NOW()` is [emulated in PHP](https://github.com/WordPress/sqlite-database-integration/blob/4c63355e4eb959fcffc20ec1a0b8395a15e845b6/wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php#L126), although it would be probably more correct to just convert it to SQLite-native `CURRENT_TIMESTAMP`.
4. `CURRENT_TIMESTAMP()` can be only stripped off the optional parentheses and consumed to the SQL query.

JanJakes:
In short, I think the `sqliteCreateFunction` has no effect because `CURRENT_TIMESTAMP` is a SQLite-native function, so it probably cannot be overwritten. The quickest fix would be to remove the parentheses.

PR #130, opened 2024-07-18
Support CREATE TABLE unique keys with no key name

akirk:
In a table like
```sql
CREATE TABLE `test` (
	`id` INT,
	`name` VARCHAR(255),
	`other` VARCHAR(255),
	PRIMARY KEY (id),
	UNIQUE KEY (name)
)
```

the internal name for the unique key would be `(` because typically a key would be defined with
```
UNIQUE KEY `name` (name)
```

Usually this wrong key name is not a problem unless a `INSERT .. ON DUPLICATE KEY` statement is used where the resulting `ON CONFLICT( "keyname" )` would be transformed to `ON CONFLICT( "" )` which then fails.

To fix this, this now derives a key name in case it is omitted..

cc @costasovo

### DIFF
--- a/tests/WP_SQLite_Query_Tests.php
+++ b/tests/WP_SQLite_Query_Tests.php
@@ -519,6 +519,23 @@ QUERY;
 			ON DUPLICATE KEY UPDATE `text` = "test1"'
 		);
 	}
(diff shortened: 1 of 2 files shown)


brandonpayton (approved):
This looks good, and the updated tests pass. Thanks, @akirk!

PR #132, opened 2024-07-19
Fix PHP 8 issue with `strtoupper()`

### LINKED ISSUE #54: Error en modo depuración
EN MODO DEPURACION TENGO ESTE ERROR

Deprecated: strtoupper(): Passing null to parameter  # 1 ($string) of type string is deprecated in /bitnami/wordpress/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-token.php on line 222

Algún experto me dice porque?

Saludos

aristath:
Fixes #54

### DIFF
--- a/wp-includes/sqlite/class-wp-sqlite-token.php
+++ b/wp-includes/sqlite/class-wp-sqlite-token.php
@@ -218,7 +218,7 @@ class WP_SQLite_Token {
 		return (
 			( null === $type || $this->type === $type )
 			&& ( null === $flags || ( $this->flags & $flags ) )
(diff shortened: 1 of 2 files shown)


jeroenpf:
I found a similar issue in the `execute_show` [here](https://github.com/WordPress/sqlite-database-integration/blob/develop/wp-includes/sqlite/class-wp-sqlite-translator.php#L3249).

`Deprecated: strtoupper(): Passing null to parameter #1 ($string) of type string is deprecated`

aristath:
Thanks @jeroenpf, fixed 👍

Issue #124, opened 2024-07-17
Identifiers for keys are too long in `show create` output


jeroenpf:
When running `show create` queries, the output returns a create statement that has new key names that exceed the 64-character limit for key identifiers in MySQL.

Here is an example of such a table:

```
CREATE TABLE wp_woocommerce_downloadable_product_permissions (
	`permission_id` bigint(20) unsigned PRIMARY KEY AUTO_INCREMENT NOT NULL,
	...
	`download_count` bigint(20) unsigned NOT NULL,
	KEY wp_woocommerce_downloadable_product_permissions__user_order_remaining_expires (user_id, order_id, downloads_remaining, access_expires),
	KEY wp_woocommerce_downloadable_product_permissions__order_id (order_id),
	KEY wp_woocommerce_downloadable_product_permissions__download_order_product (download_id, order_id, product_id),
	KEY wp_woocommerce_downloadable_product_permissions__download_order_key_product (product_id, order_id, order_key, download_id)
);
```

Additionally, the key names do not match those of the original MySQL table:

```
CREATE TABLE `wp_woocommerce_downloadable_product_permissions` (
   `permission_id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
   ...
   `download_count` bigint(20) unsigned NOT NULL DEFAULT 0,
   PRIMARY KEY (`permission_id`),
   KEY `download_order_key_product` (`product_id`,`order_id`,`order_key`(16),`download_id`),
   KEY `download_order_product` (`download_id`,`order_id`,`product_id`),
   KEY `order_id` (`order_id`),
   KEY `user_order_remaining_expires` (`user_id`,`order_id`,`downloads_remaining`,`access_expires`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
```

We should either use the original key names or generate names that are no longer than 64 characters.

PR #133, opened 2024-07-19
Fix length of index names for a create table statement

jeroenpf:
Fixes: #124 

This PR introduces some changes to how we determine the name of an index. SQLite requires that each index is uniquely named across the database (not just the table). This causes issues when multiple tables have the same key names when importing from MySQL. To avoid these issues, the index name is prefixed with the table name when it is created. This however results in the indexes having names that are too long for MySQL so when we run `show create table` the create statements have these prefixed keys that are too long.

To solve this,  we keep adding the table name as a prefix but we remove it when we generate the create statement. Furthermore, we ensure that the table name itself has no double underscores in it so we can be sure that we can split on `__` safely. The result is that each create table statement will have the original index name and prevents it from being too long.

### DIFF
--- a/wp-includes/sqlite/class-wp-sqlite-translator.php
+++ b/wp-includes/sqlite/class-wp-sqlite-translator.php
@@ -898,8 +898,7 @@ class WP_SQLite_Translator {
 			if ( 'UNIQUE INDEX' === $index_type ) {
 				$unique = 'UNIQUE ';
 			}
-			$index_name = "{$table->name}__{$constraint->name}";
-
+			$index_name = $this->generate_index_name( $table->name, $constraint->name );
 			$this->execute_sqlite_query(
 				"CREATE $unique INDEX \"$index_name\" ON \"{$table->name}\" (\"" . implode( '", "', $constraint->columns ) . '")'
 			);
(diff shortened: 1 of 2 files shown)


jeroenpf:
@bgrgicak Could you have a look at this one?

bgrgicak (approved):

bgrgicak:
Thank you @jeroenpf for contributing these fixes!

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -4188,4 +4192,18 @@ public function rollback() {
...
+	 * Create an index name consisting of table name and original index name.
+	 * This is to avoid duplicate index names in SQLite.
+	 *
+	 * @param $table
+	 * @param $original_index_name
+	 *
+	 * @return string
+	 */
+	private function generate_index_name( $table, $original_index_name ) {
+		// Strip the occurrences of 2 or more consecutive underscores to allow easier splitting on __ later.
```
brandonpayton, line 4206: @jeroenpf it looks like this would make it so we cannot derive the original mysql index name from sqlite index name because we cannot tell which underscores used to be double-underscores.

So import/export operations are not symmetrical/reversible. Do we care about that? Offhand, I think maybe we should.

jeroenpf: @brandonpayton for this reason i strip out double underscores from the table name alone. Later we can split on the first occurrence of double underscores. The idea was that it does not matter if the index name itself has double underscores so they would be preserved. Maybe I am overlooking something?

Example:

`table__name` and `some_index__name` becomes `table_name__some_index__name` and later we split on the first occurrence of `__` which which leaves us with [`table_name`, `some_index__name`].

I do however see one error in the explode logic that is used to get the index name; i should have passed in a limit so that it only separates on the first occurrence of `__`.  I will create a new PR to address this.

brandonpayton: > for this reason i strip out double underscores from the table name alone.

@jeroenpf thank you for pointing that out. I had missed that this was just for the separator between table name and index. The naming scheme makes sense to me now.

PR #134, opened 2024-07-19
Ensure that an index is also created using IF NOT EXISTS if it comes from a CREATE TABLE IF NOT EXISTS

akirk:
Before this the unit test failed.

cc @JanJakes

### DIFF
--- a/tests/WP_SQLite_Query_Tests.php
+++ b/tests/WP_SQLite_Query_Tests.php
@@ -537,7 +537,7 @@ QUERY;
 		);
 	}
 
-	public function testOnCreateTableIfNotExistswithIndexAdded() {
+	public function testOnCreateTableIfNotExistsWithIndexAdded() {
 		$this->assertQuery(
 			'CREATE TABLE IF NOT EXISTS `test` (
 				`id` INT,
--- a/tests/WP_SQLite_Query_Tests.php
+++ b/tests/WP_SQLite_Query_Tests.php
@@ -539,7 +539,7 @@ QUERY;
 
 	public function testOnCreateTableIfNotExistsWithIndexAdded() {
 		$this->assertQuery(
-			'CREATE TABLE IF NOT EXISTS `test` (
+			'CREATE TABLE IF not EXISTS `test` (
 				`id` INT,
 				`name` VARCHAR(255),
 				`other` VARCHAR(255),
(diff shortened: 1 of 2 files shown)


tests/WP_SQLite_Query_Tests.php
```diff
@@ -537,6 +537,27 @@ public function testOnDuplicateKeyWithUnnamedKeys() {
 		);
 	}
 
+	public function testOnCreateTableIfNotExistswithIndexAdded() {
```
brandonpayton, line 540: Small nit:
```suggestion
	public function testOnCreateTableIfNotExistsWithIndexAdded() {
```

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -888,6 +888,8 @@ private function execute_create_table() {
 			')'
 		);
 
+		$if_not_exists = false === strpos( $create_query, 'IF NOT EXISTS' ) ? '' : 'IF NOT EXISTS';
```
brandonpayton, line 891: @akirk, something just occurred to me that probably affects other parts of this plugin as well:
Aren't these keywords case insensitive, and if so, we should probably make the check case insensitive.

akirk: Well, there is also potential whitespace deviations. Either [we normalize the output here](https://github.com/WordPress/sqlite-database-integration/blob/develop/wp-includes/sqlite/class-wp-sqlite-query-rewriter.php#L84C18-L84C35), or we use a regex with `\s+` instead of spaces.

aristath: We could also use `stripos` instead of `strpos` to make this case-insensitive... I'd rather we limit the use of regex if not absolutely necessary.

akirk: But what if someone writes ` if NOT   exists`? One other option is to change `get_updated_query()` so that it will return normalized SQL, like unify whitespace, uppercase lexer tokens, then we can just use `strpos`.

aristath: Fair point 👍

aristath (approved):

PR #136, opened 2024-07-25
Allow double underscores in index names

jeroenpf:
See: https://github.com/WordPress/sqlite-database-integration/pull/133#discussion_r1690678568

When creating new tables in SQLite, original index names are prefixed with the table name. This is because unlike MySQL, SQLite does not allow identical index names across the database. Adding a table prefix ensures the index names are unique.

When running a `show create table` query, we restore the original index name by removing the table prefix. We split on double underscores but we did not take into account that the original index name can also contain `__`. This PR fixes an issue where we explode without a limit. We only want to explode on the first occurrence of `__`. Unit tests have been added to ensure this behavior.

### DIFF
--- a/wp-includes/sqlite/class-wp-sqlite-translator.php
+++ b/wp-includes/sqlite/class-wp-sqlite-translator.php
@@ -3570,9 +3570,7 @@ class WP_SQLite_Translator {
 			$key_definition[] = 'KEY';
 
 			// Remove the prefix from the index name if there is any. We use __ as a separator.
-			$index_name = strstr( $key['index']['name'], '__' )
-				? explode( '__', $key['index']['name'] )[1]
-				: $key['index']['name'];
+			$index_name = explode( '__', $key['index']['name'], 2 )[1] ?? $key['index']['name'];
 
 			$key_definition[] = sprintf( '`%s`', $index_name );
 
@@ -4214,7 +4212,8 @@ class WP_SQLite_Translator {
 	 * @return string
 	 */
 	private function generate_index_name( $table, $original_index_name ) {
-		// Strip the occurrences of 2 or more consecutive underscores to allow easier splitting on __ later.
+		// Strip the occurrences of 2 or more consecutive underscores from the table name
+		// to allow easier splitting on __ later.
 		return preg_replace( '/_{2,}/', '_', $table ) . '__' . $original_index_name;
 	}
 }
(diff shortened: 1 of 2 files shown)


wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3570,9 +3570,7 @@ private function get_key_definitions( $table_name, $columns ) {
 			$key_definition[] = 'KEY';
 
 			// Remove the prefix from the index name if there is any. We use __ as a separator.
-			$index_name = strstr( $key['index']['name'], '__' )
-				? explode( '__', $key['index']['name'] )[1]
-				: $key['index']['name'];
+			$index_name = explode( '__', $key['index']['name'], 2 )[1] ?? $key['index']['name'];
```
brandonpayton, line 3573: This looks good to me, but it uses a null coalescing operator which is not supported by our declared minimum PHP version of 5.6. We already have a number of uses of the null coalescing operator, so it's likely we just need to bump the minimum PHP version.

https://github.com/WordPress/sqlite-database-integration/blob/5a2576ba82053c42c9388702727e049a2ccf1eb0/readme.txt#L6

@aristath what do you think?

brandonpayton (commented):
This looks like a solid change.

All that remains is [the question about the null coalescing operator and minimum supported PHP version](https://github.com/WordPress/sqlite-database-integration/pull/136/files#r1692254131) .

wp-includes/sqlite/class-wp-sqlite-translator.php
jeroenpf: I went by the PHP version specified in [composer.php](https://github.com/WordPress/sqlite-database-integration/blob/5a2576ba82053c42c9388702727e049a2ccf1eb0/composer.json#L11) and also found other instances of code that wouldn't be compatible with 5.6. I agree that we should bump the version in `readme.txt` so that it matches the version defined in `composer.php`.

aristath: You're right, we just forgot to change the minimum required PHP version in the readme file 😅 
I pushed a commit in the `develop` branch already, upping the version 👍

brandonpayton: Thank you both!

brandonpayton (approved):
This looks good to me, and the unit tests pass. Thanks, @jeroenpf!

PR #147, opened 2024-08-07
Add handling for adding/modifying columns with FIRST/AFTER

JanJakes:
Add handling for adding/modifying columns with FIRST/AFTER, such as:

```sql
ALTER TABLE _tmp_table ADD COLUMN new_first_column VARCHAR(255) NOT NULL DEFAULT '' FIRST;
ALTER TABLE _tmp_table ADD COLUMN new_column VARCHAR(255) NOT NULL DEFAULT '' AFTER id;
ALTER TABLE _tmp_table CHANGE id id int(11) NOT NULL FIRST;
ALTER TABLE _tmp_table CHANGE id id int(11) NOT NULL AFTER name;
```

SQLite doesn't support FIRST/AFTER, so this implementation simply ignores these. Previously, for ADD COLUMN statements, it ended up with `SQLSTATE[HY000]: General error: 1 near "FIRST": syntax error.` and `SQLSTATE[HY000]: General error: 1 near "AFTER": syntax error.`; for CHANGE statements, this was already ignored and I only added a test.

### DIFF
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -1027,6 +1027,31 @@ class WP_SQLite_Translator_Tests extends TestCase {
 		);
 	}
 
+	public function testAlterTableWithColumnFirstAndAfter() {
+		$this->assertQuery(
+			"CREATE TABLE _tmp_table (
+				id int(11) NOT NULL,
+				name varchar(20) NOT NULL default ''
+			);"
+		);
+
+		$this->assertQuery(
+			"ALTER TABLE _tmp_table ADD COLUMN new_first_column VARCHAR(255) NOT NULL DEFAULT '' FIRST"
+		);
+
+		$this->assertQuery(
+			"ALTER TABLE _tmp_table ADD COLUMN new_column VARCHAR(255) NOT NULL DEFAULT '' AFTER id"
+		);
+
+		$this->assertQuery(
+			"ALTER TABLE _tmp_table CHANGE id id int(11) NOT NULL DEFAULT '' FIRST"
+		);
+
+		$this->assertQuery(
+			"ALTER TABLE _tmp_table CHANGE id id int(11) NOT NULL  DEFAULT '' AFTER name"
+		);
+	}
+
 	public function testAlterTableAddIndex() {
 		$result = $this->assertQuery(
 			"CREATE TABLE _tmp_table (
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -1035,22 +1035,10 @@ class WP_SQLite_Translator_Tests extends TestCase {
 			);"
 		);
 
+		// ADD COLUMN with FIRST
 		$this->assertQuery(
 			"ALTER TABLE _tmp_table ADD COLUMN new_first_column VARCHAR(255) NOT NULL DEFAULT '' FIRST"
 		);
-
-		$this->assertQuery(
-			"ALTER TABLE _tmp_table ADD COLUMN new_column VARCHAR(255) NOT NULL DEFAULT '' AFTER id"
-		);
-
-		$this->assertQuery(
-			"ALTER TABLE _tmp_table CHANGE id id int(11) NOT NULL DEFAULT '' FIRST"
-		);
-
-		$this->assertQuery(
-			"ALTER TABLE _tmp_table CHANGE id id int(11) NOT NULL  DEFAULT '' AFTER name"
-		);
-
 		$results = $this->assertQuery( 'DESCRIBE _tmp_table;' );
 		$this->assertEquals(
 			array(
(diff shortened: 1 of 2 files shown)


tests/WP_SQLite_Translator_Tests.php
```diff
@@ -1027,6 +1027,31 @@ public function testAlterTableAddNotNullVarcharColumn() {
...
+		$this->assertQuery(
+			"CREATE TABLE _tmp_table (
+				id int(11) NOT NULL,
+				name varchar(20) NOT NULL default ''
+			);"
+		);
+
+		$this->assertQuery(
+			"ALTER TABLE _tmp_table ADD COLUMN new_first_column VARCHAR(255) NOT NULL DEFAULT '' FIRST"
+		);
```
brandonpayton, line 1040: It would be good to also make assertions about the resulting table structure like this:
https://github.com/WordPress/sqlite-database-integration/blob/dea3efd7adaa1eb5fdc66248669a8f3707450c20/tests/WP_SQLite_Translator_Tests.php#L885-L907

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -2939,6 +2939,35 @@ private function execute_alter() {
...
+				);
+
+				$comma = $this->rewriter->peek(
+					array(
+						'type'  => WP_SQLite_Token::TYPE_OPERATOR,
+						'value' => ',',
+					)
+				);
+
+				if ( $column_position && ( ! $comma || $column_position->position < $comma->position ) ) {
```
brandonpayton, line 2958: What kind of SQL syntax is this comma check intended to address? Could you provide an example?

JanJakes: @brandonpayton Without the check, it can consume too much, up to the following column definitions.

E.g., for a query like this one:
```sql
ALTER TABLE _tmp_table
ADD COLUMN new1 VARCHAR(255),
ADD COLUMN new2 VARCHAR(255) FIRST
```

You would get a `SQLSTATE[HY000]: General error: 1 near ",": syntax error.`, since you've consumed up to `FIRST` from the second column when processing the first column.

tests/WP_SQLite_Translator_Tests.php
JanJakes: 👍 Added in https://github.com/WordPress/sqlite-database-integration/pull/147/commits/c4efdfd59d6c362237bc1b08477182c4db7a44b5.

JanJakes:
@brandonpayton Thanks! I added a commit and tried the to explain the need for the comma check.

tests/WP_SQLite_Translator_Tests.php
brandonpayton: 👍 Let's add them after each `assertQuery()` here to check each is doing what is expected.

wp-includes/sqlite/class-wp-sqlite-translator.php
brandonpayton: Ah, I see. I hadn't noticed that the translator loops to support the comma-separated MySQL ALTER TABLE syntax.

Let's add a test for comma-separated syntax as well.

brandonpayton: Thanks for explaining!

tests/WP_SQLite_Translator_Tests.php
JanJakes: OK! Added in https://github.com/WordPress/sqlite-database-integration/pull/147/commits/4615adc48f81d0e5fabfe918c500ad6de9bddead.

wp-includes/sqlite/class-wp-sqlite-translator.php
JanJakes: Added a test in https://github.com/WordPress/sqlite-database-integration/pull/147/commits/2c1fe61e9c308564309bcad9ae0227d28b31a4fa.

tests/WP_SQLite_Translator_Tests.php
```diff
@@ -1027,6 +1027,279 @@ public function testAlterTableAddNotNullVarcharColumn() {
...
+
+		// ADD COLUMN with FIRST
+		$this->assertQuery(
+			"ALTER TABLE _tmp_table ADD COLUMN new_first_column VARCHAR(255) NOT NULL DEFAULT '' FIRST"
+		);
+		$results = $this->assertQuery( 'DESCRIBE _tmp_table;' );
+		$this->assertEquals(
+			array(
+				(object) array(
+					'Field'   => 'id',
```
brandonpayton, line 1046: Shouldn't `new_first_column` be first now?

tests/WP_SQLite_Translator_Tests.php
```diff
@@ -1027,6 +1027,279 @@ public function testAlterTableAddNotNullVarcharColumn() {
...
+					'Type'    => 'int(11)',
+					'Null'    => 'NO',
+					'Key'     => '',
+					'Default' => '0',
+					'Extra'   => '',
+				),
+				(object) array(
+					'Field'   => 'name',
```
brandonpayton, line 1089: Shouldn't `new_column` be after `id` now, so the resulting column order would be:
- new_first_column
- id
- new_column
- name

tests/WP_SQLite_Translator_Tests.php
```diff
@@ -1027,6 +1027,279 @@ public function testAlterTableAddNotNullVarcharColumn() {
...
+				),
+				(object) array(
+					'Field'   => 'new_column',
+					'Type'    => 'varchar(255)',
+					'Null'    => 'NO',
+					'Key'     => '',
+					'Default' => '',
+					'Extra'   => '',
+				),
+			),
```
brandonpayton, line 1155: After this change, I think the resulting column order should be:
- id
- new_first_column
- new_column
- name

Am I missing something?

tests/WP_SQLite_Translator_Tests.php
```diff
@@ -1027,6 +1027,279 @@ public function testAlterTableAddNotNullVarcharColumn() {
...
```
brandonpayton, line 1198: After this change, I think the resulting column order should be:
- new_first_column
- new_column
- name
- id

Am I missing something?

tests/WP_SQLite_Translator_Tests.php
```diff
@@ -1027,6 +1027,279 @@ public function testAlterTableAddNotNullVarcharColumn() {
...
+				),
+				(object) array(
+					'Field'   => 'new3',
+					'Type'    => 'varchar(255)',
+					'Null'    => 'NO',
+					'Key'     => '',
+					'Default' => '',
+					'Extra'   => '',
+				),
+			),
```
brandonpayton, line 1252: After this change, shouldn't the resulting column order be the following?
- new2
- id
- new1
- new3

brandonpayton (changes requested):
IIUC, something must not be working properly because the column order in the tests do not appear to be changing.

JanJakes:
@brandonpayton Please, see the description of the PR. As it says, "SQLite doesn't support FIRST/AFTER, so this implementation simply ignores these".

Actually implementing the order would mean changing simple ADD COLUMN rewriting to CHANGE-like copy-table behavior.

brandonpayton (approved):
> @brandonpayton Please, see the description of the PR. As it says, "SQLite doesn't support FIRST/AFTER, so this implementation simply ignores these".

🤦‍♂️  Right. I read that originally but lost the detail while multi-tasking.

These changes look reasonable to me and test well.

PR #149, opened 2024-08-07
Sqlite: Fix LIKE BINARY queries

kozer:
This PR modifies the SQLite database integration plugin ensuring LIKE BINARY queries are supported. 

The translator has been updated to utilize the  `GLOB` SQLite function to support the `LIKE BINARY` clause in Sqlite, ensuring it respects case sensitivity.

### Example query
**MySQL**
```sql
SELECT * FROM _tmp_table WHERE name LIKE BINARY 'f%'
```

**SQLite**
```sql
SELECT * FROM _tmp_table WHERE name GLOB 'f*'
```

### DIFF
--- a/.github/workflows/cs.yml
+++ b/.github/workflows/cs.yml
@@ -1,0 +1,68 @@
+name: CS
+
+on:
+  # Run on all relevant pushes (except to main) and on all relevant pull requests.
+  push:
+    paths:
+      - '**.php'
+      - 'composer.json'
+      - 'composer.lock'
+      - '.phpcs.xml.dist'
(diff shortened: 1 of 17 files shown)


jeroenpf (approved):

jeroenpf:
@aristath - could you have a look at this PR? The LIKE BINARY queries are used in the wp cli search-replace commands so it would be good to include this fix.

aristath (approved):

kozer:
@aristath , can we deploy a new version so we can use it in studio? Is it possible? 
Thank you very much

aristath:
Done.

Issue #148, opened 2024-08-07
Support `ON UPDATE CURRENT_TIMESTAMP` columns


JanJakes:
Unfortunately, `ON UPDATE current_timestamp` doesn't seem to be supported by SQLite.

While for **creating a table** the test following case passes, I don't think it has any effect:

```php
$this->assertQuery(
	'CREATE TABLE _tmp_table (
		id int(11) NOT NULL,
		created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
	);'
);
```

For **altering a table**, the following test case fails with `SQLSTATE[HY000]: General error: 1 near "ON": syntax error.`:

```php
$this->assertQuery(
	'CREATE TABLE _tmp_table (
		id int(11) NOT NULL
	);'
);

$this->assertQuery(
	'ALTER TABLE _tmp_table ADD COLUMN created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP'
);
```

I think the only way to support this would be to [create a trigger](https://stackoverflow.com/a/61763145), but I suppose it would also need to be removed before any other `ALTER TABLE` statement for the same column, as if the statement omits the `ON UPDATE` part, it would need to be removed.

PR #150, opened 2024-08-08
Add support for `ON UPDATE CURRENT_TIMESTAMP` columns

JanJakes:
The `ON UPDATE CURRENT_TIMESTAMP` is a special MySQL-specific construct that can be used in column definitions. There is no such functionality in SQLite. This pull request implements the `ON UPDATE CURRENT_TIMESTAMP` functionality for both `CREATE` and `ALTER` table commands using **triggers**.

With queries such as:
```sql
CREATE TABLE _tmp_table (
	id int(11) NOT NULL,
	created_at timestamp NULL ON UPDATE CURRENT_TIMESTAMP
);

ALTER TABLE _tmp_table ADD COLUMN updated_at timestamp NULL ON UPDATE CURRENT_TIMESTAMP;
```

The following triggers will be added:
```sql
-- created_at
CREATE TRIGGER "___tmp_table_created_at_on_update__"
AFTER UPDATE ON "_tmp_table"
FOR EACH ROW
BEGIN
  UPDATE "_tmp_table" SET "created_at" = CURRENT_TIMESTAMP WHERE id = NEW.id;
END;

-- updated_at
CREATE TRIGGER "___tmp_table_updated_at_on_update__"
AFTER UPDATE ON "_tmp_table"
FOR EACH ROW
BEGIN
  UPDATE "_tmp_table" SET "updated_at" = CURRENT_TIMESTAMP WHERE id = NEW.id;
END;
```

When the `ON UPDATE` is dropped using a query such as the following:
```sql
ALTER TABLE _tmp_table
CHANGE created_at created_at timestamp NULL,
CHANGE COLUMN updated_at updated_at timestamp NULL
```

The triggers are dropped as well. To ensure this always works, the trigger is always dropped when a column is being altered. Then, if `ON UPDATE` is present, it will be recreated; otherwise, it will remain dropped.

Resolves https://github.com/WordPress/sqlite-database-integration/issues/148.

### DIFF
--- a/wp-includes/sqlite/class-wp-sqlite-translator.php
+++ b/wp-includes/sqlite/class-wp-sqlite-translator.php
@@ -923,10 +923,8 @@ class WP_SQLite_Translator {
 			);
 		}
 
-		foreach ( $table->fields as $field ) {
-			if ( $field->on_update ) {
-				$this->add_column_on_update_current_timestamp( $table->name, $field->name );
-			}
+		foreach ( $on_updates as $column => $on_update ) {
+			$this->add_column_on_update_current_timestamp( $table->name, $column );
 		}
 	}
 
--- a/wp-includes/sqlite/class-wp-sqlite-translator.php
+++ b/wp-includes/sqlite/class-wp-sqlite-translator.php
@@ -1120,11 +1120,17 @@ class WP_SQLite_Translator {
 				continue;
 			}
 
-			if ( $token->matches(
-				WP_SQLite_Token::TYPE_KEYWORD,
-				WP_SQLite_Token::FLAG_KEYWORD_RESERVED,
-				array( 'ON UPDATE' )
-			) ) {
+			if (
+				$token->matches(
+					WP_SQLite_Token::TYPE_KEYWORD,
+					WP_SQLite_Token::FLAG_KEYWORD_RESERVED,
+					array( 'ON UPDATE' )
+				) && $this->rewriter->peek()->matches(
+					WP_SQLite_Token::TYPE_KEYWORD,
+					WP_SQLite_Token::FLAG_KEYWORD_RESERVED,
+					array( 'CURRENT_TIMESTAMP' )
+				)
+			) {
 				$this->rewriter->skip();
 				$result->on_update = true;
 				continue;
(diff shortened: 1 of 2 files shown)


wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -917,6 +922,12 @@ private function execute_create_table() {
 				$constraint->value
 			);
 		}
+
+		foreach ( $table->fields as $field ) {
+			if ( $field->on_update ) {
+				$this->add_column_on_update_current_timestamp( $table->name, $field->name );
+			}
+		}
```
bgrgicak, line 930: @JanJakes from what I can see we don't check for `CURRENT_TIMESTAMP` anywhere.
I tried replacing `CURRENT_TIMESTAMP` in tests with `NOT_CURRENT_TIMESTAMP` and they still passed.

Could we run `$this->add_column_on_update_current_timestamp` only if the query includes `ON UPDATE CURRENT_TIMESTAMP`?

JanJakes: @bgrgicak 👍 Good point. There is no other `ON UPDATE SOMETHING` construct for columns in MySQL, but it's better to be more explicit. Updated in https://github.com/WordPress/sqlite-database-integration/pull/150/commits/ea37a9f6ded2ea77fc080df5f72be30bea1416e3.

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -864,6 +864,7 @@ private function execute_create_table() {
 		$table = $this->parse_create_table();
 
 		$definitions = array();
+		$on_updates  = array();
```
bgrgicak, line 867: What does this array do?

Should it be used in the loop on line 926?

JanJakes: @bgrgicak Ahh, good catch! Originally, I was looping the array, but then I realized I can simply loop the fields, and seems like I somehow forgot to completely remove it. Which variant do you think is better? With or without the `$on_updates` array?

bgrgicak: With `on_update` seems better to me.

JanJakes: @bgrgicak Thanks! Addressed in https://github.com/WordPress/sqlite-database-integration/pull/150/commits/50279c4c126bb091ae62abc4c44ead9136a2bc88.

bgrgicak (approved):

bgrgicak:
Thanks for the PR @JanJakes!

PR #151, opened 2024-08-08
Add support for `CURRENT_TIMESTAMP()` calls with parentheses

JanJakes:
While SQLite does support `CURRENT_TIMESTAMP` function calls natively, it doesn't support calling the function with parentheses in the form of `CURRENT_TIMESTAMP()`.

This pull request removes the parentheses after `CURRENT_TIMESTAMP` keyword for all types of queries.

The following types of queries will now work:

```sql
SELECT
    current_timestamp AS t1,
    CURRENT_TIMESTAMP AS t2,
    current_timestamp() AS t3,
    CURRENT_TIMESTAMP() AS t4;

INSERT INTO _dates (option_name, option_value) VALUES ('first', current_timestamp());

UPDATE _dates SET option_value = CURRENT_TIMESTAMP();

DELETE FROM _dates WHERE option_value = CURRENT_TIMESTAMP();
```

Closes https://github.com/WordPress/sqlite-database-integration/pull/129.

### DIFF
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -2859,6 +2859,46 @@ QUERY
 		);
 	}
 
+	public function testCurrentTimestamp() {
+		// SELECT
+		$results = $this->assertQuery(
+			'SELECT
+				current_timestamp AS t1,
+				CURRENT_TIMESTAMP AS t2,
+				current_timestamp() AS t3,
+				CURRENT_TIMESTAMP() AS t4'
+		);
+		$this->assertIsArray( $results );
+		$this->assertCount( 1, $results );
+		$this->assertRegExp( '/\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d/', $results[0]->t1 );
+		$this->assertRegExp( '/\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d/', $results[0]->t2 );
+		$this->assertRegExp( '/\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d/', $results[0]->t3 );
+
+		// INSERT
+		$this->assertQuery(
+			"INSERT INTO _dates (option_name, option_value) VALUES ('first', current_timestamp())"
+		);
+		$results = $this->assertQuery( 'SELECT option_value AS t FROM _dates' );
+		$this->assertCount( 1, $results );
+		$this->assertRegExp( '/\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d/', $results[0]->t );
+
+		// UPDATE
+		$this->assertQuery( 'UPDATE _dates SET option_value = NULL' );
+		$results = $this->assertQuery( 'SELECT option_value AS t FROM _dates' );
+		$this->assertCount( 1, $results );
+		$this->assertEmpty( $results[0]->t );
+
+		$this->assertQuery( 'UPDATE _dates SET option_value = CURRENT_TIMESTAMP()' );
+		$results = $this->assertQuery( 'SELECT option_value AS t FROM _dates' );
+		$this->assertCount( 1, $results );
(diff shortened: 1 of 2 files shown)


tests/WP_SQLite_Translator_Tests.php
```diff
@@ -2859,6 +2859,46 @@ public function testDefaultNullValue() {
...
+		);
+		$this->assertIsArray( $results );
+		$this->assertCount( 1, $results );
+		$this->assertRegExp( '/\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d/', $results[0]->t1 );
+		$this->assertRegExp( '/\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d/', $results[0]->t2 );
+		$this->assertRegExp( '/\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d/', $results[0]->t3 );
+
+		// INSERT
+		$this->assertQuery(
+			"INSERT INTO _dates (option_name, option_value) VALUES ('first', current_timestamp())"
```
bgrgicak, line 2879: NIT to keep the code consistent. 
```suggestion
			"INSERT INTO _dates (option_name, option_value) VALUES ('first', CURRENT_TIMESTAMP())"
```

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -772,7 +772,23 @@ public function get_return_value() {
...
+				$paren_open  = $tokens[ $i + 1 ] ?? null;
+				$paren_close = $tokens[ $i + 2 ] ?? null;
+				if ( WP_SQLite_Token::TYPE_OPERATOR === $paren_open->type && '(' === $paren_open->value
+					&& WP_SQLite_Token::TYPE_OPERATOR === $paren_close->type && ')' === $paren_close->value ) {
+					unset( $tokens[ $i + 1 ], $tokens[ $i + 2 ] );
+				}
+			}
+		}
+		$tokens = array_values( $tokens );
+
```
bgrgicak, line 791: This code works as expected, but this seems like the wrong place to put it.
Ideally, we should clean up the brackets in the Lexer, but I couldn't find a good place to do it there.

@adamziel @aristath do you have any suggestions on where to put this code?

JanJakes: @bgrgicak I tried to move it here: https://github.com/WordPress/sqlite-database-integration/pull/151/commits/43b628d7c512c2d694182f5f0f0821b9f459d84a

Does that make sense?

bgrgicak: This definitely better but it still feels like a "random" piece of code in that function.

I will let @adamziel and @aristath decide on this one as I don't know enough about the codebase to make this decision.

adamziel: Unfortunately, I have the opposite feedback. This Lexer class is a direct port of the [PHPMyAdmin Lexer](https://github.com/phpmyadmin/sql-parser/blob/master/src/Lexer.php) and is meant to tokenize MySQL queries. parentheses are valid in MySQL and removing them in there sounds like an abstraction leak. Not that it matters with the AST Parser/Lexer explorations coming. I'm happy to get this change in as it is as I don't think we'll stick to the current lexer in the longer term, but if you were keen to relocate it then its original location in the Translator class seem fine to me.

JanJakes: @adamziel If you prefer the variant from [the first commit](https://github.com/WordPress/sqlite-database-integration/pull/151/commits/4043e6d5b2c1d6b7eb12450226a9ba5c725391f0), I can just drop the last one and repush, and we can go with that for now. I guess if the new approach replaces the old one, it can be still valuable to add the test case, and we can have the functionality earlier.

adamziel: Yes please and agreed

JanJakes: @adamziel Thanks! Repushed.

adamziel:
Thank you!

PR #152, opened 2024-08-09
Implement altering column with `SET DEFAULT` and `DROP DEFAULT`

JanJakes:
This PR implements the support for the following MySQL statements:

```sql
ALTER TABLE _tmp_table CHANGE name SET DEFAULT 'abc';
ALTER TABLE _tmp_table CHANGE name DROP DEFAULT;
```

It required a small refactor so that the CHANGE COLUMN logic can be extracted to a reusable method.

### DIFF
--- a/wp-includes/sqlite/class-wp-sqlite-translator.php
+++ b/wp-includes/sqlite/class-wp-sqlite-translator.php
@@ -3081,6 +3081,70 @@ class WP_SQLite_Translator {
 				}
 				// We're done.
 				break;
+			} elseif ( 'ALTER' === $op_type ) {
+				$raw_from_name = 'COLUMN' === $op_subject ? $this->rewriter->skip()->token : $op_raw_subject;
+				$from_name     = $this->normalize_column_name( $raw_from_name );
+
+				$set_or_drop_default = $this->rewriter->peek()->matches(
+					WP_SQLite_Token::TYPE_KEYWORD,
+					WP_SQLite_Token::FLAG_KEYWORD_RESERVED,
+					array( 'SET', 'DROP' )
+				) && $this->rewriter->peek_nth( 2 )->matches(
+					WP_SQLite_Token::TYPE_KEYWORD,
+					WP_SQLite_Token::FLAG_KEYWORD_RESERVED,
+					array( 'DEFAULT' )
+				);
+
+				// Handle "CHANGE <column> DROP DEFAULT" and "CHANGE <column> SET DEFAULT <value>".
+				if ( $set_or_drop_default ) {
+					$this->execute_change(
+						function ( $old_name, $old_column ) use ( $from_name ) {
+							//$old_column->consume_all();
+							//var_dump($old_column->get_updated_query());ob_flush();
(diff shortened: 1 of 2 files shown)


wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3094,8 +3109,21 @@ private function execute_alter() {
...
-						|| $from_name !== $this->normalize_column_name( $token->value ) ) {
+
+					// Column name can be represented as:
+					//   1. "column_name" (string)
+					//   2. `column_name` (symbol)
+					//   3. column_name   (keyword — not a reserved one)
+					// While CREATE TABLE normalizes column names to strings, it's better to handle all
+					// variants to handle table definitions with names stored in any valid form.
+					$is_column_name = WP_SQLite_Token::TYPE_STRING === $token->type
+						|| WP_SQLite_Token::TYPE_SYMBOL === $token->type
```
brandonpayton, line 3120: @JanJakes, I think this change should probably be in a separate PR.

AFAICT, it is unrelated to the main purpose of this PR, and a separate PR would allow us to review the codebase and address other places where we could treat names more thoroughly. Maybe there could be a new helper method to check correctly wherever it is needed.

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3053,16 +3053,31 @@ private function execute_alter() {
...
 				// Parse the new column definition.
-				$raw_from_name    = 'COLUMN' === $op_subject ? $this->rewriter->skip()->token : $op_raw_subject;
-				$from_name        = $this->normalize_column_name( $raw_from_name );
-				$new_field        = $this->parse_mysql_create_table_field();
-				$alter_terminator = end( $this->rewriter->output_tokens );
-				$this->update_data_type_cache(
-					$this->table_name,
-					$new_field->name,
-					$new_field->mysql_data_type
+				$raw_from_name = 'COLUMN' === $op_subject ? $this->rewriter->skip()->token : $op_raw_subject;
```
brandonpayton, line 3056: Is `ALTER TABLE <table-name> CHANGE COLUMN <column-name> SET DEFAULT <value>` valid MySQL syntax?

Based on the docs [here](https://dev.mysql.com/doc/refman/8.4/en/alter-table.html), it looks like the syntax should be:
```sql
ALTER TABLE table_name ALTER COLUMN column_name SET DEFAULT value;
```

And this seems to be correct in some testing:
```
MariaDB [149697504]> ALTER TABLE bptest CHANGE COLUMN stuff SET DEFAULT 777;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'SET DEFAULT 777' at line 1
MariaDB [149697504]> ALTER TABLE bptest ALTER COLUMN stuff SET DEFAULT 777;
Query OK, 0 rows affected (0.002 sec)
Records: 0  Duplicates: 0  Warnings: 0
```

adamziel:
Superseded by #179

Issue #106, opened 2024-04-30
Support full SELECT syntax


adamziel:
Let's go through the MySQL documentation pages and make sure even the complex SELECT queries are supported by the SQLite integration plugin. This likely means rewriting `execute_select` as a state machine that reason about each token and can also be reused by `execute_insert_or_replace` when a subquery is encountered. 

* https://dev.mysql.com/doc/refman/8.0/en/select.html
* https://dev.mysql.com/doc/refman/8.0/en/join.html
* https://dev.mysql.com/doc/refman/8.0/en/union.html
* https://dev.mysql.com/doc/refman/8.0/en/intersect.html
* https://dev.mysql.com/doc/refman/8.0/en/except.html
* https://dev.mysql.com/doc/refman/8.0/en/with.html

For starters, just migrating to a state machine approach would be more than enough as it would unlock support for UNIONs and easy ignoring of tokens like `HIGH_PRIORITY` or `SQL_SMALL_RESULT`.

adamziel:
Perhaps the Rewriter model is no longer useful for this approach. Here's an alternative idea:

```php
/**
 * Processes the SELECT statement (https://dev.mysql.com/doc/refman/8.0/en/select.html)
 *    SELECT
 *       [ALL | DISTINCT | DISTINCTROW ]
 *       [HIGH_PRIORITY]
 *       [STRAIGHT_JOIN]
 *       [SQL_SMALL_RESULT] [SQL_BIG_RESULT] [SQL_BUFFER_RESULT]
 *       [SQL_NO_CACHE] [SQL_CALC_FOUND_ROWS]
 *        select_expr [, select_expr] ...
 */
protected function consume_select_query() {
    $token = $this->next_token();
    $token->assert_is_keyword('SELECT');

    $token = $this->next_token();
    if($token->is_keyword(['ALL', 'DISTINCT', 'DISTINCTROW'])) {
         $this->append( $token );
         $token = $this->rewriter->next_token();
    }
    if($token->is_keyword('HIGH_PRIORITY')) {
         $token = $this->rewriter->next_token();
    }
    // ... keep going token by token, don't just skip over things like we do now
    //     with a while loop ...
    if($is_subquery) {
        $this->consume_sub_query();
    }
    // inevitably at some point:
    if($token->is_keyword('UNION')) {
        $this->consume_select_query();
    }
}
```

[...]

PR #153, opened 2024-08-11
Explorations: MySQL Query -> AST Parser

adamziel:
> [!NOTE]  
> Read the [MySQL parser proposal](https://github.com/WordPress/sqlite-database-integration/issues/106#issuecomment-2276082489) for the full context on this PR.

## Description

This PR explores a full MySQL Lexer and Parser to enable an AST-based MySQL -> SQLite, PostgreSQL, etc. bridge. This would be much more stable, easy to maintain, and easy to expand than token processing approach we use now.

The proposed `MySQLLexer.php` and `MySQLParser.php` have plenty of opportunities for optimization and refactoring, but overall they are fantastic starting points for this work and should save us a few months of explorations. The entire code is ~1MB (or 100kb gzipped), has no dependencies, and can parse 500 complex SELECT queries in ~800ms. I think we can reduce that time 10x or so.

## How was it done?

I fed the official [MySQLParser.g4](https://github.com/mysql/mysql-workbench/blob/8.0/library/parsers/grammars/MySQLParser.g4) grammar to Google Gemini Pro and asked it to build a PHP parsed based on that. The 2 million token input context window made it really viable. Gemini can only output ~8000 tokens, so I had to take the response and feed it back to the model as an input. It took some time, so I ran a loop overnight and in the morning I had a decent starting point.

I used the `generate_parser.py` script included in this PR to generate the code. I built that script with [AI studio](https://aistudio.google.com/app/) where I initially uploaded the MySQL grammar, tuned the model parameters, and perfected the prompt. After the export I only had to make a few adjustments such as adding a loop, local file cache etc.

Once the parser was ready, I then used the lexing grammar and a similar prompt to generate the Lexer. Yes, the Lexer came second. I asked Gemini to make the lexer class plug-and-play with the parser class. However, once the Lexer was done, I realized I used the wrong Parser grammar and regenerated the entire Parser from scratch, only this time I included the Lexer class with my prompt for reference.

The entire project costed ~$520 in Google Cloud charges, $300 out of which I covered using free trial credits. 

## Rationale

As [the proposal explains](https://github.com/WordPress/sqlite-database-integration/issues/106#issuecomment-2276082489):

* processing AST is much easier than processing tokens
* No existing parser available to us can correctly parse the MySQL syntax

## Testing instructions

Run PHPUnit tests as follows:

```shell
phpunit -c ./phpunit.xml.dist  --filter WP_MySQL_Parser_Tests
```

## Next steps

* Discuss the idea, look for any blockers
- [x] Ask Gemini to implement all the missing methods
* Refactor the Lexer to avoid all the extra work it does on lookaheads
* Evaluate switching from PHP Arrays to PHP Objects where applicable (huge speed gains may be possible)
* Bring in a solid test suite from one of MySQL projects to ensure we can parse all kinds of queries
* If everything went well, bring this parser into the project behind a `try/catch` statement, try to use it by default and fall back to the existing approach on failure. Collect logs when users opted-in to that. Once we achieve a feature and stability parity, switch to this parser entirely.

cc @dmsnell @aristath @brandonpayton @schlessera


JanJakes:
@adamziel Wow, this is an incredible demonstration of what's possible with current AI tools! Intuitively, I would've never guessed that it could do such a great job.

I'll add one question that's on my mind: Why not apply this to a language that can be compiled to WASM, such as Rust, for instance? How would that affect performance and bundle size? Would there be a binding cost (since PHP is WASM as well)?

Additionally, would new builds for new releases of MySQL be created similarly, or would the parser require a manual maintenance from now on?

adamziel:
Thank you @JanJakes, and great questions! 

> I'll add one question that's on my mind: Why not apply this to a language that can be compiled to WASM, such as Rust, for instance? How would that affect performance and bundle size? Would there be a binding cost (since PHP is WASM as well)

For Playground, that would be brilliant. For WordPress core, that wouldn't help use SQLite at all since PHP has no WASM support yet. I think we may eventually maintain a semi-automatically-maintained C or Rust implementation to speed things up.

> Additionally, would new builds for new releases of MySQL be created similarly, or would the parser require a manual maintenance from now on?

The ANTLR .g4 grammar doesn't seem to be maintained. We'd either need to find an up-to-data and well-maintained MariaDB grammar file and figure out a semi-automated workflow, or, if we can't find one, do manual maintenance.

adamziel:
As I'm working through this parser, there's quite a few places requiring small adjustments in the condition structure, or reordering a few if/else branches, or looking ahead a few tokens more than what Gemini did. It can all be done with a week or two, but I now wonder whether converting `MySQLParser.g4` to another format would enable generating a parser that correctly handles all these nuances off the bat. I'll explore that.

wp-content/plugins/sqlite-database-integration/wp-includes/mysql-parser/MySQLLexer.php
```diff
@@ -0,0 +1,9061 @@
...
+    public const PARAM_MARKER = 38;
+    // Data types
+    public const INT_SYMBOL = 39;
+    public const TINYINT_SYMBOL = 40;
+    public const SMALLINT_SYMBOL = 41;
+    public const MEDIUMINT_SYMBOL = 42;
+    public const BIGINT_SYMBOL = 43;
+    public const REAL_SYMBOL = 44;
+    public const DOUBLE_SYMBOL = 45;
+    public const FLOAT_SYMBOL    = 46;
```
dmsnell, line 81: next time tell the AI to run its output through WPCS 😆

wp-content/plugins/sqlite-database-integration/wp-includes/mysql-parser/MySQLLexer.php
```diff
@@ -0,0 +1,9061 @@
...
+    public const ASENSITIVE_SYMBOL = 94;
+    public const AT_SYMBOL = 95;
+    public const AUTOEXTEND_SIZE_SYMBOL = 96;
+    public const AUTO_INCREMENT_SYMBOL = 97;
+    public const AVG_ROW_LENGTH_SYMBOL = 98;
+    public const AVG_SYMBOL = 99;
+    public const BACKUP_SYMBOL = 100;
+    public const BEFORE_SYMBOL = 101;
+    public const BEGIN_SYMBOL = 102;
+    public const BETWEEN_SYMBOL = 103;
```
dmsnell, line 139: is 104 skipped in the grammar?

wp-content/plugins/sqlite-database-integration/wp-includes/mysql-parser/MySQLLexer.php
```diff
@@ -0,0 +1,9061 @@
...
+    {
+        $this->input = $input;
+        $this->serverVersion = $serverVersion;
+        $this->sqlModes = $sqlModes;
+    }
+
+    const PipesAsConcat = 1;
+    const HighNotPrecedence = 2;
+    const NoBackslashEscapes = 4;
+    public const ANSI_QUOTES = 8;
```
dmsnell, line 880: this is an interesting bit. it appears like this was arbitrarily chosen as the only escaping mode? given an arbitrary enum value?

dmsnell (commented):
Crazy idea: ask it to replace all `const` values with string literals and see how that impacts the built size.

adamziel:
I'll close this PR in favor of https://github.com/WordPress/sqlite-database-integration/pull/157.

PR #154, opened 2024-08-12
Fix the behavior of `HAVING` clause

JanJakes:
The current implementation has a special handling for `HAVING` clauses without `GROUP BY` (this is [valid in MySQL](https://stackoverflow.com/questions/6924896/having-without-group-by) in some SQL modes). This implementation takes advantage of the equivalence of ungrouped HAVING with WHERE and rewrites such a `HAVING <condition>` clause to `AND <condition>`.

However, the current implementation has some issues:
1. The presence of `GROUP BY` in a query is incorrectly detected. The detection always fails, and thus every HAVING query is considered to be ungrouped. I added a fix in the first commit.
2. Rewriting `HAVING <condition>` to `AND <condition>` fails with an error `near "AND": syntax error` when no WHERE is included in the query. This is fixed in the second commit.
3. When an aggregate function is used in HAVING (e.g, `COUNT(*) > 1`), the query fails with: `misuse of aggregate function COUNT()`. This is because aggregate functions can't be used in the WHERE clause. This is fixed in the second commit.

To fix all of these issues, I implemented a fix for the `GROUP BY` detection (case 1), and changed the rewriting of `HAVING <condition>`  without `GROUP BY` to `GROUP BY 1 HAVING <condition>` (which fixes both case 2 and 3).

### DIFF
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -3077,6 +3077,56 @@ QUERY
 		);
 	}
 
+	public function testHavingWithoutGroupBy() {
+		$this->assertQuery(
+			'CREATE TABLE _tmp_table (
+				name varchar(20)
+			);'
+		);
+
+		$this->assertQuery(
+			"INSERT INTO _tmp_table VALUES ('a'), ('b'), ('b'), ('c'), ('c'), ('c')"
+		);
+
+		// HAVING condition satisfied
+		$result = $this->assertQuery(
+			"SELECT 'T' FROM _tmp_table HAVING COUNT(*) > 1"
+		);
+		$this->assertEquals(
+			array(
+				(object) array(
+					':param0' => 'T',
+				),
+			),
+			$result
+		);
+
+		// HAVING condition not satisfied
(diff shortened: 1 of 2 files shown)


brandonpayton (approved):
@JanJakes, this looks good to me, but I am troubled that our previous test set didn't catch that GROUP BY capture was broken. Before merging this, could we also add a dedicated test for GROUP BY without HAVING?

JanJakes:
@brandonpayton Thanks for checking this!

GROUP BY capture was indeed broken, but not its execution. The capture itself is used only for the HAVING clause.

When you use GROUP BY without HAVING, then the value of `has_group_by` is not used, the only reason to capture that variable is the HAVING clause. Additionally, the GROUP BY statement itself is never translated to anything. That means that GROUP BY without HAVING works well, and I'm afraid the detection is not really testable with only a GROUP BY statement (I could check the private property `has_group_by`, I suppose, but then I'm not testing the public API).

The way to test this via the public API is GROUP BY with HAVING, and I guess I added some tests for that.

brandonpayton:
@JanJakes Thanks for the background and explanation. Let's merge this.

PR #156, opened 2024-08-16
Fix & improve `ON UPDATE` trigger behavior

JanJakes:
This just fixes and improves https://github.com/WordPress/sqlite-database-integration/pull/150.

With some further testing, I realized there were some issues in my original implementation:
1. The `ON UPDATE` trigger uses `id` instead of `rowid`. This is a regression — a query can fail now for anyone who uses `ON UPDATE` and doesn't have an `id` column. I somehow missed that 🤦‍♂️
2. The trigger is deleted with every column operation, not only for column `CHANGE`s, and it is actually missed by the `CHANGE` for creation as well, as that branch uses `continue`/`break` and skips the trigger creation below.

I fixed both and added test coverage.


brandonpayton (approved):
This reads well and tests well. Thank you, @JanJakes!

PR #158, opened 2024-08-20
Fix creating a table with `AUTOINCREMENT` and `ON UPDATE` followed by `PRIMARY KEY`

JanJakes:
This reverts e59eff8c2856a5c96fc0db5da1402d5a258c8aa8. The fix was wrong. When spotting it, I misunderstood that the first token was already consumed.

Without the fix, the new test fails with:
```
Cannot combine AUTOINCREMENT and multiple primary keys in SQLite.
```

This is because we've consumed one extra token (`,`), and the `PRIMARY KEY` then gets assigned to the `created_at` column as well.


brandonpayton (approved):
Thanks for catching this and providing a fix, @JanJakes!

Issue #162, opened 2024-11-04
Advanced MySQL compatibility


JanJakes:
This is a tracking issue for introducing a new MySQL query **parser** and an AST-based SQLite **driver** to significantly improve MySQL compatibility and expand WordPress plugin support.

I think the project can be represented in roughly the following phases:
1. Initial version of the MySQL parser — reasonably fast, small, and correct _("make it work")_.
2. Initial version of the SQLite driver — reach parity with the current test suite _("make it work")_.
3. Advanced MySQL support in the SQLite driver — see also test suites below _("make it right")_.
4. Advanced MySQL parser — add version support, verify against MySQL server grammar, fix edge cases _("make it right")_.
5. Advanced tooling — implement a custom parser for ANTLR or ENBF, add custom grammar tooling _("make it right")_.
6. Performance optimizations — explore possibilities to further optimize the new parser _("make it fast")_.

Here's a list of tasks and issues:

### Core
- [x] [Exhaustive MySQL Parser](https://github.com/Automattic/sqlite-database-integration/commit/ac75e90e052ad70b40729cb7b4bdc0c691fc6b52)
- [x] https://github.com/Automattic/sqlite-database-integration/pull/1 — match current MySQL compatibility
- [ ] SQLite driver — improve compatibility to satisfy multiple extensive test suites (see below)

### Test suites
- [ ] SQL errors when activating most popular WP plugins: [Playground Tester](https://github.com/bgrgicak/playground-tester) → [SQL errors](https://github.com/bgrgicak/playground-tester/blob/main/reports/sql-errors.md)
- [ ] WP unit tests: https://github.com/WordPress/wordpress-playground/issues/111#issuecomment-2434885741
- [ ] Statically extracted queries from WP plugins: https://github.com/WordPress/sqlite-database-integration/issues/159
- [ ] Supporting all WP CLI commands: https://github.com/WordPress/wordpress-playground/issues/1618#issuecomment-2380150978

### Advanced parser and tooling
- [ ] MySQL server version support in the new parser
- [ ] A generic grammar parser (ANTLR or https://github.com/WordPress/wordpress-playground/issues/1981)

### Supported MySQL constructs

**CREATE TABLE**
- [x] `CREATE TABLE t (id INT)`
- [x] `CREATE TABLE t (id INT NOT NULL)`
- [x] `CREATE TABLE t (id INT) DEFAULT 0`
- [x] `CREATE TABLE t (...) ENGINE=InnoDB`
[...]

PR #157, opened 2024-08-17
Exhaustive MySQL Parser 

adamziel:
## Context
This PR ships an exhaustive MySQL **lexer** and **parser** that produce a MySQL query AST. This is the first step to significantly improve MySQL compatibility and expand WordPress plugin support on SQLite. It's an easier, more stable, and an easier to maintain method than the current token processing. It will also dramatically improve WordPress Playground experience – database integration is the single largest source of issues.

This PR is part of the [Advanced MySQL support project](https://github.com/WordPress/sqlite-database-integration/issues/162).

See the [MySQL parser proposal](https://github.com/WordPress/sqlite-database-integration/issues/106#issuecomment-2276082489) for additional context.

## This PR ships
1. A **MySQL lexer**, adapted from the AI-generated one by @adamziel. It's over 3x smaller and close to 2x faster.
2. A **MySQL grammar** written in ANTLR v4 format, adapted from the [MySQL Workbench grammar](https://github.com/mysql/mysql-workbench/blob/8.0.38/library/parsers/grammars/MySQLParser.g4) by adding and fixing some cases and reordering some rules.
3. A **script to factor, convert, and compress the grammar** to a PHP array.
4. A **dynamic recursive parser** implemented by @adamziel.
5. A **script to extract tests** from the MySQL repository.
6. A **test suite of almost 70k queries**.
7. WIP **SQLite driver** by @adamziel, a demo and foundation for the next phase.

At the moment, all the new files are omitted from the plugin build, so they have no effect on production whatsoever.

## Running tests
The lexer & parser tests suite is not yet integrated into the CI and existing test commands. To run the tests, use:
```php
php tests/parser/run-lexer-tests.php
php tests/parser/run-parser-tests.php
```
This will lex / lex & parse all the ~70k queries.

### Parser

A simple recursive parser to transform `(token stream, grammar) => parse tree`. In this PR, we use MySQL tokens and MySQL grammar, but the same parser could also support XML, IMAP, many other grammars (as long as they have some specific properties).

The `parse_recursive()` method is just 100 lines of code (excluding comments). All of the parsing rules are provided by the grammar.

### run-mysql-driver.php

A quick and dirty implementation of what a `MySQL parse tree ➔ SQLite` database driver could look like. It easily supports `WITH` and `UNION` queries that would be really difficult to implement the current SQLite integration plugin.

[...]

Issue #146, opened 2024-08-07
Support for `information_schema` tables


JanJakes:
Queries to `information_schema` appear to succeed without errors, but they may not be returning correct results. I found out that [there seems to be some rewriting done](https://github.com/WordPress/sqlite-database-integration/blob/25610002ef71ea0822b1f46dbdfd592b795ffa0d/wp-includes/sqlite/class-wp-sqlite-translator.php#L2668), but my guess is it may only be handling a few cases.

Checking for whether a table, column, or an index exist seems to return wrong result.

Here are the failing test cases:

```php
public function testTableExists() {
	$this->assertQuery( 'CREATE TABLE test (id INT)' );
	$return = $this->assertQuery( "SELECT 1 FROM information_schema.columns WHERE table_schema = DATABASE() AND table_name = 'test'" );
	$this->assertEquals( [(object) ["1" => "1"]], $return );
}

public function testColumnExists() {
	$this->assertQuery( 'CREATE TABLE test (id INT)' );
	$return = $this->assertQuery( "SELECT 1 FROM information_schema.columns WHERE table_schema = DATABASE() AND table_name = 'test' AND column_name = 'id'" );
	$this->assertEquals( [(object) ["1" => "1"]], $return );
}

public function testIndexExists() {
	$this->assertQuery( 'CREATE TABLE test (id INT)' );
	$this->assertQuery( 'ALTER TABLE test ADD INDEX test_id_idx (id)' );
	$return = $this->assertQuery( "SELECT 1 FROM information_schema.statistics WHERE table_schema = DATABASE() AND table_name = 'test' AND index_name = 'test_id_idx'" );
	$this->assertEquals( [(object) ["1" => "1"]], $return );
}
```

What is the status of `information_schema` support? Is it technically possible to rewrite all kinds of information schema queries, or would an easier approach be to create and manage the tables?

JanJakes:
This was implemented in trunk and will be included in the 3.0 release.

Issue #78, opened 2024-03-20
Support SELECT FROM information_schema


adamziel:
Related to https://github.com/WordPress/sqlite-database-integration/issues/75

JanJakes:
This was implemented in trunk and will be included in the 3.0 release.

PR #164, opened 2024-11-14
AST-based SQLite driver

JanJakes:
This PR is part of the [Advanced MySQL support project](https://github.com/WordPress/sqlite-database-integration/issues/162) and builds on the previous work in [Exhaustive MySQL parser](https://github.com/WordPress/sqlite-database-integration/pull/157).

This is a WIP, currently stacked on top of https://github.com/WordPress/sqlite-database-integration/pull/157.

### DIFF
--- a/tests/parser/WP_Parser_Node_Tests.php
+++ b/tests/parser/WP_Parser_Node_Tests.php
@@ -1,0 +1,144 @@
+<?php
+
+require_once __DIR__ . '/../../wp-includes/parser/class-wp-parser-node.php';
+
+use PHPUnit\Framework\TestCase;
+
+class WP_Parser_Node_Tests extends TestCase {
+	public function testEmptyChildren(): void {
+		$node = new WP_Parser_Node( 1, 'root' );
+
+		$this->assertFalse( $node->has_child() );
+		$this->assertFalse( $node->has_child_node() );
+		$this->assertFalse( $node->has_child_token() );
+
+		$this->assertNull( $node->get_child() );
+		$this->assertNull( $node->get_child_node() );
+		$this->assertNull( $node->get_child_node( 'root' ) );
+		$this->assertNull( $node->get_child_token() );
+		$this->assertNull( $node->get_child_token( 1 ) );
+
+		$this->assertNull( $node->get_descendant_node() );
+		$this->assertNull( $node->get_descendant_token() );
+
+		$this->assertEmpty( $node->get_children() );
+		$this->assertEmpty( $node->get_child_nodes() );
+		$this->assertEmpty( $node->get_child_nodes( 'root' ) );
+		$this->assertEmpty( $node->get_child_tokens() );
+		$this->assertEmpty( $node->get_child_tokens( 1 ) );
+
+		$this->assertEmpty( $node->get_descendants() );
+		$this->assertEmpty( $node->get_descendant_nodes() );
+		$this->assertEmpty( $node->get_descendant_nodes( 'root' ) );
+		$this->assertEmpty( $node->get_descendant_tokens() );
+		$this->assertEmpty( $node->get_descendant_tokens( 1 ) );
+	}
+
+	public function testNodeTree(): void {
+		// Prepare nodes and tokens.
+		$root      = new WP_Parser_Node( 1, 'root' );
+		$n_keyword = new WP_Parser_Node( 2, 'keyword' );
+		$n_expr_a  = new WP_Parser_Node( 3, 'expr' );
+		$n_expr_b  = new WP_Parser_Node( 3, 'expr' );
+		$n_expr_c  = new WP_Parser_Node( 3, 'expr' );
+		$t_select  = new WP_Parser_Token( 100, 'SELECT' );
+		$t_comma   = new WP_Parser_Token( 200, ',' );
+		$t_plus    = new WP_Parser_Token( 300, '+' );
+		$t_one     = new WP_Parser_Token( 400, '1' );
+		$t_two_a   = new WP_Parser_Token( 400, '2' );
+		$t_two_b   = new WP_Parser_Token( 400, '2' );
+		$t_eof     = new WP_Parser_Token( 500, '' );
+
+		// Prepare a tree.
+		//
+		// A simplified testing tree for an input like "SELECT 1 + 2, 2".
+		//
+		// root
+		//   |- keyword
+		//   |    |- "SELECT"
+		//   |- expr [a]
+		//   |    |- "1"
+		//   |    |- "+"
+		//   |    |- expr [c]
+		//   |    |    |- "2" [b]
+		//   |- ","
+		//   |- expr [b]
+		//   |    |- "2" [a]
+		//   |- EOF
+		$root->append_child( $n_keyword );
+		$root->append_child( $n_expr_a );
+		$root->append_child( $t_comma );
+		$root->append_child( $n_expr_b );
+		$root->append_child( $t_eof );
+
+		$n_keyword->append_child( $t_select );
+		$n_expr_a->append_child( $t_one );
+		$n_expr_a->append_child( $t_plus );
+		$n_expr_a->append_child( $n_expr_c );
+		$n_expr_b->append_child( $t_two_a );
+		$n_expr_c->append_child( $t_two_b );
+
+		// Test "has" methods.
+		$this->assertTrue( $root->has_child() );
+		$this->assertTrue( $root->has_child_node() );
+		$this->assertTrue( $root->has_child_token() );
+
+		// Test single child methods.
+		$this->assertSame( $n_keyword, $root->get_child() );
+		$this->assertSame( $n_keyword, $root->get_child_node() );
+		$this->assertSame( $n_keyword, $root->get_child_node( 'keyword' ) );
+		$this->assertSame( $n_expr_a, $root->get_child_node( 'expr' ) );
+		$this->assertSame( $t_comma, $root->get_child_token() );
+		$this->assertSame( $t_comma, $root->get_child_token( 200 ) );
+		$this->assertNull( $root->get_child_token( 100 ) );
+
+		// Test multiple children methods.
+		$this->assertSame( array( $n_keyword, $n_expr_a, $t_comma, $n_expr_b, $t_eof ), $root->get_children() );
+		$this->assertSame( array( $n_keyword, $n_expr_a, $n_expr_b ), $root->get_child_nodes() );
+		$this->assertSame( array( $n_expr_a, $n_expr_b ), $root->get_child_nodes( 'expr' ) );
+		$this->assertSame( array(), $root->get_child_nodes( 'root' ) );
+		$this->assertSame( array( $t_comma, $t_eof ), $root->get_child_tokens() );
+		$this->assertSame( array( $t_comma ), $root->get_child_tokens( 200 ) );
+		$this->assertSame( array(), $root->get_child_tokens( 100 ) );
+
+		// Test single descendant methods.
+		// @TODO: Consider breadth-first search vs depth-first search.
+		$this->assertSame( $n_keyword, $root->get_descendant_node() );
+		$this->assertSame( $n_expr_a, $root->get_descendant_node( 'expr' ) );
+		$this->assertSame( null, $root->get_descendant_node( 'root' ) );
+		$this->assertSame( $t_comma, $root->get_descendant_token() );
+		$this->assertSame( $t_one, $root->get_descendant_token( 400 ) );
+		$this->assertSame( null, $root->get_descendant_token( 123 ) );
+
+		// Test multiple descendant methods.
+		// @TODO: Consider breadth-first search vs depth-first search.
+		$this->assertSame(
+			array( $n_keyword, $n_expr_a, $t_comma, $n_expr_b, $t_eof, $t_select, $t_one, $t_plus, $n_expr_c, $t_two_a, $t_two_b ),
+			$root->get_descendants()
+		);
+		$this->assertSame(
+			array( $n_keyword, $n_expr_a, $n_expr_b, $n_expr_c ),
+			$root->get_descendant_nodes()
+		);
+		$this->assertSame(
+			array( $n_expr_a, $n_expr_b, $n_expr_c ),
+			$root->get_descendant_nodes( 'expr' )
+		);
+		$this->assertSame(
+			array(),
+			$root->get_descendant_nodes( 'root' )
+		);
+		$this->assertSame(
+			array( $t_comma, $t_eof, $t_select, $t_one, $t_plus, $t_two_a, $t_two_b ),
+			$root->get_descendant_tokens()
+		);
+		$this->assertSame(
+			array( $t_one, $t_two_a, $t_two_b ),
+			$root->get_descendant_tokens( 400 )
+		);
+		$this->assertSame(
+			array(),
+			$root->get_descendant_tokens( 123 )
+		);
+	}
+}
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -995,8 +995,115 @@ class WP_SQLite_Driver {
 	}
 
 	private function execute_alter_table_statement( WP_Parser_Node $node ): void {
-		$table_name = $this->translate( $node->get_descendant_node( 'tableRef' ) );
-		$actions    = $node->get_descendant_nodes( 'alterListItem' );
+		$table_name = $this->unquote_sqlite_identifier(
+			$this->translate( $node->get_descendant_node( 'tableRef' ) )
+		);
+
+		// Save all column names from the original table.
+		$column_names = $this->execute_sqlite_query(
+			'SELECT COLUMN_NAME FROM _mysql_information_schema_columns WHERE table_schema = ? AND table_name = ?',
+			array( $this->db_name, $table_name )
+		)->fetchAll( PDO::FETCH_COLUMN );
+
+		// Track column renames and removals.
+		$column_map = array_combine( $column_names, $column_names );
+		foreach ( $node->get_descendant_nodes( 'alterListItem' ) as $action ) {
+			$first_token = $action->get_child_token();
+
+			if ( WP_MySQL_Lexer::DROP_SYMBOL === $first_token->id ) {
+				$name = $this->translate( $action->get_child_node( 'columnInternalRef' ) );
+				if ( null !== $name ) {
+					$name = $this->unquote_sqlite_identifier( $name );
+					unset( $column_map[ $name ] );
+				}
+			}
+
+			if ( WP_MySQL_Lexer::CHANGE_SYMBOL === $first_token->id ) {
+				$old_name = $this->unquote_sqlite_identifier(
+					$this->translate( $action->get_child_node( 'columnInternalRef' ) )
+				);
+				$new_name = $this->unquote_sqlite_identifier(
+					$this->translate( $action->get_child_node( 'identifier' ) )
+				);
+
+				$column_map[ $old_name ] = $new_name;
+			}
+
+			if ( WP_MySQL_Lexer::RENAME_SYMBOL === $first_token->id ) {
+				$column_ref = $action->get_child_node( 'columnInternalRef' );
+				if ( null !== $column_ref ) {
+					$old_name = $this->unquote_sqlite_identifier(
+						$this->translate( $column_ref )
+					);
+					$new_name = $this->unquote_sqlite_identifier(
+						$this->translate( $action->get_child_node( 'identifier' ) )
+					);
+
+					$column_map[ $old_name ] = $new_name;
+				}
+			}
+		}
+
+		$this->information_schema_builder->record_alter_table( $node );
+
+		/*
+		 * See:
+		 *   https://www.sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes
+		 */
+
+		// 1. If foreign key constraints are enabled, disable them.
+		// @TODO
+
+		// 2. Create a new table with the new schema.
+		$tmp_table_name     = "_tmp__{$table_name}_" . uniqid();
+		$queries            = $this->get_sqlite_create_table_statement( $table_name, $tmp_table_name );
+		$create_table_query = $queries[0];
+		$constraint_queries = array_slice( $queries, 1 );
+		$this->execute_sqlite_query( $create_table_query );
+
+		// 3. Copy data from the original table to the new table.
+		$this->execute_sqlite_query(
+			sprintf(
+				'INSERT INTO "%s" (%s) SELECT %s FROM "%s"',
+				$tmp_table_name,
+				implode(
+					', ',
+					array_map(
+						function ( $column ) {
+							return '"' . $column . '"';
+						},
+						$column_map
+					)
+				),
+				implode(
+					', ',
+					array_map(
+						function ( $column ) {
+							return '"' . $column . '"';
+						},
+						array_keys( $column_map )
+					)
+				),
+				$table_name
+			)
+		);
+
+		// 4. Drop the original table.
+		$this->execute_sqlite_query( sprintf( 'DROP TABLE "%s"', $table_name ) );
+
+		// 5. Rename the new table to the original table name.
+		$this->execute_sqlite_query( sprintf( 'ALTER TABLE "%s" RENAME TO "%s"', $tmp_table_name, $table_name ) );
+
+		// 6. Reconstruct indexes, triggers, and views.
+		foreach ( $constraint_queries as $query ) {
+			$this->execute_sqlite_query( $query );
+		}
+
+		// @TODO: Triggers and views.
+
+		$this->results      = 1;
+		$this->return_value = $this->results;
+		return;
 
 		/*
 		 * SQLite supports only a small subset of MySQL ALTER TABLE statement.
@@ -1018,42 +1125,6 @@ class WP_SQLite_Driver {
 		 *
 		 *    @TODO: Address these nuances.
 		 */
-		foreach ( $actions as $action ) {
-			$token = $action->get_child_token();
-
-			// ADD column/constraint.
-			if ( WP_MySQL_Lexer::ADD_SYMBOL === $token->id ) {
-				// ADD COLUMN.
-				$field_definition = $action->get_descendant_node( 'fieldDefinition' );
-				if ( null !== $field_definition ) {
-					$field_name = $this->translate( $action->get_child_node( 'identifier' ) );
-					$field      = $this->translate( $field_definition );
-					$this->execute_sqlite_query(
-						sprintf( 'ALTER TABLE %s ADD COLUMN %s %s', $table_name, $field_name, $field )
-					);
-				}
-
-				// ADD CONSTRAINT.
-				$constraint = $action->get_descendant_node( 'tableConstraintDef' );
-				if ( null !== $constraint ) {
-					$constraint_name = $this->translate( $constraint->get_child_node( 'identifier' ) );
-					$constraint      = $this->translate( $constraint );
-					$this->execute_sqlite_query(
-						sprintf( 'ALTER TABLE %s ADD CONSTRAINT %s %s', $table_name, $constraint_name, $constraint )
-					);
-				}
-			} elseif ( WP_MySQL_Lexer::DROP_SYMBOL === $token->id ) {
-				// DROP COLUMN.
-				$field_name = $this->translate( $action->get_child_node( 'columnInternalRef' ) );
-				if ( null !== $field_name ) {
-					$this->execute_sqlite_query(
-						sprintf( 'ALTER TABLE %s DROP COLUMN %s', $table_name, $field_name )
-					);
-				}
-			}
-		}
-
-		$this->set_result_from_affected_rows();
 	}
 
 	private function translate( $ast ) {
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -1317,6 +1317,8 @@ class WP_SQLite_Driver {
 				$token = $ast->get_child_token();
 				if ( WP_MySQL_Lexer::LIKE_SYMBOL === $token->id ) {
 					return $this->translate_like( $ast );
+				} elseif ( WP_MySQL_Lexer::REGEXP_SYMBOL === $token->id ) {
+					return $this->translate_regexp_functions( $ast );
 				}
 				return $this->translate_sequence( $ast->get_children() );
 			case 'systemVariable':
@@ -1414,6 +1416,29 @@ class WP_SQLite_Driver {
 		 *   https://www.sqlite.org/lang_corefunc.html#like
 		 */
 		return $this->translate_sequence( $node->get_children() );
+	}
+
+	private function translate_regexp_functions( WP_Parser_Node $node ): string {
+		$tokens    = $node->get_descendant_tokens();
+		$is_binary = isset( $tokens[1] ) && WP_MySQL_Lexer::BINARY_SYMBOL === $tokens[1]->id;
+
+		/*
+		 * If the query says REGEXP BINARY, the comparison is byte-by-byte
+		 * and letter casing matters – lowercase and uppercase letters are
+		 * represented using different byte codes.
+		 *
+		 * The REGEXP function can't be easily made to accept two
+		 * parameters, so we'll have to use a hack to get around this.
+		 *
+		 * If the first character of the pattern is a null byte, we'll
+		 * remove it and make the comparison case-sensitive. This should
+		 * be reasonably safe since PHP does not allow null bytes in
+		 * regular expressions anyway.
+		 */
+		if ( true === $is_binary ) {
+			return 'REGEXP CHAR(0) || ' . $this->translate( $node->get_child_node() );
+		}
+		return 'REGEXP ' . $this->translate( $node->get_child_node() );
 	}
 
 	private function get_sqlite_create_table_statement( string $table_name, ?string $new_table_name = null ): array {
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -1321,6 +1321,8 @@ class WP_SQLite_Driver {
 					return $this->translate_regexp_functions( $ast );
 				}
 				return $this->translate_sequence( $ast->get_children() );
+            case 'runtimeFunctionCall':
+                return $this->translate_runtime_function_call( $ast );
 			case 'systemVariable':
 				// @TODO: Emulate some system variables, or use reasonable defaults.
 				//        See: https://dev.mysql.com/doc/refman/8.4/en/server-system-variable-reference.html
@@ -1440,6 +1442,25 @@ class WP_SQLite_Driver {
 		}
 		return 'REGEXP ' . $this->translate( $node->get_child_node() );
 	}
+
+    private function translate_runtime_function_call( WP_Parser_Node $node ): string {
+        $child = $node->get_child();
+        if ( $child instanceof WP_Parser_Node ) {
+            return $this->translate( $child );
+        }
+
+        switch ( $child->id ) {
+            case WP_MySQL_Lexer::LEFT_SYMBOL:
+                $nodes = $node->get_child_nodes();
+                return sprintf(
+                    'SUBSTRING(%s, 1, %s)',
+                    $this->translate($nodes[0]),
+                    $this->translate($nodes[1])
+                );
+            default:
+                return $this->translate_sequence( $node->get_children() );
+        }
+    }
 
 	private function get_sqlite_create_table_statement( string $table_name, ?string $new_table_name = null ): array {
 		// 1. Get table info.
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -21,6 +21,11 @@ class WP_SQLite_Driver {
 	const SQLITE_BUSY   = 5;
 	const SQLITE_LOCKED = 6;
 
+	/**
+	 * A map of MySQL tokens to SQLite data types.
+	 *
+	 * This is used to translate a MySQL data type to an SQLite data type.
+	 */
 	const DATA_TYPE_MAP = array(
 		// Numeric data types:
 		WP_MySQL_Lexer::BIT_SYMBOL                => 'INTEGER',
@@ -78,6 +83,72 @@ class WP_SQLite_Driver {
 		WP_MySQL_Lexer::GEOMETRYCOLLECTION_SYMBOL => 'TEXT',
 
 		// SERIAL, SET, and JSON types are handled in the translation process.
+	);
+
+	/**
+	 * A map of normalized MySQL data types to SQLite data types.
+	 *
+	 * This is used to generate SQLite CREATE TABLE statements from the MySQL
+	 * INFORMATION_SCHEMA tables. They keys are MySQL data types normalized
+	 * as they appear in the INFORMATION_SCHEMA. Values are SQLite data types.
+	 */
+	const DATA_TYPE_STRING_MAP = array(
+		// Numeric data types:
+		'bit'                => 'INTEGER',
+		'bool'               => 'INTEGER',
+		'boolean'            => 'INTEGER',
+		'tinyint'            => 'INTEGER',
+		'smallint'           => 'INTEGER',
+		'mediumint'          => 'INTEGER',
+		'int'                => 'INTEGER',
+		'integer'            => 'INTEGER',
+		'bigint'             => 'INTEGER',
+		'float'              => 'REAL',
+		'double'             => 'REAL',
+		'real'               => 'REAL',
+		'decimal'            => 'REAL',
+		'dec'                => 'REAL',
+		'fixed'              => 'REAL',
+		'numeric'            => 'REAL',
+
+		// String data types:
+		'char'               => 'TEXT',
+		'varchar'            => 'TEXT',
+		'nchar'              => 'TEXT',
+		'nvarchar'           => 'TEXT',
+		'tinytext'           => 'TEXT',
+		'text'               => 'TEXT',
+		'mediumtext'         => 'TEXT',
+		'longtext'           => 'TEXT',
+		'enum'               => 'TEXT',
+		'set'                => 'TEXT',
+		'json'               => 'TEXT',
+
+		// Date and time data types:
+		'date'               => 'TEXT',
+		'time'               => 'TEXT',
+		'datetime'           => 'TEXT',
+		'timestamp'          => 'TEXT',
+		'year'               => 'TEXT',
+
+		// Binary data types:
+		'binary'             => 'INTEGER',
+		'varbinary'          => 'BLOB',
+		'tinyblob'           => 'BLOB',
+		'blob'               => 'BLOB',
+		'mediumblob'         => 'BLOB',
+		'longblob'           => 'BLOB',
+
+		// Spatial data types:
+		'geometry'           => 'TEXT',
+		'point'              => 'TEXT',
+		'linestring'         => 'TEXT',
+		'polygon'            => 'TEXT',
+		'multipoint'         => 'TEXT',
+		'multilinestring'    => 'TEXT',
+		'multipolygon'       => 'TEXT',
+		'geomcollection'     => 'TEXT',
+		'geometrycollection' => 'TEXT',
 	);
 
 	const DATA_TYPES_CACHE_TABLE = '_mysql_data_types_cache';
@@ -902,86 +973,25 @@ class WP_SQLite_Driver {
 			return;
 		}
 
-		/*
-		 * We need to handle some differences between MySQL and SQLite:
-		 *
-		 * 1. Inline index definitions:
-		 *
-		 *  In MySQL, we can define an index inline with a column definition.
-		 *  In SQLite, we need to define indexes separately, using extra queries.
-		 *
-		 * 2. Column and constraint definition order:
-		 *
-		 *  In MySQL, column and constraint definitions can be arbitrarily mixed.
-		 *  In SQLite, column definitions must come first, followed by constraints.
-		 *
-		 * 2. Auto-increment:
-		 *
-		 *  In MySQL, there can at most one AUTO_INCREMENT column, and it must be
-		 *  a PRIMARY KEY, or the first column in a multi-column KEY.
-		 *
-		 *  In SQLite, there can at most one AUTOINCREMENT column, and it must be
-		 *  a PRIMARY KEY, defined inline on a single column.
-		 *
-		 *  Therefore, the following valid MySQL construct is not supported:
-		 *    CREATE TABLE t ( a INT AUTO_INCREMENT, b INT, PRIMARY KEY (a, b) );
-		 *  @TODO: Support it with a single-column PK and a multi-column UNIQUE KEY.
-		 */
+		$table_name = $this->unquote_sqlite_identifier(
+			$this->translate( $node->get_descendant_node( 'tableName' ) )
+		);
 
-		// Collect column, index, and constraint nodes.
-		$columns                = array();
-		$constraints            = array();
-		$indexes                = array();
-		$has_autoincrement      = false;
-		$primary_key_constraint = null; // Does not include inline PK definition.
+		// Save information to information schema tables.
+		$this->information_schema_builder->record_create_table( $node );
 
-		foreach ( $element_list->get_descendant_nodes( 'columnDefinition' ) as $child ) {
-			if ( null !== $child->get_descendant_token( WP_MySQL_Lexer::AUTO_INCREMENT_SYMBOL ) ) {
-				$has_autoincrement = true;
-			}
-			// @TODO: Collect inline index definitions.
-			$columns[] = $child;
+		// Generate CREATE TABLE statement from the information schema tables.
+		$queries            = $this->get_sqlite_create_table_statement( $table_name );
+		$create_table_query = $queries[0];
+		$constraint_queries = array_slice( $queries, 1 );
+
+		$this->execute_sqlite_query( $create_table_query );
+		$this->results      = $this->last_exec_returned;
+		$this->return_value = $this->results;
+
+		foreach ( $constraint_queries as $query ) {
+			$this->execute_sqlite_query( $query );
 		}
-
-		foreach ( $element_list->get_descendant_nodes( 'tableConstraintDef' ) as $child ) {
-			if ( null !== $child->get_descendant_token( WP_MySQL_Lexer::PRIMARY_SYMBOL ) ) {
-				$primary_key_constraint = $child;
-			} else {
-				$constraints[] = $child;
-			}
-		}
-
-		/*
-		 * If we have a PRIMARY KEY constraint:
-		 *  1. Without auto-increment, we can put it back to the list of constraints.
-		 *  2. With auto-increment, we need to later move it to the column definition.
-		 */
-		if ( null !== $primary_key_constraint ) {
-			if ( ! $has_autoincrement ) {
-				$constraints[] = $primary_key_constraint;
-			} elseif ( count( $primary_key_constraint->get_descendant_nodes( 'keyPart' ) ) > 1 ) {
-				throw $this->not_supported_exception(
-					'Composite primary key with AUTO_INCREMENT'
-				);
-			}
-		}
-
-		$query_parts = array( 'CREATE' );
-		foreach ( $node->get_child_node()->get_children() as $child ) {
-			if ( $child instanceof WP_Parser_Node && 'tableElementList' === $child->rule_name ) {
-				$query_parts[] = $this->translate_sequence( array_merge( $columns, $constraints ), ' , ' );
-			} else {
-				$part = $this->translate( $child );
-				if ( null !== $part ) {
-					$query_parts[] = $part;
-				}
-			}
-		}
-
-		// @TODO: Execute queries for inline index definitions.
-
-		$this->execute_sqlite_query( implode( ' ', $query_parts ) );
-		$this->set_result_from_affected_rows();
 	}
 
 	private function execute_alter_table_statement( WP_Parser_Node $node ): void {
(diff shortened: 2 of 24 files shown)


JanJakes:
@adamziel Looking deeper into this, I think that translating specific AST subnodes is pretty straightforward, but what's an open question is the driver/translator architecture, and I'm kind of going back and forth on multiple options.

**Splitting the driver & translator**
I'm not sure whether having all in one large class is a good idea, and I'm exploring ways to split the code a bit. That said, there is the large switch/case statement in the prototype that doesn't need to know much about the AST structure — it just handles some nodes specifically, and passes many other ones down to the translator again, recursively. This has some advantages, such as we don't need to worry much about where a specific grammar rule can appear.

**Splitting top-level constructs**
That said, some kind of distinction makes a lot of sense to me, at least at the highest level. There is a big difference whether we're running a `SELECT` statement, or an `UPDATE` statement, or `CREATE TABLE`, etc. These top-level statements have pretty different semantics, they can return different things, and some special constructs make sense only in some cases (e.g., `SQL_CALC_FOUND_ROWS` works only with `SELECT` statements). Therefore, I'm pretty confident that on the top level, it makes sense to split these use cases a bit into separate classes.

To add a bit more context, even though the top-level statement rule can be re-used in a compound statement for stored procedures, stored functions, triggers, and events, it's probably good to distinguish that case from the top-level use case. E.g., an `ALTER TABLE ...` statement on the top-level may result in directly executing multiple SQLite queries, while when the same is passed as a procedure body, we may need to say it's unsupported.

**Splitting further?**
However, going a little bit deeper, any kind of distinction can be problematic. There are surely some grammar rules that can appear only in the context of others — `tableElementList` can never appear in the `SELECT` clause, for instance — but would it make any sense to implement such rules in CREATE/UPDATE translator class? Or, should there be one huge switch to handle all the grammar rules, with maybe the exception of the top-level ones, as suggested above?

**Translating vs. interpreting**
Another way of looking at organizing the code is the difference between translating MySQL syntax into SQLite syntax vs. emulating a MySQL statement with multiple different SQLite statements or PHP calls. We can't completely separate these two, but I think it would be nice to be clear about where we are just translating, and where some logic is actually executed. Again, it seems to me that the top-level vs. deeper level could help here — each top-level statement could be handled by a "handler" class that would interpret the query in full, and these handlers could use a generic "translator" within them, whose task would only be to translate some MySQL constructs into SQLite syntax.

```php
class WP_SQLite_Create_Table_Translator {
	public function translate_create_table( WP_Parser_Node $ast ) {
		// CREATE TABLE ... LIKE ...
		// CREATE TABLE ... (LIKE ...)
		$has_like = $ast->get_descendant_token( WP_MySQL_Lexer::LIKE_SYMBOL );
		if ($has_like) {
			// Here we will need to run a query to get the table structure,
			// and then create the table:
			return $this->create_table_like( $ast );
		}

		// CREATE TABLE ... ( elements ) options
		// CREATE TABLE ... AS ...
		$elements_ast = $ast->get_child( 'tableElementList' );
		$has_as = $ast->get_descendant_token( WP_MySQL_Lexer::AS_SYMBOL );
		if ( $elements_ast && $has_as ) {
			throw new \Exception('"CREATE TABLE" with both column/constraint definition and "AS" clause is not supported');
		}
		
		// SQLite supporst both ( elements ) and "AS", so we can just rewrite
		// the query and run it. Rewriting can be done either partly here,
		// or fully in the translator.
		$query = $this->translator->translate( $ast );
		return $this->connection->query( $query );

		// Or, we rewrite the top-level `CREATE TABLE` here and nested nodes in the translator:
		$query = new WP_SQLite_Expression(
			array_merge(
				array( WP_SQLite_Token_Factory::raw( 'CREATE TABLE' ) ),
				array(
					$this->translate(
						$ast->get_child( 'tableName' )->get_child()
					),
				),
				array( WP_SQLite_Token_Factory::raw( '(' ) ),
				$this->translator->translate( $ast->get_child( 'tableElementList' ) ),
				array( WP_SQLite_Token_Factory::raw( ')' ) )
			)
		);
		return $this->connection->query( $query );
```

I'm not yet sure what's the optimal solution, but it feels like at least a bit of a concern separation would be useful. It would also be nice if we could solidify some AST node translation by unit testing the translated fragment as well (not only by the end-to-end tests that are run against SQLite), as such test would give immediate feedback and great diffs.

adamziel:
Is there any downside to starting with a single giant class and splitting once that no longer serves the purpose?

Separation of concerns is a useful principle but it cuts both ways. Take [Enterprise FizzBuzz](https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition). Great separation of concerns, extremely unreadable codebase. Too many concerns were defined and separated.

Therefore, I propose starting with just a single concern: Running MySQL queries in SQLite. Why? Because we don't really need modular code here. I don't think we'll ever need to use, say, SELECT statement translation as a component in any other context – therefore is it really useful to separate it?

I only propose that approach as a starting point. Perhaps it will make things unnecessarily complex and we'll have to look for other ways to slice it. I'm just not too excited about preemptively decoupling things that may not need decoupling at all.

The HTML Tag processor, for example, is just a single class even though it tackles a complex problem. It's a huge advantage – most methods are private so we can adjust them without breaking BC, you never have to analyze a dependency graph or wonder where should a specific piece of logic go, you always have full access to the entire state, and you can extend it easily. 

I've tried splitting that logic a few times in the current iteration of the plugin and the outcome was always messy. Making a distinction between translating and interpreting is tempting, but it complicates the design. Perhaps things are different now that we have AST – if yes, then we should hit the wall really soon with a single class approach.

JanJakes:
@adamziel Thanks for your insights! I did not consider that separation of concerns also creates new public APIs as a side effect. That's a very important point. Your proposal makes a lot of sense to me. Somehow, it seemed natural to me that the HTML Tag Processor, or the MySQL lexer, for instance, is handled by a single class, but I wasn't sure about the driver. I guess for parser-like logic, a single huge class is quite common, but in the case of the driver, I started to see too many different things (DB connection, query translation, query execution, etc.), and got myself caught overthinking it.

> Is there any downside to starting with a single giant class and splitting once that no longer serves the purpose?

The only downside is that the giant class can will have multiple responsibilities, and it can get a bit difficult to read. For instance, some state variables will make sense for selects, but not for inserts, etc. But we can mitigate that with good naming, docs, and making the state as simple as possible.

> Separation of concerns is a useful principle but it cuts both ways. Take [Enterprise FizzBuzz](https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition). Great separation of concerns, extremely unreadable codebase. Too many concerns were defined and separated.

Wow, what a codebase! 😂 And a nice demonstration of what over-architecting can cause.

> Therefore, I propose starting with just a single concern: Running MySQL queries in SQLite.

💯

> I only propose that approach as a starting point. Perhaps it will make things unnecessarily complex and we'll have to look for other ways to slice it. I'm just not too excited about preemptively decoupling things that may not need decoupling at all.

Makes sense!

> It's a huge advantage – most methods are private so we can adjust them without breaking BC

This! 💯

---

Looking at the architecture, I'd like to simplify also the current draft with `WP_SQLite_Expression` and `WP_SQLite_Query_Builder`. I think maybe we could make it work without them.

The `WP_SQLite_Query_Builder` takes the expression and stringifies it, joining all strings with a space and wrapping sub-expressions with parentheses. It covers many use cases out of the box, but it's not really a definition of the SQLite SQL dialect or anything like that, and it's not very flexible — for instance, a list of column definitions is delimited by a comma, not a space, and maybe the current implementation would put spaces even into places where they aren't expected. Apart from that, it performs some operations on some tokens (quoting identifiers, for instance), which could be done by the driver itself.

For `WP_SQLite_Expression`, it may be possible to replace it within the driver with return types of the recursive `translate` calls — either we return a token, or we could return an array (~ the expression). In any case, if we'll need to keep this one for some reason, I'd consider renaming it to something like `WP_SQLite_Node` to avoid confusion with [SQL Language Expressions](https://www.sqlite.org/lang_expr.html) or any hints that it maps to some SQLite construct.

I will quickly try out these simplifications to see if they could make sense.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -763,6 +784,8 @@ private function translate_token( WP_MySQL_Token $token ) {
 				return null;
 			case WP_MySQL_Lexer::IDENTIFIER:
 				return '"' . trim( $token->value, '`"' ) . '"';
+			case WP_MySQL_Lexer::AUTO_INCREMENT_SYMBOL:
+				return 'AUTOINCREMENT';
```
adamziel, line 1358: So easy now :heart:

tests/parser/WP_Parser_Node_Tests.php
```diff
@@ -0,0 +1,144 @@
...
+		// @TODO: Consider breadth-first search vs depth-first search.
+		$this->assertSame( $n_keyword, $root->get_descendant_node() );
+		$this->assertSame( $n_expr_a, $root->get_descendant_node( 'expr' ) );
+		$this->assertSame( null, $root->get_descendant_node( 'root' ) );
+		$this->assertSame( $t_comma, $root->get_descendant_token() );
+		$this->assertSame( $t_one, $root->get_descendant_token( 400 ) );
+		$this->assertSame( null, $root->get_descendant_token( 123 ) );
+
+		// Test multiple descendant methods.
+		// @TODO: Consider breadth-first search vs depth-first search.
```
JanJakes, line 114: @adamziel I initially kept the logic from the first prototype (which was traversing descendants breadth-first), and added this note to look into it later, but now it seems to me it may be more useful and natural to do a dept-first search (which would return tokens in the order from the original query).

I'm still not 100% sure which one we need more — usually, I just want to know "is there this node or that token somewhere in this subtree", and in that case, both searches work. But sometimes it may be useful to check for a next token or more, for instance, and then `get_descendant_tokens()` would be more natural with a depth-first traversal. If really needed, we could also provide both. What do you think?

JanJakes:
@adamziel Another thought regarding `WP_Parser_Node`: We now provide child and descendant methods. Should we also provide parent methods? It can be useful, but I think it has a major flaw. I'll explain:

Sometimes, it could be useful in the driver, if we want to translate some constructs in a way "if I'm in the context of `createTable`, do this, otherwise do that". This is a simple way to express some specific nuances, but it's not the only way — we could also handle it from the top within the `createTable` node instead, constructing the statement a bit more top-down (but it can be more verbose at times).

That said, I think there's a problem — adding parent references to `WP_Parser_Node` would probably cause memory leaks. PHP GC uses reference counting, which means that [it only frees the memory occupied by an object when there are no more references to it](https://phpstan.org/blog/phpstan-2-0-released-level-10-elephpants#lower-memory-consumption-leads-to-faster-performance), and that circular parent-child references would never allow for garbage collection. Weak refs were only introduced in PHP 7.4, so I guess currently, it doesn't make sense to do this. It also complicates the tree a bit and makes subtrees not "independent". There is the alternative of saving some context (node stack) in the driver during the translation, for instance.

adamziel:
@JanJakes what would be a specific code example where a parent reference would help? I get "do this, do that," but what's "this" and "that"?

adamziel:
Also the HTML and the XML API solve the same problem with the get_ breadcrumbs a method that gives you a stack of parent elements names. Maybe the same would use here for AST notes?

JanJakes:
> @JanJakes what would be a specific code example where a parent reference would help? I get "do this, do that," but what's "this" and "that"?

@adamziel For now, I've run into the `INTEGER PRIMARY KEY` "without" vs "with" autoincrement — in the former case, I'd like to translate it as `INT PRIMARY KEY` while in the latter one as `INTEGER PRIMARY KEY` due to [the "bug" that's preserved in SQLite for back compatibility](https://www.sqlite.org/lang_createtable.html#rowids_and_the_integer_primary_key). In the AST, it would look something like this:

```
createStatement
   |- createTable
     |- tableElementList
       |- tableElement          <-- Here I can reliably determine PRIMARY KEY
       |  |- columnDefinition
       |    |- fieldDefinition  <-- Here I can reliably determine AUTOINCREMENT
       |     |- dataType        <-- Here I want to know about PRIMARY KEY and AUTOINCREMENT
       |     |- columnAttribute <-- Here PRIMARY KEY and AUTOINCREMENT can be specified
       |     ...
       |- tableElement
         |- tableConstraintDef  <-- But PRIMARY KEY can also be specified here (AUTOINCREMENT not)
```

Without the `tableConstraintDef` option for primary keys, it would be simple-enough to handle this case in `fieldDefinition` in a top-down manner. Something like:

```php
case 'fieldDefinition':
	$has_primary_key   = ...;
	$has_autoincrement = ...;
	if ( $has_primary_key ) {
		$children       = $ast->get_children();
		$data_type_node = array_shift( $children );
		$data_type      = $this->translate( $data_type_node );
		if ( 'INTEGER' === $data_type ) {
			$data_type = $has_autoincrement ? 'INTEGER' : 'INT';
		}
		return $data_type . $this->translate_sequence( $children );
	}
	return $this->translate_sequence( $ast->get_children() );
```

Here, the `$has_primary_key` can peek below for descendant token, but it has no way to verify the `tableConstraintDef`, that would be in a different `tableElement`. I would either need a reference to the parent `tableElement` to use it to search for a primary key, or I would need to save it somehow (`$this->set_context('primary_key', ...)`) and then read it below.

---

However, in the case of `CREATE TABLE`, I'm also wondering whether it may make more sense to fully parse the create table statement (into columns, constraints, etc.), and only then construct a query from that, like we do in the current translator. It's granular changes on specific AST nodes vs reading the full MySQL statement and generating a new SQLite statement from that. I don't know yet which of these options will be ultimately more convenient and reliable.

adamziel:
Two ideas I got to avoid a parent reference:

* Could the driver or parser class keep an array like $child => $parent and expose a method such as get_parent_of?
* The query execution process could keep track of the stack of nodes it entered

adamziel:
> It's granular changes on specific AST nodes vs reading the full MySQL statement and generating a new SQLite statement from that. I don't know yet which of these options will be ultimately more convenient and reliable.

Are these actually different? You need to traverse the tree node by node in either case. Or would the latter mean doing one pass to build a flat index of, say, modifiers and options, and then doing the translation in the second pass? Because I don't think that index could be flat - we'd still need a recursive data structure to account for subqueries.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1093 @@
...
+				$update_nodes[] = $child;
+				if (
+					$child instanceof WP_Parser_Node
+					&& 'updateList' === $child->rule_name
+				) {
+					// Skip WHERE, ORDER BY, and LIMIT.
+					break;
+				}
+			}
+			$query = $this->translate_sequence( $update_nodes )
```
adamziel, line 979: Thought: we could have a single translate() method that would support both strings and arrays

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1093 @@
...
+			);
+
+			$update_nodes = array();
+			foreach ( $node->get_children() as $child ) {
+				$update_nodes[] = $child;
+				if (
+					$child instanceof WP_Parser_Node
+					&& 'updateList' === $child->rule_name
+				) {
+					// Skip WHERE, ORDER BY, and LIMIT.
```
adamziel, line 975: I'd love to see a relevant part of the grammar here. Let's comment it inline if it's not too large

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1093 @@
...
+		 *
+		 * For instance, the following query:
+		 *   UPDATE t SET c = 1 WHERE c = 2 LIMIT 1;
+		 * Will be rewritten to:
+		 *   UPDATE t SET c = 1 WHERE rowid IN ( SELECT rowid FROM t WHERE c = 2 LIMIT 1 );
+		 */
+		if ( $has_order || $has_limit ) {
+			$subquery = 'SELECT rowid FROM ' . $this->translate_sequence(
+				array(
+					$node->get_descendant_node( 'tableReference' ),
```
adamziel, line 961: Is there any danger of sourcing a wrong descendant here? E.g. from a subquery? I wonder what's the right balance between shorthands and verbosity. E.g. would a `get_node( [ "whereClause", "columnsList" ] )` be both useful and safe? Or is there, say, 20 possible paths and listing them all would be tedious?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel Looking at this simple use case, I agree that it would make sense, but the whole context is that has an additional parameter (and maybe more, in the future) that doesn't really fit into the `translate()` recursion.

Joining the subtrees by space is the default:
```php
private function translate_sequence( array $nodes, string $separator = ' ' ): string
```

But there are other use cases with a different separator, such as a list of column definitions — `translate_sequence( $nodes, ', ' )`, etc. I think we may also need a `$prefix` and `$suffix` in the future, e.g. to wrap the subtrees in parentheses, etc. It really is just a helper so we don't have to repeat the same loop in multiple places.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: > Is there any danger of sourcing a wrong descendant here?

@adamziel Ah, there is! Great point. It also relates to [the question of breadth-first vs depth-first descendant search](https://github.com/WordPress/sqlite-database-integration/pull/164#discussion_r1854073858). With the current breadth-first implementation, it should pick the correct table reference in this case:

```antlr
updateStatement:
    ({serverVersion >= 80000}? withClause)? UPDATE_SYMBOL LOW_PRIORITY_SYMBOL? IGNORE_SYMBOL? tableReferenceList SET_SYMBOL
        updateList whereClause? orderClause? simpleLimitClause?
;
```

I will write some tests for this.

As for the general problem, it can be more complicated. For instance, if the `tableReferenceList` was optional, then it could be ambiguous.

---

> E.g. would a get_node( [ "whereClause", "columnsList" ] ) be both useful and safe? Or is there, say, 20 possible paths and listing them all would be tedious?

I like this! It's more explicit and safer. In this particular case, it only has a single path `tableReferenceList > tableReference`.

(As a side note, it may seem to be all more complex due to the possibility of multiple table references, but thankfully, that's not possible in MySQL: `However, you cannot use ORDER BY or LIMIT with a multiple-table UPDATE`; [link](https://dev.mysql.com/doc/refman/8.4/en/update.html)).

---

Now, if we make the node-querying API more powerful, does it still belong to the node class, or should we extract it somehow? Reflecting more broadly, it reminds me of DOM query selectors, and these are available on every level, so we may be fine with that. I don't want to over-engineer it at this stage, but taking some inspiration from DOM queries could be interesting. Maybe we could add `$node->query( ... )` and `$node->query_all( ... )` methods:

1. Child: `$node->query( 'tableReferenceList' )`
2. Path: `$node->query( [ 'tableReferenceList', 'tableReference' ] )`
3. Wildcard: `$node->query( ['tableReferenceList', '*', 'tableReference'] )`
4. Recursive search: `$node->query( ['tableReferenceList', '**', 'tableReference'] )`

At the same time, it is a question what would be a reasonable balance between API simplicity and power.

JanJakes:
> > It's granular changes on specific AST nodes vs reading the full MySQL statement and generating a new SQLite statement from that. I don't know yet which of these options will be ultimately more convenient and reliable.
> 
> Are these actually different? You need to traverse the tree node by node in either case. Or would the latter mean doing one pass to build a flat index of, say, modifiers and options, and then doing the translation in the second pass? Because I don't think that index could be flat - we'd still need a recursive data structure to account for subqueries.

@adamziel I think they are a bit different:

1. The first approach with only the necessary granular changes often relies on the `default` case that just recursively translates all children and concatenates them with a space. At the moment, this often works thanks to the similarity of the two SQL dialects, so even tokens are reused (`SELECT`, parentheses, etc.). In a way, we're just taking the MySQL query and doing some adjustments, where needed. Over time, we'll probably cover the translation more and more on each node, and will rely a bit less on the shape of the original MySQL query.
2. The second approach would be oriented more towards reading data and then constructing a whole new query, rather than doing only the necessary changes. The index of the elements could be flat, at least in this case. More generally, even with this approach, the recursion is still possible — if I know that I have a list of some definitions, I can still get them with `$definition = $this->translate( ... )`, in case they can contain expressions, subqueries, etc.

Looking at `CREATE TABLE` in particular, and the grammar branch that contains `tableElementList` (the last below),

```antlr
createTable:
    TEMPORARY_SYMBOL? TABLE_SYMBOL ifNotExists? tableName (
        LIKE_SYMBOL tableRef
        | OPEN_PAR_SYMBOL LIKE_SYMBOL tableRef CLOSE_PAR_SYMBOL
        | (OPEN_PAR_SYMBOL tableElementList CLOSE_PAR_SYMBOL)? createTableOptions? partitionClause? duplicateAsQueryExpression?
    )
;
```
we would be able to get a list of all column and constraint definitions, organize them somehow, and then generate the new `CREATE TABLE` query. The question is whether we should do that, or if we should just go with option 1.

---

Here's maybe a good example for thinking about this:

A `PRIMARY KEY` constraint can be defined either within a particular column definition, or in a different table element, separately from the column definition. Now:

1. With option 1, we'd keep the definition where it was in MySQL, not normalizing it in any way, and relying on the fact that both MySQL and SQLite support defining `PRIMARY KEY` in both places. When we would run into a specific constraint that can be defined in both places in MySQL, but only in a single one in SQLite, we'd just handle that specific case.
2. With option 2, we would first create a list of all columns and constraints, and then generate a query from that. Inline constraints would be normalized to separate constraint definitions (or the other way round), etc.

As a consequence, option 1 fixes only the particular cases that SQLite supports differently, leaving the rest mostly to their similarities. If there is a specific keyword or modifier we need to ignore, it needs to be done explicitly.

With option 2, we would be picking and allow-listing the things we support, possibly ignoring any other ones. In many cases, I think option 2 may be a bit more verbose, unless there are many exceptions and edge cases that would have to be handled in option 1.

Ultimately, we can do a little bit of both — sometimes we can be more descriptive for a specific node, we can decompose it and fully reconstruct in an option 2 fashion, while still leveraging option 1 as well by translating some subnodes recursively. The decision may be needed on a case-by-case, but it is a question of what would be a useful guideline to deciding this.

---

And here's a particular example:

In MySQL, a PRIMARY KEY with AUTOINCREMENT can be defined in both of these ways:
a) `CREATE TABLE t (id INT PRIMARY KEY AUTO_INCREMENT);`
b) `CREATE TABLE t (id INT AUTO_INCREMENT, PRIMARY KEY (id) );`

While in SQLite, b) is a syntax error. It means that we have to rewrite it to something like a) because AUTOINCREMENT in SQLite can be only in the column definition and only after a PRIMARY KEY constraint. And here we're touching on the questions — handling this particular case vs reading and normalizing the full statement.

---

Another interesting example:
- In MySQL, we can write both `PRIMARY KEY AUTO_INCREMENT` and `AUTO_INCREMENT PRIMARY KEY`.
- In SQLite, it always needs to be `PRIMARY KEY AUTOINCREMENT`, the other one is a syntax error.

Going back to the two options, with option 1, we need to catch and handle this case, while with option 2, it would probably just work due to the full statement reading and normalization, at the likely cost of more verbosity.

Ultimately, it may also work that we only normalize the PRIMARY KEY & AUTOINCREMENT definitions, leaving the rest as-is. At the moment, I don't know yet if there will be more of such cases in this particular statement.

adamziel:
Fantastic explanation @JanJakes, thank you so much for making the distinction so clear!

I don't know the answer. We'll have to discover it through exploration. In my attempts, walking over the AST and manipulating the resulting query seemed to be quite handy.

Translating the MySQL AST into another intermediate data structure seemed cumbersome, but perhaps iterating over the AST and storing metadata would be helpful, as in `$table_ast_node['metadata']['primary_key'] = ['column 1', 'column 2']`. But perhaps a computation would be just as good, as in `$primary_key = $this->find_primary_key_columns($table_ast_node)`.

Either way, it seems like we're in a multi-pass land where at least one pass is needed to scan the data and perhaps construct a query scaffold, and another pass is needed to finalize the query construction. With a computation-based approach such as `$this->get_some_info($ast_node)` the two passes would use co-located logic, but we'd also do more work than necessary. With a metadata tracking approach, such as `$this->pk[] = $node['column_name']`, we'd do explicit two passes but also distribute parts of logic that are tightly related. Tradeoffs, tradeoffs.

It would be useful to branch out, start a sub-pr for each idea, and discuss the ups and downs.

JanJakes:
@adamziel Looking closer, I think it also depends on the AST-querying capabilities. If we could be more specific (e.g., querying for `tableElementList > tableElement > tableConstraintDef > "PRIMARY"`, then that could be safer and better optimized than a full descendant search. That said, this still doesn't give me "the node contains the PRIMARY keyword" and I don't have a parent reference.

When looking at the "storing metadata" vs "doing a computation" problem, I realize that storing metadata can involve a computation as well, and in the particular case of the PRIMARY KEY constraint, both options will actually involve a pretty similar one. To get the metadata, we need to look into the `tableElementList` subtree first, as we don't know whether the primary key constraint is defined before or after the column specification. The issue with computation-only would likely be that we would call it for every column again and again, while the metadata approach would "cache" it.

What I want to say is that if we have powerful and performant AST-querying capabilities, we could probably always do computations. Then it's a question whether such capabilities belong to `WP_Parser_Node` or rather something external to it (that could support even parent lookups):

```php
$query = new WP_Parser_Query( $ast );
$query->selector( $node, ['tableElementList', 'tableElement', 'tableConstraintDef', ...] );
$query->selector_all( ... );
$query->parent( $node );
// etc.
```

However, this would likely mean doing a full AST pass first to store some lookup tables, or we'd need to make them lazily while querying. Also, I don't want to overengineer it, and this could go in that direction. At the same time, having better AST search capabilities at hand would be great, and it would work for any other grammar as well.

Anyway, I guess I'm throwing out here too many implementation details, but it does help to write things down to get the thoughts more organized and have a better overall view.

JanJakes:
@adamziel I took a stab at this in https://github.com/WordPress/sqlite-database-integration/pull/164/commits/797c3b78f4035f9da4262be75f55190e989a92a5. It combines a bit of top-down (`execute_create_table_statement`) with the recursion approach. It's not overly verbose, but the downside is that some problems are handled in multiple different nodes (the PRIMARY key & AUTOINCREMENT issue). It also handles some other MySQL vs SQLite CREATE TABLE specifics (see the comment in `execute_create_table_statement`).

Saving metadata into a state wouldn't change that much here, I suppose. At least not the main architecture of having to do some logic close to top-level, and some other logic deeper. What would change it significantly is the full parsing approach, but as we discussed, that may not be great either, and it will be certainly much more verbose. We would need methods, such as [parse_create_table](https://github.com/WordPress/sqlite-database-integration/blob/ac75e90e052ad70b40729cb7b4bdc0c691fc6b52/wp-includes/sqlite/class-wp-sqlite-translator.php#L952), [parse_mysql_create_table_field](https://github.com/WordPress/sqlite-database-integration/blob/ac75e90e052ad70b40729cb7b4bdc0c691fc6b52/wp-includes/sqlite/class-wp-sqlite-translator.php#L1074), [parse_mysql_create_table_constraint](https://github.com/WordPress/sqlite-database-integration/blob/ac75e90e052ad70b40729cb7b4bdc0c691fc6b52/wp-includes/sqlite/class-wp-sqlite-translator.php#L1222), etc.

I could try to go further with this approach, focusing on more complex things like ALTER TABLE, and see where it takes us. What do you think?

JanJakes:
After exploring the rabbit hole of emulating the `ALTER TABLE` statements that are not supported by SQLite, I brainstormed with @adamziel about the possible ways to address this, and came with the following idea:

**Let's implement a subset of MySQL's `INFORMATION_SCHEMA` tables.**

This will provide us with the foundation for supporting these constructs:
- Some `ALTER TABLE ... CHANGE ...`, `ALTER TABLE ... MODIFY ...`, and other `ALTER ...` statements.
- Some `DESCRIBE` statements.
- `SHOW CREATE TABLE` statement.
- Querying the `INFORMATION_SCHEMA` tables.
- Likely more that I'm not aware of at the moment.

This approach will serve multiple different use cases while implementing the support for `INFORMATION_SCHEMA` queries at the same time (https://github.com/WordPress/sqlite-database-integration/issues/78, https://github.com/WordPress/sqlite-database-integration/issues/146).

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -1149,6 +1172,8 @@ private function translate( $ast ) {
 					return null;
 				}
 				return $this->translate_sequence( $ast->get_children() );
+			case 'defaultCollation':
+				return null;
```
adamziel, line 1335: Can we actually support anything else than UTF-8? If not, we could make this an error for non-utf8 values. Or a warning at least.

wp-includes/sqlite-ast/class-wp-sqlite-information-schema-builder.php
```diff
@@ -0,0 +1,1000 @@
...
+		// TABLES
+		"CREATE TABLE IF NOT EXISTS _mysql_information_schema_tables (
+			TABLE_CATALOG TEXT NOT NULL DEFAULT 'def',   -- always 'def'
+			TABLE_SCHEMA TEXT NOT NULL,                  -- database name
+			TABLE_NAME TEXT NOT NULL,                    -- table name
+			TABLE_TYPE TEXT NOT NULL,                    -- 'BASE TABLE' or 'VIEW'
+			ENGINE TEXT NOT NULL,                        -- storage engine
+			VERSION INTEGER NOT NULL DEFAULT 10,         -- unused, in MySQL 8 hardcoded to 10
+			ROW_FORMAT TEXT NOT NULL,                    -- row storage format @TODO - implement
+			TABLE_ROWS INTEGER NOT NULL DEFAULT 0,       -- not implemented
```
adamziel: I can't think of anything better than computing this and other values here upon query – perhaps a view would work for that. Also, not implementing these seems fine for v1.

wp-includes/sqlite-ast/class-wp-sqlite-information-schema-builder.php
```diff
@@ -0,0 +1,1000 @@
...
+		$this->query_callback = $query_callback;
+	}
+
+	public function ensure_tables(): void {
+		foreach ( self::CREATE_INFORMATION_SCHEMA_QUERIES as $query ) {
+			$this->query( $query );
+		}
+	}
+
+	public function create_table( WP_Parser_Node $node ): void {
```
adamziel: Just `create_table` suggests this method actually creates a table. How about `store_create_table` or `notify_create_table` or something to that effect?

wp-includes/sqlite-ast/class-wp-sqlite-information-schema-builder.php
JanJakes: @adamziel Yeah, a view sounds like a good idea. In fact, from MySQL 8, all the information schema tables are views on fewer virtual and invisible system tables under the `mysql` database (`mysql.tables`, etc.), which hints that maybe we could store this info in fewer more universal tables, and then just create views, which could also read from other tables like `sqlite_master` and other ones.

That said, in the first phase, I want to cover as much as possible in the most straightforward way because otherwise it could get a bit too complex for me to tackle at once. Once it works and the test coverage is decent, we could further simplify, optimize, etc.

wp-includes/sqlite-ast/class-wp-sqlite-information-schema-builder.php
JanJakes: @adamziel Makes sense! I'll revisit the naming as well, there's likely more to improve at the moment.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel Good point! I just removed it here because SQLite doesn't support `COLLATE = ...` on tables, and I'm storing the exact value in the information schema. I think we'll need to allow all the `utf8_something`, `utf8mb3_something`, and `utf8mb4_something` collations at least, although I suppose we won't be able to sort the values the same way as MySQL does. We could also consider allowing single-byte charsets, etc.

adamziel:
@JanJakes lovely work here, and I'm really happy about how the information schema is coming together. What's the delta between here and passing all the unit tests?

JanJakes:
> What's the delta between here and passing all the unit tests?

@adamziel I think the hardest parts are the `CREATE TABLE`, `ALTER TABLE`, and some of the `SHOW`/`DESCRIBE` statements. In other words, having the information schema will unlock many of the most complex features, and I think with a decent information schema coverage at the moment, we can start exploring that. The implementation of many of the other missing features could be portable from the legacy translator, and some of them are already in your prototype — SQL functions, `SQL_CALC_FOUND_ROWS`, `FROM DUAL`, etc.

Now I want to look into the `information_schema` -> SQLite `CREATE TABLE` conversion, which is needed for `ALTER` statements that need to copy the table. And having that, I think we can simply use it for CREATE TABLE statements as well. Then we also need `information_schema` -> MySQL `CREATE TABLE` for `SHOW` statements, etc.

There is still quite a bit of work to do, but with the hardest stuff getting done first, I think the number of failing tests will decrease faster and faster. In the end, I think that when we pass the existing test case, we'll immediately have a higher MySQL compatibility due to the nature of the new AST & information schema approach.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
adamziel: Makes sense, thank you! Let's make it an error for now, then, and document what you just said inline to make sure we give this the attention it needs in the future.

JanJakes:
### 🎄 Update

I'm finally here with an update, I'm sorry for being overdue. It's a longer read, probably should've been split into multiple.

I'll start with a **summary**:

1. **61%** (71/116) of the original tests are passing. Before starting the `INFORMATION_SCHEMA` rewrite, it was 33% (38/116).
2. A significant portion of some of the hardest statements is done — `CREATE TABLE` and `ALTER TABLE` operations.
3. The basic forms of `SHOW` and `DESCRIBE` using the `INFORMATION_SCHEMA` are implemented.
4. However, I think there's also still quite a bit of work left to get to 100%, especially to get everything correct and stable.
5. The remaining failing tests are:
     - String matching and regexp support (`LIKE`, `RLIKE`, etc.) — 8 tests.
     - Date/time related columns — 5 tests.
     - `SQL_CALC_FOUND_ROWS` — 4 tests.
     - `LEFT` function — 4 tests.
     - `SHOW TABLES` with `LIKE` and `FROM` — 4 tests.
     - `ON CONFLICT ...` — 3 tests.
     - `SELECT FROM DUAL` — 2 tests.
     - Other things like `CREATE TEMPORARY TABLE`, `HAVING COUNT` variations, `CURRENT_TIMESTAMP` syntax, `ON UDPATE`, `CAST` as binary, etc.
6. For more details, check out the [`WP_SQLite_Information_Schema_Builder`](https://github.com/WordPress/sqlite-database-integration/blob/ast-sqlite-driver/wp-includes/sqlite-ast/class-wp-sqlite-information-schema-builder.php).

The full list of failures as of today can be found here: https://gist.github.com/JanJakes/23e36446b24c88bee7ca81415d9f6059

--- 

Now, I'd like to share some more details about the progress, and the rabbit holes I encountered along the way.

#### Simple statements -> Basic DDL -> Information schema
A couple of weeks ago, I started to implement the new SQLite driver to translate MySQL statements to SQLite statements, using the new [MySQL parser](https://github.com/WordPress/sqlite-database-integration/pull/164). I implemented the basics of a few simple statements, such as `SELECT`, `INSERT`, `UPDATE`, `REPLACE`, and `DELETE`, and then I switched my focus to DDL statements, such as `CREATE TABLE` and `ALTER TABLE`.

This turned out to bring [some complexities](https://github.com/WordPress/sqlite-database-integration/pull/164#issuecomment-2497443502), and after further discussions, and realizing that a pure MySQL -> SQLite DDL conversion is lossy, [we decided](https://github.com/WordPress/sqlite-database-integration/pull/164#issuecomment-2514858406) to implement a subset of MySQL's `INFORMATION_SCHEMA` tables.

#### Implementing `INFORMATION_SCHEMA` tables
First, I decided to focus on the schema tables that are necessary to describe tables, columns, indexes, and constraints. I started by implementing the `CREATE TABLE` statement. This is essentially an AST to information schema query conversion. In its basics, it's a simple logic, but reaching high coverage and correctness turned out to be pretty extensive.

Here are some interesting issues I encountered:
- Some columns depend on each other. Sometimes, a charset can determine a collation (we need a full map of default collations for each charsets), sometimes a collation can determine a charset (here, it's just its prefix). Column lengths and numeric precisions depend on data types, charset (we need a map of charset to maximum bytes per character), etc.
- Some tables depend on each other. Adding or dropping a constraint may affect the nullability and some other attributes of a column as well, and this data needs to be synchronized. Maybe in the future, we could reduce the number of actual information schema tables, and use views to present the data (MySQL has a similar approach from version 8).
- Some operations have non-trivial effects on the schema. Dropping a column will remove it from existing indexes, meaning that multi-column index records may need to be renumbered, and other column attributes may change.
- Some information schema values require a lot of normalization. There are so many was to write varchars, national varchars, and text fields (`char varying`, `character varying`, `varchar`, `national varchar`, `nvarchar`, `nchar varchar`, `national char varying`, `nchar varying`, `long char varying`, `long character varying`, `long varchar`, ...), and some keywords can change their meaning (`long` converts varchar to text, appending `CHARSET BINARY` seems to change varchar to varbinary, etc.).
- For some values, we may need to implement an AST -> MySQL string conversion. More on that below.

I've either addressed/`TODO`-ed all the issues I've encountered, but a more extensive test case for some of them is still needed. More details can be seen in [`WP_SQLite_Information_Schema_Builder`](https://github.com/WordPress/sqlite-database-integration/blob/ast-sqlite-driver/wp-includes/sqlite-ast/class-wp-sqlite-information-schema-builder.php) which implements all of the AST -> information schema logic.

#### MySQL AST -> string conversion
As I was working on the information schema builder, I realized I not only need to read the MySQL AST, but in some cases, it's also necessary to convert some nodes back to strings. A simple example of this is reading a table name from a node that has multiple nesting levels below (`identifier -> pureIdentifier -> IDENTIFIER`). A more complex example is a column definition with a `DEFAULT(expr)`, in which case I need to save the serialized expression in the `INFORMATION_SCHEMA`.

I think there are multiple approaches we could consider:
1. Maybe just walking the tree and joining all tokens with spaces would work.
2. We could store the original whitespaces in the AST, and then join the values.
3. We could store offsets in the original SQL input for each node, and then just cut out what we need.
4. We could implement an AST -> SQL string printer, maybe a bit more involved than option 1.

Additionally, I often need to get a bit more than the original string. For identifiers, the value is not enough — I need to unquote it first (replace quoted sequences by characters, trim quotes), and it is a question where such logic belongs.

#### ALTER TABLE
The next topic after storing the `CREATE TABLE` information in the `INFORMATION_SCHEMA` was implementing support for the most common table altering operations. For that, I refactored my initial draft to reusable methods, and realized that ALTER TABLE statements brings another set of challenges.

[SQLite has a limited support for ALTER TABLE statements](https://www.sqlite.org/lang_altertable.html), so I started by trying to map those statements that SQLite can handle to their SQLite counterparts, while falling back to [a table-copying approach](https://www.sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes) for the other ones. This turned out to be a rabbit hole:

1. ALTER TABLE statements can contain multiple operations, and they often do. Therefore, I needed to try to first decide whether I needed to copy the table, or whether I will be able to just map a single MySQL statement to multiple SQLite ALTER TABLE statements.
6. Determining which MySQL ALTER TABLE statements are directly supported by SQLite is not straightforward. It may seem that an operation, such as ADD COLUMN is simply supported, but the reality is that it might not be in all cases (e.g., with `FIRST`/`AFTER ...`).
7. Implementing the "direct" plus "fallback to full copy" approach means doing many things twice.

For these reasons and after some experimenting, I decided to first focus on implementing the full table-copy ALTER TABLE algorithm, leaving the direct translation of some operations as an optimization for later. The approach is pretty straightforward:

1. Reflect all ALTER TABLE operations in the `INFORMATION_SCHEMA`.
2. Run the SQLite ALTER TABLE algorithm with full table copy [as per the SQLite docs](https://www.sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes).

This turned out to work flawlessly and to be very helpful in the initial implementation phase. Implementing a new ALTER TABLE operations is as simple as reflecting its changes in the information schema at the moment. This is great for broadening the statement support before focusing on optimizing some of the statements.

Finally, [the ALTER TABLE algorithm recommended in the SQLite docs](https://www.sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes) requires only a one table copy, so in the new driver, we don't do two copies like we did in the old one.

#### Connecting the pieces
What turned out to be challenging in this first phase of implementing the `INFORMATION_SCHEMA` support was the fact that until I got to some level of completeness, I wasn't able to see any progress in terms of passing tests, and it wasn't easy to estimate where exactly we were regarding the compatibility with the existing SQLite translator.

This is because:
1. The actual database operations can't be implemented before the data is stored in `INFORMATION_SCHEMA`.
2. Saving only some data in the `INFORMATION_SCHEMA` means the tests will fail because the data is incomplete.
6. Implementing the `INFORMATION_SCHEMA` builder and the CREATE/ALTER TABLE operations is not enough either. Most of the tests verify the results using `SHOW` and `DESCRIBE` statements.

Therefore, while working on the `INFORMATION_SCHEMA`, I haven't seen any progress on the original SQL translation test suite (but I did write a new set of tests along the way). But the reward comes at the end — quickly prototyping `SHOW` and `DESCRIBE` statements with the `INFORMATION_SCHEMA` at hand turned out to be easy and straightforward, and suddenly, many tests started to pass, while for some others, this revealed incorrect assertions, as the old driver doesn't always have 100% correct data at hand.

#### Next steps
There is still a good amount of work to get to 100% of the test suite for the current SQLite translator, but I think in the current state, there is a good foundation for many of the hard parts, and building on that will hopefully be easier. Above that, there are some MySQL features that can likely be ported from the old translator with not so much of a rewrite.

Additionally, the existing test suite may not fully cover everything that the current translator implements, so I think it would be good to look at multiple measures:

1. The percentage of passing tests in the existing test suite.
2. Checking the existing translator method by method and ensuring we implement all the operations.
3. Running WordPress with the new driver.
4. Examining the new driver using [Playground Tester](https://github.com/bgrgicak/playground-tester).

That said, maybe the path to get this merged, e.g., with a feature flag, is to focus on point 1, and do the rest in other PRs.

adamziel:
Monumental work here @JanJakes, and a tremendous update. Mad props. 

> That said, maybe the path to get this merged, e.g., with a feature flag, is to focus on point 1, and do the rest in other PRs.

I would love that! It feels like a good spot to park, do a round or a few rounds of review, and merge. Let's just rewire the CI tests to the old driver to avoid failing checks on trunk.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+	/**
+	 * Destructor
+	 *
+	 * If SQLITE_MEM_DEBUG constant is defined, append information about
+	 * memory usage into database/mem_debug.txt.
+	 *
+	 * This definition is changed since version 1.7.
+	 */
+	public function __destruct() {
+		if ( defined( 'SQLITE_MEM_DEBUG' ) && SQLITE_MEM_DEBUG ) {
```
adamziel, line 440: Did you run into memory usage issues?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+					"Code: $status",
+					"Error Message: $err_message"
+				);
+				$this->is_error         = true;
+				$this->error_messages[] = $message;
+				return;
+			}
+		}
+
+		new WP_SQLite_PDO_User_Defined_Functions( $pdo );
```
adamziel, line 386: Code style nitpick: It's weird to trigger side effects by creating a new class instance that's not even used later on. How about a named static call? e.g. `WP_SQLite_PDO_User_Defined_Functions::register_for( $pdo )` or so?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+		);
+		$this->information_schema_builder->ensure_information_schema_tables();
+
+		// Load MySQL grammar.
+		if ( null === self::$grammar ) {
+			self::$grammar = new WP_Parser_Grammar( require self::GRAMMAR_PATH );
+		}
+
+		// Fixes a warning in the site-health screen.
+		$this->client_info = SQLite3::version()['versionString'];
```
adamziel, line 413: How does it fix it?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+
+		register_shutdown_function( array( $this, '__destruct' ) );
```
adamziel, line 415: won't this call the destructor twice? once when the object is disposed and once on shutdown? I guess the reference will stick around until the shutdown hooks run, though

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+
+		// WordPress happens to use no foreign keys.
+		$statement = $this->pdo->query( 'PRAGMA foreign_keys' );
```
adamziel, line 418: This comment is a good start, but let's briefly explain what PRAGMA foreign_keys does

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+		// phpcs:ignore Universal.Operators.StrictComparisons.LooseEqual
+		if ( $statement->fetchColumn( 0 ) == '0' ) {
+			$this->pdo->query( 'PRAGMA foreign_keys = ON' );
+		}
+		$this->pdo->query( 'PRAGMA encoding="UTF-8";' );
+
+		$valid_journal_modes = array( 'DELETE', 'TRUNCATE', 'PERSIST', 'MEMORY', 'WAL', 'OFF' );
+		if ( defined( 'SQLITE_JOURNAL_MODE' ) && in_array( SQLITE_JOURNAL_MODE, $valid_journal_modes, true ) ) {
```
adamziel, line 426: How easy/hard would if be to avoid referencing any global constants and variables in this driver? Ideally we could use it outside of WordPress as a drop-in PDO replacement and it's not easy to see all the configuration options when we rely on those constants. Can we source this information from a constructor argument? If we need a bit of dedicated code to connect it with WordPress, we could have a separate function that glues all the WordPress-specific bits with `new WP_SQLite_Driver()`

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+	 * Method to return the number of rows affected.
+	 */
+	public function get_affected_rows() {
+		return $this->affected_rows;
+	}
+
+	/**
+	 * Method to execute query().
+	 *
+	 * Divide the query types into seven different ones. That is to say:
```
adamziel, line 494: nice comment

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+			 */
+			$pre = apply_filters( 'pre_query_sqlite_db', null, $this, $statement, $mode, $fetch_mode_args );
+			if ( null !== $pre ) {
+				return $pre;
+			}
+		}
+		$this->pdo_fetch_mode = $mode;
+		$this->mysql_query    = $statement;
+		if (
+			preg_match( '/^\s*START TRANSACTION/i', $statement )
```
adamziel, line 556: Let's get rid of all the `preg_match()` calls now that we're dealing with ASTs. I know this one is likely a carry-over from the previous driver, but I can already think of a few queries it would mishandle:

```
START
TRANSACTION
```

```
START TRANSACTIONTIRIRIRIRIRRI
```

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+			throw new Exception( sprintf( 'Expected "query" node, got: "%s"', $ast->rule_name ) );
+		}
+
+		$children = $ast->get_child_nodes();
+		if ( count( $children ) !== 1 ) {
+			throw new Exception( sprintf( 'Expected 1 child, got: %d', count( $children ) ) );
+		}
+
+		$ast = $children[0]->get_child_node();
+		switch ( $ast->rule_name ) {
```
adamziel, line 843: This reads so much better than the old driver while having less intermediate state, fuzzy logic, and opportunity for error. Good job!

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+			'SELECT COLUMN_NAME FROM _mysql_information_schema_columns WHERE table_schema = ? AND table_name = ?',
+			array( $this->db_name, $table_name )
+		)->fetchAll( PDO::FETCH_COLUMN );
+
+		// Track column renames and removals.
+		$column_map = array_combine( $column_names, $column_names );
+		foreach ( $node->get_descendant_nodes( 'alterListItem' ) as $action ) {
+			$first_token = $action->get_child_token();
+
+			if ( WP_MySQL_Lexer::DROP_SYMBOL === $first_token->id ) {
```
adamziel, line 1034: These ifs are mutually exclusive – how about a `switch()`?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+					$new_name = $this->unquote_sqlite_identifier(
+						$this->translate( $action->get_child_node( 'identifier' ) )
+					);
+
+					$column_map[ $old_name ] = $new_name;
+				}
+			}
+		}
+
+		$this->information_schema_builder->record_alter_table( $node );
```
adamziel, line 1068: The `$column_map` idea is great

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+		);
+
+		// Save all column names from the original table.
+		$column_names = $this->execute_sqlite_query(
+			'SELECT COLUMN_NAME FROM _mysql_information_schema_columns WHERE table_schema = ? AND table_name = ?',
+			array( $this->db_name, $table_name )
+		)->fetchAll( PDO::FETCH_COLUMN );
+
+		// Track column renames and removals.
+		$column_map = array_combine( $column_names, $column_names );
```
adamziel, line 1030: The `$column_map` idea is great and the modifications below are super clear

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+		/*
+		 * See:
+		 *   https://www.sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes
+		 */
+
+		// 1. If foreign key constraints are enabled, disable them.
+		// @TODO
+
+		// 2. Create a new table with the new schema.
+		$tmp_table_name     = "_tmp__{$table_name}_" . uniqid();
```
adamziel, line 1079: nitpick: we'll need a reserved table prefix for any system operations, and we'll want to make sure explicit user queries won't affect any tables with that prefix

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+		}
+
+		// @TODO: Triggers and views.
+
+		$this->results      = 1;
+		$this->return_value = $this->results;
+		return;
+
+		/*
+		 * SQLite supports only a small subset of MySQL ALTER TABLE statement.
```
adamziel, line 1130: Great comment

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -0,0 +1,1777 @@
...
+
+		$this->set_results_from_fetched_data( $index_info );
+	}
+
+	private function execute_describe_statement( WP_Parser_Node $node ): void {
+		$table_name = $this->unquote_sqlite_identifier(
+			$this->translate( $node->get_child_node( 'tableRef' ) )
+		);
+
+		$column_info = $this->execute_sqlite_query(
```
adamziel, line 1234: nice!

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
adamziel: I want to re-surface the original concern as we seem to be still using `get_descendant_node` here

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
adamziel: makes sense, thanks for explaining!

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: Ah, this is a copy-paste from the original translator. To start with, I copy pasted all the main methods, helpers, and query state properties, excluding the translation logic. This made it easy to quickly try to run the tests with the same API, etc. It also calls the same hooks and keeps some of these debug helpers. I didn't yet look into why this is there and if it's needed. Maybe Git history will help here.

I will just quickly mark other comments regarding copy-pastes from the legacy translator, so that it's clear.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: Just marking that this is copy-pasted from the legacy translator. We should revise the API :+1:

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: Just marking that this is copy-pasted from the legacy translator. I don't know, I'll check it out. Also strange that it uses `SQLite3`, while in other cases, we're using `PDO`.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: Just marking that this is copy-pasted from the legacy translator. I didn't even notice this one. Indeed looks pretty strange 🤔

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: Just marking that this is copy-pasted from the legacy translator. I'm not yet sure how exactly this works. If we're trying to turn on the FKs, then we could just always do `PRAGMA foreign_keys = ON` without checking the current state first, I guess.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: Just marking that this is copy-pasted from the legacy translator. It will be definitely great to refactor this to an explicit set of config options 👍

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: Even this one is from the legacy translator. So the comment won't be fully up to date, I guess and needs to be improved.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: Just marking that this is copy-pasted from the legacy translator. We should definitely fix that 💯

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -995,8 +995,115 @@ private function execute_create_table_statement( WP_Parser_Node $node ): void {
...
+		// 2. Create a new table with the new schema.
+		$tmp_table_name     = "_tmp__{$table_name}_" . uniqid();
+		$queries            = $this->get_sqlite_create_table_statement( $table_name, $tmp_table_name );
+		$create_table_query = $queries[0];
+		$constraint_queries = array_slice( $queries, 1 );
+		$this->execute_sqlite_query( $create_table_query );
+
+		// 3. Copy data from the original table to the new table.
+		$this->execute_sqlite_query(
+			sprintf(
```
brandonpayton, line 1087: It looks like `execute_sqlite_query($sql, $params)` supports param binding. And it looks like we can even pass params to bind to specific names within the SQL string (e.g., `:new_table` and `:original_table`). (You may already know this about the PDOStatement interface that is used, but I did not :)

In this case, we've parsing the original SQL and maybe shouldn't be too vulnerable to SQL insertion exploits, but if we can use statement preparation to be sure, maybe we should.

JanJakes: @brandonpayton Unfortunately, prepared statements don't allow for parametrizing table and column names. But good point that these are not properly escaped and there is no `TODO`, so this will be a great reminder to fix them.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel I'm fixing this particular case by using `get_child_node` calls, but I'll also add a TODO somewhere to address this more generally, I think it would be good to create a ticket to implement a better AST-querying API and use it everywhere.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: For this one, I'm adding a TODO, as we also need to replace `SQLite3` with `PDO` here.

adamziel:
Superseded by #179

PR #166, opened 2024-12-11
Fix datetime field with empty default

### LINKED ISSUE #165: Create table with empty default timestamp is incorrect
When we dump an SQLite database table with a timestamp that has a default set to the empty string, it produces an SQL dump that can't be inserted into a MySQL database.

The following table:

```
CREATE TABLE `test_table` (
  `timestamp1` datetime NOT NULL,
  `timestamp2` date NOT NULL,
  `timestamp3` time NOT NULL,
  `timestamp4` timestamp NOT NULL,
  `timestamp5` year NOT NULL
);
```

Is exported as:

```
CREATE TABLE `test_table` (
  `timestamp1` datetime NOT NULL DEFAULT '',
  `timestamp2` date NOT NULL DEFAULT '',
  `timestamp3` time NOT NULL DEFAULT '',
  `timestamp4` timestamp NOT NULL DEFAULT '',
  `timestamp5` year NOT NULL DEFAULT ''
);
```

**Done is:**
- [x] `DEFAULT ''` does not appear in produced table structure

wojtekn:
Fixes https://github.com/WordPress/sqlite-database-integration/issues/165

I propose to fix an issue where dumping the datetime field that has empty value results with producing incorrect MySQL.

### DIFF
--- a/wp-includes/sqlite/class-wp-sqlite-translator.php
+++ b/wp-includes/sqlite/class-wp-sqlite-translator.php
@@ -2659,13 +2659,9 @@ class WP_SQLite_Translator {
 			! $token->matches(
 				WP_SQLite_Token::TYPE_KEYWORD,
 				WP_SQLite_Token::FLAG_KEYWORD_RESERVED,
-				array( 'GROUP' )
+				array( 'GROUP BY' )
 			)
 		) {
-			return false;
-		}
-		$next = $this->rewriter->peek_nth( 2 )->value;
-		if ( 'BY' !== strtoupper( $next ?? '' ) ) {
 			return false;
 		}
 
@@ -2675,7 +2671,7 @@ class WP_SQLite_Translator {
 	}
 
 	/**
-	 * Translate WHERE something HAVING something to WHERE something AND something.
+	 * Translate HAVING without GROUP BY to GROUP BY 1 HAVING.
 	 *
 	 * @param WP_SQLite_Token $token The token to translate.
 	 *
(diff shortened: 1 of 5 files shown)


wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3858,6 +3859,22 @@ function ( $row ) use ( $name_map ) {
...
+	 * Checks if column should define the default.
+	 *
+	 * @param stdClass $column The table column
+	 * @param string $mysql_type The MySQL data type
+	 *
+	 * @return boolean If column should have a default definition.
+	 */
+	private function column_has_default( $column, $mysql_type ) {
+		if ( null !== $column->dflt_value && '' !== $column->dflt_value && ! in_array( strtolower( $mysql_type ), array( 'datetime', 'date', 'time', 'timestamp', 'year' ), true ) ) {
+			return true;
```
katinthehatsite, line 3872: I noticed that some CREATE_TABLE statements also include `expires` as datetime e.g.:

```
`expires` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
```
Although this one has a value and not an empty string, should we also include it into this array in case it might come up with an empty string? What do you think?

katinthehatsite:
I initially had a question but I figured it so I removed it (in case you got a notification 😅 ). The fix looks good to me. I looked through The MySQL data types in the backups from Studio and Jetpack and the fix seems to cover the cases that we ran into. 👍 

Out of curiosity, I see that the fix is covered with unit tests so there are no manual testing steps needed. If I were to test this manually with Studio, do you know what steps I would need to take? 🙇

wojtekn:
> Out of curiosity, I see that the fix is covered with unit tests so there are no manual testing steps needed. If I were to test this manually with Studio, do you know what steps I would need to take? 🙇

The easiest way would be:
1. Create and start Studio site
2. Copy the `wp-includes/sqlite/class-wp-sqlite-translator.php` file from this PR to `wp-content/mu-plugins/sqlite-database-integration/wp-includes` in your site
3. Run Export or Push

katinthehatsite:
Nice thanks!

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3858,6 +3859,22 @@ function ( $row ) use ( $name_map ) {
...
+		}
```
JanJakes, line 3873: @wojtekn This seems to remove all defaults for date/time fields, including valid defaults, such as `col YEAR DEFAULT '2024'`. Could we add a test case with a valid default for each of these data types, and ensure that it is preserved?

My guess is that the problem could be that the original condition, despite looking fine (`'' !== $column->dflt_value`), might be missing the fact that the default can include a quoted string, so it should be rather something like `'' !== trim( $column->dflt_value, "'" )`.

JanJakes: Thinking further, what's tricky, though, is that the `''` value as a default is valid for string fields. So I guess we need to trim the quotes only for the case of non-string or date/time fields?

wojtekn: Good catch. I untangled the conditions to handle all cases and make it more readable in 86b43fdb1a07162533362a0e292335235c18fc5a

JanJakes (approved):
Looks great, thanks for the improvements!

PR #168, opened 2024-12-12
Fix unlimited key lengths issue

### LINKED ISSUE #167: Create table produces too long keys
When we dump an SQLite database table that had keys set on text fields with specific lengths set, it produces an SQL dump that can't be inserted into a MySQL database.

The following table is imported to SQLite:

```
CREATE TABLE `wp_wc_orders_meta` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `order_id` bigint(20) unsigned DEFAULT NULL,
  `meta_key` varchar(255) DEFAULT NULL,
  `meta_value` text DEFAULT NULL,
  PRIMARY KEY (`id`),
  KEY `meta_key_value` (`meta_key`(100),`meta_value`(82)),
  KEY `order_id_meta_key_meta_value` (`order_id`,`meta_key`(100),`meta_value`(82))
) ENGINE=InnoDB DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci
```

After exporting, key length properties are forgotten, and such a table can't be created on MySQL server anymore due to too long key:

```
CREATE TABLE `wp_wc_orders_meta` (
	`id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
	`order_id` bigint(20) unsigned DEFAULT NULL,
	`meta_key` varchar(255) DEFAULT NULL,
	`meta_value` text DEFAULT NULL,
	PRIMARY KEY (`id`),
	KEY `order_id_meta_key_meta_value` (`order_id`, `meta_key`, `meta_value`),
	KEY `meta_key_value` (`meta_key`, `meta_value`)
);
```

**Done is:**
- [ ] Produced create table query can be imported to MySQL

wojtekn:
Fixes https://github.com/WordPress/sqlite-database-integration/issues/167

I propose to fix an issue where dumping the table with keys that use multiple long fields produces SQL code that produces an error.

### DIFF
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -267,8 +267,8 @@ class WP_SQLite_Translator_Tests extends TestCase {
 				ID BIGINT PRIMARY KEY AUTO_INCREMENT NOT NULL,
 				option_name VARCHAR(255) default '',
 				option_value TEXT NOT NULL,
-				UNIQUE KEY option_name (option_name),
-				KEY composite (option_name, option_value)
+				UNIQUE KEY option_name (option_name(100)),
+				KEY composite (option_name(100), option_value(100))
 			);"
 		);
 
@@ -283,8 +283,8 @@ class WP_SQLite_Translator_Tests extends TestCase {
 	`option_name` varchar(255) DEFAULT '',
 	`option_value` text NOT NULL DEFAULT '',
 	PRIMARY KEY (`ID`),
-	KEY `composite` (`option_name`, `option_value`),
-	UNIQUE KEY `option_name` (`option_name`)
+	KEY `composite` (`option_name`(100), `option_value`(100)),
+	UNIQUE KEY `option_name` (`option_name`(100))
 );",
 			$results[0]->{'Create Table'}
 		);
@@ -337,8 +337,8 @@ class WP_SQLite_Translator_Tests extends TestCase {
 				ID BIGINT PRIMARY KEY AUTO_INCREMENT NOT NULL,
 				option_name VARCHAR(255) default '',
 				option_value TEXT NOT NULL,
-				UNIQUE KEY option_name (option_name),
-				KEY composite (option_name, option_value)
+				UNIQUE KEY option_name (option_name(100)),
+				KEY composite (option_name, option_value(100))
 			);"
 		);
 
@@ -353,8 +353,8 @@ class WP_SQLite_Translator_Tests extends TestCase {
 	`option_name` varchar(255) DEFAULT '',
 	`option_value` text NOT NULL DEFAULT '',
 	PRIMARY KEY (`ID`),
-	KEY `composite` (`option_name`, `option_value`),
-	UNIQUE KEY `option_name` (`option_name`)
+	KEY `composite` (`option_name`(100), `option_value`(100)),
+	UNIQUE KEY `option_name` (`option_name`(100))
 );",
 			$results[0]->{'Create Table'}
 		);
@@ -418,8 +418,8 @@ class WP_SQLite_Translator_Tests extends TestCase {
 					ID BIGINT PRIMARY KEY AUTO_INCREMENT NOT NULL,
 					option_name VARCHAR(255) default '',
 					option_value TEXT NOT NULL,
-					KEY `option_name` (`option_name`),
-					KEY `double__underscores` (`option_name`, `ID`)
+					KEY `option_name` (`option_name`(100)),
+					KEY `double__underscores` (`option_name`(100), `ID`)
 				);"
 		);
 
@@ -428,8 +428,8 @@ class WP_SQLite_Translator_Tests extends TestCase {
 					ID BIGINT PRIMARY KEY AUTO_INCREMENT NOT NULL,
 					option_name VARCHAR(255) default '',
 					option_value TEXT NOT NULL,
-					KEY `option_name` (`option_name`),
-					KEY `double__underscores` (`option_name`, `ID`)
+					KEY `option_name` (`option_name`(100)),
+					KEY `double__underscores` (`option_name`(100), `ID`)
 				);"
 		);
 	}
@@ -440,8 +440,8 @@ class WP_SQLite_Translator_Tests extends TestCase {
 					ID BIGINT PRIMARY KEY AUTO_INCREMENT NOT NULL,
 					option_name VARCHAR(255) default '',
 					option_value TEXT NOT NULL,
-					KEY `option_name` (`option_name`),
-					KEY `double__underscores` (`option_name`, `ID`)
+					KEY `option_name` (`option_name`(100)),
+					KEY `double__underscores` (`option_name`(100), `ID`)
 				);"
 		);
 
@@ -455,8 +455,8 @@ class WP_SQLite_Translator_Tests extends TestCase {
 	`option_name` varchar(255) DEFAULT \'\',
 	`option_value` text NOT NULL DEFAULT \'\',
 	PRIMARY KEY (`ID`),
-	KEY `double__underscores` (`option_name`, `ID`),
-	KEY `option_name` (`option_name`)
+	KEY `double__underscores` (`option_name`(100), `ID`),
+	KEY `option_name` (`option_name`(100))
 );',
 			$results[0]->{'Create Table'}
 		);
@@ -486,8 +486,8 @@ class WP_SQLite_Translator_Tests extends TestCase {
 	`meta_key` varchar(255) DEFAULT NULL,
 	`meta_value` text DEFAULT NULL,
 	PRIMARY KEY (`id`),
-	KEY `order_id_meta_key_meta_value` (`order_id`,`meta_key`(100),`meta_value`(82)),
-	KEY `meta_key_value` (`meta_key`(100),`meta_value`(82))
+	KEY `order_id_meta_key_meta_value` (`order_id`, `meta_key`(100), `meta_value`(100)),
+	KEY `meta_key_value` (`meta_key`(100), `meta_value`(100))
 );',
 			$results[0]->{'Create Table'}
 		);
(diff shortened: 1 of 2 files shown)


tests/WP_SQLite_Translator_Tests.php
```diff
@@ -421,6 +421,37 @@ public function testShowCreateTablePreservesDoubleUnderscoreKeyNames() {
...
+		$results = $this->engine->get_query_results();
+		$this->assertEquals(
+			'CREATE TABLE `_tmp__table` (
+	`id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+	`order_id` bigint(20) unsigned DEFAULT NULL,
+	`meta_key` varchar(255) DEFAULT NULL,
+	`meta_value` text DEFAULT NULL,
+	PRIMARY KEY (`id`),
+	KEY `order_id_meta_key_meta_value` (`order_id`,`meta_key`(100),`meta_value`(82)),
+	KEY `meta_key_value` (`meta_key`(100),`meta_value`(82))
```
wojtekn: @JanJakes, currently, it skips all lengths.

As we don't store the lengths, do you have any idea how to approach that in a hotfix? Assume 100 for any varchar or text fields?

JanJakes: @wojtekn Using something like 100 sounds like a reasonable hotfix. Maybe we could go for 191, as [WordPress seems to use that value](https://github.com/WordPress/wordpress-develop/blob/9104190b6d36c2959f652e0b65bf5bb7af42ffcc/src/wp-admin/includes/schema.php#L53).

katinthehatsite: When I was fixing these CREATE TABLE statements manually to attempt using it for import in Jetpack, I used `191` for length as well and that fixed the issue for me and I was able to work with these statements. This was an example:

```
DROP TABLE IF EXISTS `wp_wc_orders_meta`;
CREATE TABLE `wp_wc_orders_meta` (
	`id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
	`order_id` bigint(20) unsigned DEFAULT NULL,
	`meta_key` varchar(255) DEFAULT NULL,
	`meta_value` text DEFAULT NULL,
	PRIMARY KEY (`id`),
	KEY `order_id_meta_key_meta_value` (`order_id`, `meta_key`, `meta_value`(191)),
	KEY `meta_key_value` (`meta_key`, `meta_value`(191))
);
```

wojtekn: > @wojtekn Using something like 100 sounds like a reasonable hotfix. Maybe we could go for 191, as [WordPress seems to use that value](https://github.com/WordPress/wordpress-develop/blob/9104190b6d36c2959f652e0b65bf5bb7af42ffcc/src/wp-admin/includes/schema.php#L53).

Thanks for sharing @JanJakes . It seems they calculated 191 assuming the limit of 767 bytes and `utf8mb4`, which uses 4 bytes per character. It works for the index with one field, but if we used such a value for an index that has more fields, it wouldn't work.

Currently, at least on WoA sites, the limit is 3072 bytes, and the collation used is `latin1_swedish_ci`, which uses 1 byte per character.  I played with the queries shown in the issue description, and it allowed me to add an index when I capped the text field length to 2809 characters. With one more, it failed:

```
	KEY `order_id_meta_key_meta_value` (`order_id`, `meta_key`, `meta_value`(2810))
```

However, if collation were set to `utf8mb4`, I would need to set it to 511 instead. If the limit for the index was 767 bytes, the `meta_key` alone would exceed it.

I propose going with an arbitrary value of 100 to keep things simple. With that one, for any varchar and text field, for given settings, it could support:
- `utf8mb4` and limit of 767 bytes: one text/varchar field plus a few integers
- `utf8mb4` and limit of 3072 bytes: seven text/varchar fields plus a few integers
- `latin1_swedish_ci` and limit of 767 bytes: seven text/varchar fields plus a few integers
- `latin1_swedish_ci` and limit of 3072 bytes: 30 text/varchar fields plus a few integers

If we dropped it to 80, in the first, worst-case scenario, it would support two text/varchar fields and a few integers. Thoughts?

wojtekn: @katinthehatsite, according to my tests, values higher than 191 would make it work for you, too.

wojtekn:
It seems we are not there yet. The current implementation results in producing an incorrect `CREATE TABLE` as the key field length can't be longer than field length itself:

```
DROP TABLE IF EXISTS `_tmp_table`;
CREATE TABLE `_tmp_table` (
	`field` varchar(20) NOT NULL DEFAULT 'Y',
	KEY `field` (`field`(100))
);
```

tests/WP_SQLite_Translator_Tests.php
JanJakes: @wojtekn

> It works for the index with one field, but if we used such a value for an index that has more fields, it wouldn't work.

Ah, right!

> I propose going with an arbitrary value of 100 to keep things simple.
> If we dropped it to 80, in the first, worst-case scenario, it would support two text/varchar fields and a few integers.

That sounds reasonable to me. I think 80 is fine as well. Unless some string values are expected to share very long prefixes, I actually don't see much need to index that much. That said, WP tries to make use of the 191 limit, so pushing it to something like 80 or 100 might be the way to go.

On the other hand, if we expect some plugins to index 3 string fields (which I'm sure can happen), then we may even go with something like 50. After all, these numbers are the maximums of what can be indexed, and I would expect the actual values to be much shorter in most cases.

wojtekn:
The issue mentioned in previous comment is fixed.

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3755,7 +3756,20 @@ private function get_key_definitions( $table_name, $columns ) {
 			$key_definition[] = sprintf( '`%s`', $index_name );
 
 			$cols = array_map(
-				function ( $column ) {
+				function ( $column ) use ( $table_name, $key_length_limit ) {
+					$data_type   = strtolower( $this->get_cached_mysql_data_type( $table_name, $column['name'] ) );
+					$data_length = $key_length_limit;
+
+					// Extract the length from the data type. Make it lower if needed.
+					if ( 1 === preg_match( '/^(\w+)\((\d+)\)$/', $data_type, $matches ) ) {
```
JanJakes: @wojtekn I think this would fail to capture types with spaces, such as `varchar  (20)`, `character varying(20)`, etc. I'm not sure if they are normalized when saved.

Not too important, but I guess it can be done without a regex with something like `(int) (explode( '(', $data_type )[ 1 ] ?? '' )`, which would return `> 0` when the length exists and `0` otherwise.

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3755,7 +3756,20 @@ private function get_key_definitions( $table_name, $columns ) {
...
+						$data_type   = $matches[1]; // "varchar"
+						$data_length = min( $matches[2], $key_length_limit ); // "255"
+					}
+
+					// Set the data length to the varchar and text key length
+					if ( 'text' === $data_type || str_starts_with( $data_type, 'varchar' ) ) {
```
JanJakes: This is likely to be the case as well for all the variants of `text` (`tinytext`, `text`, `mediumtext`, `longtext`), as well as `blob` (`tinyblob`, `blob`, `mediumblob`, `longblob`), and `varbinary`. With the binary data, the limits are in bytes, so the space is potentially bigger, but it's probably fine to cap them at 100 too.

With `char` and `binary` it seems to work fine, as they are limited to the length of 255 (`tinytext` and `tinyblob` are too, but they still require limiting the index).

Since some of the data types can be written in many ways (`char varying`, `character varying`, `varchar`, `national varchar`, `nvarchar`, `nchar varchar`, `national char varying`, `nchar varying`, `long char varying`, `long character varying`, `long varchar`, ...), and there are all the `text` and `blob` variants, we could probably just check whether the type contains one of the following substrings — `varchar`, `varying`, `text`, `blob`, `varbinary`.

JanJakes (commented):
This looks good to me, but we could try to follow more nuances of the syntax and more types, if that's not too much trouble.

fredrikekelund:
I addressed the final review comments in https://github.com/WordPress/sqlite-database-integration/pull/169. I know it's clunky to open a new PR, but I couldn't find a better way to put up my commits.

wojtekn:
@fredrikekelund, thanks for opening another PR. What if we used `$field_types_translation` to make the condition more generic? I've updated this PR.

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3755,7 +3756,25 @@ private function get_key_definitions( $table_name, $columns ) {
 			$key_definition[] = sprintf( '`%s`', $index_name );
 
 			$cols = array_map(
-				function ( $column ) {
+				function ( $column ) use ( $table_name, $key_length_limit ) {
+					$data_type   = strtolower( $this->get_cached_mysql_data_type( $table_name, $column['name'] ) );
+					$data_length = $key_length_limit;
+
+					// Extract the length from the data type. Make it lower if needed. Skip 'unsigned' parts and whitespace.
+					if ( 1 === preg_match( '/^(\w+)\s*\(\s*(\d+)\s*\)\s*(.*)?$/', $data_type, $matches ) ) {
```
JanJakes: This is just a tiny nitpick:
- I guess the trailing `(.*)?` is equivalent to just `.*`, or we can simply drop the `$` as anchoring the beginning is enough.
- The `Skip 'unsigned' parts` piece of the comment is probably rather irrelevant, as we're not trying to match integers.

wojtekn: Thanks for ideas, I cleaned up this regex more.

JanJakes (approved):
Thank you!

PR #196, opened 2025-06-04
Implement `ADD INDEX` and `DROP INDEX`

JanJakes:
This PR implements `ADD INDEX` and `DROP INDEX` statements in the new driver.

It was necessary to reorganize some bits of the information schema builder, as indexes and constraints overlap in both the grammar (ALTER TABLE uses a `tableConstraintDef` node for both indexes and constraints), and in reality (UNIQUE index needs a constraint, UNIQUE constraint needs an index, making them equivalent).


adamziel:
Let's add some more test cases, e.g. composite unique index with mixed data types, alternating parentheses and no parentheses (`colname(14), col2)`, column ordering (`CREATE INDEX reverse_name_idx ON companies (name DESC)`) etc. Let's see how far we can stretch the syntax.

JanJakes:
@adamziel Good point! I also forgot about `IF NOT EXISTS` and `IF EXISTS`, I will add those as well.

adamziel:
It's also fine to skip some things in this PR as long as we document the limitations. My only concern is merging "ADD INDEX support" without realizing we don't support this or that case.

JanJakes:
> @adamziel Good point! I also forgot about IF NOT EXISTS and IF EXISTS, I will add those as well.

I take this back 😄 Neither of these seem to be supported:
- https://dev.mysql.com/doc/refman/9.3/en/create-index.html
- https://dev.mysql.com/doc/refman/9.3/en/drop-index.html

JanJakes:
@adamziel It turns out `DESC` indexes were not fully implemented. Thanks for the catch! I've also added more complex tests and better assertions.

adamziel (approved):

Issue #194, opened 2025-06-02
Changing the database backend on a live site


JanJakes:
The plugin could support migrating from SQLite to MySQL and the other way around. It would:

* Help ship SQLite support in WordPress core
* Encourage people worried about staying locked in SQLite
* Help move the sites around

Technically, there's a streaming Data Liberation part to it where we track progress, handle deltas etc. Maybe it's a [site sync](https://github.com/adamziel/zs-sync) within one site? Regardless of the implementation, the SQLite plugin could ship the UI.

aristath:
I have a rudimentary migration script on https://github.com/aristath/sqlite-database-integration/pull/1/files
It's pretty basic, but it works - could be used as a starting point for any implementation you may want to build 👍

szabozoltan69:
I have this plugin on a site, and the Plugins admin page offered me the upgrade. I upgraded, and it broke the site, due to previously I did not have 
define( 'DB_NAME', '...' );
in wp-config.php. After inserting this (with an irrelevant content), the page was working again.
(Now I see that this is needed for the mysql line, so it's ok.)
The error message was:
```
Error: Undefined constant "DB_NAME" in .../wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/db.php on line 69
```

JanJakes:
@szabozoltan69 Oh, thanks for the report! I didn't notice the `DB_NAME` requirement also affected the current (legacy) driver. I think we can fix that, although the new driver will require the constant to be set.

PR #197, opened 2025-06-06
Require `DB_NAME` to be set only in the new driver

JanJakes:
Require `DB_NAME` to be set only in the new driver.

As reported in https://github.com/WordPress/sqlite-database-integration/issues/194#issuecomment-2939491161:

> I have this plugin on a site, and the Plugins admin page offered me the upgrade. I upgraded, and it broke the site, due to previously I did not have define( 'DB_NAME', '...' ); in wp-config.php.

The error message was:

> Error: Undefined constant "DB_NAME" in .../wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/db.php on line 69


adamziel:
In the new message let's explain it's about the DB_NAME constant and why do we need it in SQLite

adamziel (approved):

JanJakes:
@adamziel :+1: I improved the comment.

PR #199, opened 2025-06-11
Make nullable $previous optional in class-wp-sqlite-information-schema-exception.php

adamziel:
Adds `?` to `?Throwable previous = null` to fix this deprecation warning:

[11-Jun-2025 18:07:04 UTC] PHP Deprecated:  WP_SQLite_Information_Schema_Exception::__construct(): Implicitly marking parameter $previous as nullable is deprecated, the explicit nullable type must be used instead in /internal/shared/sqlite-database-integration/wp-includes/sqlite-ast/class-wp-sqlite-information-schema-exception.php on line 38

cc @draganescu – are there any other deprecation warnings you've seen?


draganescu (approved):
Seems to do what the error says we should do 🤝

PR #202, opened 2025-07-02
Document a known limitation regarding affected row count

JanJakes:
This PR adds a file to document known SQLite driver limitations, and it documents the first one.

### DIFF
--- a/docs/known-limitations.md
+++ b/docs/known-limitations.md
@@ -1,0 +1,19 @@
+# Known limitations
+This file documents the known limitations of the SQLite driver that are difficult
+to address and are not currently planned to be resolved.
+
+### Number of affected rows when values are not changed
+In some cases, the number of affected rows reported by the SQLite driver may
+differ from the number returned by MySQL. This is because SQLite always reports
+the number of rows matched by a statement, regardless of whether they were
+actually changed or set to the same values as they already held.
+
+Example:
+
+```sql
+CREATE TABLE t (id INT);
+INSERT INTO t (id) VALUES (1);
+UPDATE t SET id = 1;
+```
+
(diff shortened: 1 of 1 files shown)


adamziel:
Would a SELECT query within the same, isolated transaction give us the number of rows matched but not updated?

adamziel (approved):

JanJakes:
@adamziel I think so, yes, but in SQLite, that seems to equal the number from `SELECT CHANGES()` and the row cound from PDO APIs.

adamziel:
We'd need to add a condition such as `SELECT * FROM target_table WHERE ($original_update_conditions)` to get the number of matching rows and a condition such as `SELECT * FROM target_table WHERE ($original_update_conditions) AND NOT (field1 = :updated_value1 OR field2 = :updated_value2)` to get the number of changed rows. It is a fair limitation to have, but it also seems like a tractible problem.

JanJakes:
@adamziel Ah, right, I see what you mean now. This approach could likely be applied directly to the UPDATE by rewriting it to directly exclude the rows where the existing values are equal to the new ones:

```sql
UPDATE ... SET col1 = val1, col2 = val2 WHERE ... AND (col1 != val1 OR col2 != val2)
```

Then the number of affected rows should be correct, but maybe the caveat is that in edge cases like with triggers, this could get tricky. That said, we don't support MySQL triggers and can probably never fully support them.

adamziel:
Really good idea to fold this directly into the update query.

It's fine to disregard triggers here. They're not a big priority at this point. We could leave a comment to document what needs to be done if we ever get to support triggers.

JanJakes:
@adamziel In that case, I think I can close this PR and create an issue to implement this into the `UPDATE` query instead.

PR #203, opened 2025-07-03
Use default value when `DB_NAME` constant is not defined

JanJakes:
This is needed for integration of the new SQLite driver in Studio.

We discovered that Studio doesn't use the [default implemented in Playground](https://github.com/WordPress/wordpress-playground/blob/bf9c3fa4af0aa3dca4788569a4a0fad81e5fbc54/packages/playground/wordpress/src/boot.ts#L164), as it's using `wp-now` and [booting Playground differently](https://github.com/Automattic/wordpress-playground-private/pull/140).

As a hotfix, when there is no `DB_NAME` defined at all, we'll fall back to the same value that Studio is using through an [MU plugin with polyfills](https://github.com/WordPress/wordpress-playground/blob/bf9c3fa4af0aa3dca4788569a4a0fad81e5fbc54/packages/playground/wordpress/src/index.ts#L22).

A proper fix will be to make the SQLite driver independent of the `DB_NAME` constant.

### DIFF
--- a/constants.php
+++ b/constants.php
@@ -52,6 +52,16 @@ if ( ! defined( 'FQDB' ) ) {
 	}
 }
 
+/**
+ * The DB_NAME constant is required by the SQLite driver.
+ *
+ * When DB_NAME is not defined, let's use a default value that corresponds to
+ * the default value of the constant in "wp-config-sample.php" and in Studio.
+ */
+if ( ! defined( 'DB_NAME' ) ) {
+	define( 'DB_NAME', 'database_name_here' );
+}
+
 // Allow enabling the SQLite AST driver via environment variable.
 if ( ! defined( 'WP_SQLITE_AST_DRIVER' ) && isset( $_ENV['WP_SQLITE_AST_DRIVER'] ) && 'true' === $_ENV['WP_SQLITE_AST_DRIVER'] ) {
 	define( 'WP_SQLITE_AST_DRIVER', true );


constants.php
```diff
@@ -52,6 +52,16 @@
...
 }
 
+/**
+ * The DB_NAME constant is required by the SQLite driver.
+ *
+ * When DB_NAME is not defined, let's use a default value that corresponds to
+ * the default value of the constant in "wp-config-sample.php" and in Studio.
+ */
+if ( ! defined( 'DB_NAME' ) ) {
+	define( 'DB_NAME', 'database_name_here' );
```
adamziel, line 62: log a warning maybe? or would that not make sense?

adamziel (approved):

constants.php
JanJakes: @adamziel At the moment, this seems to be a valid use case in Studio that probably shouldn't even generate warnings, I guess.

That said, I think I figured out a way to make the information schema `DB_NAME` independent, so I think this will be only temporary. The idea is to use temporary views on top of the stored information schema tables. They are naturally per-session and enable injecting any `DB_NAME` to the affected columns, they are low cost to create (they don't materialize + it can be lazy), and they could also unlock features like row count and auto-increment values.

adamziel: This code is not studio-specific, though, but distributed to all sites where the SQLite plugin is installed. Would that make a warning reasonable?

wojtekn: @adamziel the issue is not Studio-specific. It could happen on WoA sites, too, if user uses WordPress with SQLite plugin. WoA sites don't have database constants defined in wp-config.php file.

wojtekn (approved):

constants.php
adamziel: Makes sense. I guess it makes sense for any sqlite-only site to skip the database constants.

PR #205, opened 2025-07-04
Don't require WordPress to be fully loaded in WP CLI

JanJakes:
When working on new driver integration in Studio, @wojtekn encountered problems in CLI usage via [wp-cli-sqlite-command](https://github.com/Automattic/wp-cli-sqlite-command/).

The issue is that when `WP_SQLite_Information_Schema_Reconstructor` detect that it runs in a WordPress environment, it tries to load WP table definitions using `wp_get_db_schema()`, which requires `$wpdb`. However, WP CLI commands that are run `before_wp_load` and `after_wp_config_load` don't have `$wpdb` set, and doing so in the reconstructor turned out to be a rabbit hole of import and WP load issues.

I think it's reasonable to fix this special case by falling back to the non-WordPress-specific approach of reconstructing the information schema. I can only occur when `wp sqlite export` is used with a site that was created with the old driver and was not started yet in a server environment (in which case the optimal reconstruction would occur).

In summary, for sites created with the old driver:
- If someone runs the site, the information schema will recover the exact WP table info from `wp_get_db_schema`.
- If the site is off, and someone runs `wp sqlite export`, the information schema and exported SQL will be more similar to the old driver export (less detailed definition for WP tables).

This behavior will also correspond to potentially using the driver via `mysqldump` with MySQL-to-SQLite proxy one day, which is what the WP-native `wp db export` does.

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-sqlite-information-schema-reconstructor.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-information-schema-reconstructor.php
@@ -168,9 +168,25 @@ class WP_SQLite_Information_Schema_Reconstructor {
 	 * @return array<string, WP_Parser_Node> The WordPress CREATE TABLE statements.
 	 */
 	private function get_wp_create_table_statements(): array {
+		// Bail out when not in a WordPress environment.
 		if ( ! defined( 'ABSPATH' ) ) {
 			return array();
 		}
+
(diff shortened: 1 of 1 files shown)


adamziel (approved):

wojtekn:
Thanks for fixing the issue. Works fine.

PR #209, opened 2025-07-04
Implement column metadata

### LINKED ISSUE #190: Provide column info
WPDB exposes a [get_col_info() method](https://github.com/WordPress/wordpress-develop/blob/d71f29ff0899c6eb29bc037a7c521a966405cb35/src/wp-includes/class-wpdb.php#L3884) that retrieves column metadata from the last query using `mysqli_num_fields` and `mysqli_fetch_field`.

PDO statement provides analogous methods [columnCount](https://www.php.net/manual/en/pdostatement.columncount.php) and [getColumnMeta](https://www.php.net/manual/en/pdostatement.getcolumnmeta.php), but in the case of the SQLite driver, we may sometimes execute multiple queries for a single MySQL query, so the column info may need to be reconstructed more manually in some cases.

It may also happen that the returned column info needs to be "translated" from SQLite data to MySQL data using the information stored in the `information_schema` tables.

This also relates to https://github.com/Automattic/sqlite-database-integration/issues/7 regarding how the column info should be exposed.

JanJakes:
This PR implements emulation of `PDOStatement::getColumnMeta()` values and adds support for `$wpdb::get_col_info()`.

This feature is necessary for support of MySQL tools like Adminer, phpMyAdmin, and others, as well as for any plugins that may rely on this functionality.

Resolves https://github.com/WordPress/sqlite-database-integration/issues/190.

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -301,57 +301,57 @@ class WP_SQLite_Driver {
 	 * to MySQL "PDOStatement::getColumnMeta()" data types in the "native_type"
 	 * field, as well as the "len" and "precision" fields, where applicable:
 	 *
-	 *     <mysql-column-type> => array( <native_type>, <len>, <precision> )
+	 *     <mysql-column-type> => array( <native_type>, <mysqli_type>, <len>, <precision> )
 	 *
 	 * This is used to compute the column metadata from the information schema.
 	 */
 	const COLUMN_INFO_MYSQL_TO_NATIVE_TYPES_MAP = array(
 		// Numeric data types:
-		'bit'             => array( 'BIT', 1, 0 ),
-		'tinyint'         => array( 'TINY', 4, 0 ),
-		'smallint'        => array( 'SHORT', 6, 0 ),
-		'mediumint'       => array( 'INT24', 9, 0 ),
-		'int'             => array( 'LONG', 11, 0 ),
-		'bigint'          => array( 'LONGLONG', 20, 0 ),
-		'float'           => array( 'FLOAT', 12, 31 ),
-		'double'          => array( 'DOUBLE', 22, 31 ),
-		'decimal'         => array( 'NEWDECIMAL', null, null ),
+		'bit'             => array( 'BIT', 16, 1, 0 ),
+		'tinyint'         => array( 'TINY', 1, 4, 0 ),
+		'smallint'        => array( 'SHORT', 2, 6, 0 ),
+		'mediumint'       => array( 'INT24', 9, 9, 0 ),
+		'int'             => array( 'LONG', 3, 11, 0 ),
+		'bigint'          => array( 'LONGLONG', 8, 20, 0 ),
+		'float'           => array( 'FLOAT', 4, 12, 31 ),
+		'double'          => array( 'DOUBLE', 5, 22, 31 ),
+		'decimal'         => array( 'NEWDECIMAL', 246, null, null ),
 
 		// String data types:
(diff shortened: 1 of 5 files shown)


JanJakes:
@adamziel I've got this to a point where I think we can merge it, including plenty of tests. It would be useful to advance the Adminer, phpMyAdmin, and other tool support.

I can also create a follow-up ticket for what can be further improved (MySQLi flags, etc.). Maybe also all the `mysql-type-to-something` maps in the driver could be refactored to a single map aggregating all needed properties for each type, etc., but it's an implementation detail, and I don't want to delay this PR on that.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -294,6 +294,92 @@ class WP_SQLite_Driver {
...
+	 * This maps normalized MySQL column data types (as per information schema)
+	 * to MySQL "PDOStatement::getColumnMeta()" data types in the "native_type"
+	 * field, as well as the "len" and "precision" fields, where applicable:
+	 *
+	 *     <mysql-column-type> => array( <native_type>, <mysqli_type>, <len>, <precision> )
+	 *
+	 * This is used to compute the column metadata from the information schema.
+	 */
+	const COLUMN_INFO_MYSQL_TO_NATIVE_TYPES_MAP = array(
+		// Numeric data types:
```
adamziel, line 309: can we sort it by mysqli type within each datatype group (int, string etc)? It seems like we'll be looking up entries in this mapping by that number a little bit. Alternatively, we could create a bunch of named constants and reference them here, e.g. self::MYSQLI_BIT etc.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -294,6 +294,92 @@ class WP_SQLite_Driver {
...
+	 * the "len" and "precision" fields, where applicable:
+	 *
+	 *     <sqlite-column-definition-type> => array( <native_type>, <mysqli_type>, <len>, <precision> )
+	 *     <sqlite-column-meta-type>       => array( <native_type>, <mysqli_type>, <len>, <precision> )
+	 *
+	 * This is used to compute the MySQL column metadata for non-column fields
+	 * that have no records in the information schema (i.e., expressions).
+	 */
+	const COLUMN_INFO_SQLITE_TO_NATIVE_TYPES_MAP = array(
+		'NULL'    => array( 'NULL', 6, 0, 0 ),
```
adamziel, line 373: ditto for those numbers

adamziel:
Nice! Yeah let's get it in. The limitations would be also nice to document in a comment within the code as that's where we'll be looking for this information.

adamziel (approved):

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel As we already have multiple maps in the form of `mysql-type-to-something`, I was thinking about consolidating them into something like this:

Screenshot 2025-10-02 at 9 54 39

It would be all in one place then, and likely more readable. Usage:

```php
$sqlite_type      = self::MYSQL_TYPE_MAP[$type][self::KEY_SQLITE_TYPE];
$implicit_default = self::MYSQL_TYPE_MAP[$type][self::KEY_MYSQL_IMPLICIT_DEFAULT];
...
```

What do you think?

adamziel: That sounds good!

JanJakes: Ticket: https://github.com/WordPress/sqlite-database-integration/issues/259

PR #213, opened 2025-07-10
Implement `INFORMATION_SCHEMA.SCHEMATA`, `SHOW DATABASES`; verify DB names

JanJakes:
This PR reverts https://github.com/WordPress/sqlite-database-integration/pull/203, where providing the `DB_NAME` was made optional, and implements additional checks for the correctness of the database name.

* Implements the INFORMATION_SCHEMA.SCHEMATA table to store the database name (allowing for at most one user database).
* Checking whether the current session database name is correct.
* Migration path from older versions.
* Implementing SHOW DATABASES.


JanJakes:
@adamziel Here's the DB name enforcing logic put back & improved to check for correctness. Everyone's AFK now, so I may just merge this tomorrow.

bgrgicak (approved):

JanJakes:
I was further thinking about this solution, and there's one thing I don't like — the use of the global variable, since MySQL global variables are not per database but per the whole instance.

That made me think — what's the native way of storing database names in MySQL? It's the `information_schema.schemata` and `SHOW DATABASES`. I will update it to store the DB name there; it will be more MySQL-native and future-proof (e.g., for multi-DB support). We can now have the limitation that only `information_schema` and `DB_NAME` can be stored there.

JanJakes:
> I was further thinking about this solution, and there's one thing I don't like — the use of the global variable, since MySQL global variables are not per database but per the whole instance.

> That made me think — what's the native way of storing database names in MySQL? It's the `information_schema.schemata` and `SHOW DATABASES`. I will update it to store the DB name there; it will be more MySQL-native and future-proof (e.g., for multi-DB support). We can now have the limitation that only `information_schema` and `DB_NAME` can be stored there.

I ended up changing this to:
- Implementing the `INFORMATION_SCHEMA.SCHEMATA` table to store the database name (allowing for at most one user database).
- Checking whether the current session database name is correct.
- Migration path from older versions.
- Implementing `SHOW DATABASES`.

I'm leaving it open for review, as this can wait for when we're back from AFKs.

tests/WP_SQLite_Driver_Metadata_Tests.php
```diff
@@ -61,7 +61,7 @@ public function testInformationSchemaTables() {
 				'CREATE_TIME'     => $result[0]->CREATE_TIME,
 				'UPDATE_TIME'     => null,
 				'CHECK_TIME'      => null,
-				'TABLE_COLLATION' => 'utf8mb4_general_ci',
+				'TABLE_COLLATION' => 'utf8mb4_0900_ai_ci',
```
adamziel, line 64: Why this collation?

JanJakes: It's the default in MySQL 8+. See also the last commit separately for some more context. The upside is more correct MySQL compatibility, the downside that such exports won't work on older MySQL and MariaDB - but that's already the case with MySQL 8 as well. The default on MySQL 5 seems to be 'latin1_swedish_ci', which is not great.

Issue #192, opened 2025-06-02
Charset and collation support


JanJakes:
Currently, the SQLite driver uses `utf8mb4` as the database charset, and it saves the table and column charsets and collation to the information schema, but it doesn't verify or consider them in any way.

The general idea is to check what charset and collation was specified and verify its compatibility with Unicode (SQLite). If we find a charset or collation that would lead to incorrect application behavior, we should throw an error.

For collations, we should also try to match them to SQLite-supported collations and apply them correctly. The current driver simply adds `COLLATE NOCASE` to all textual columns, because it's the default MySQL behavior, but we need to reflect what was defined in the table/column definition.

---

More particularly, @adamziel suggested:

- Throwing an exception if we see a collation referring to a different character set or comparison rules
- Adding an option like `enforce_utf8_charset` that defaults to false for all the other cases. When we see a mismatched encoding and the option is `true`, we'd log a warning and use a sensible default encoding when we see a query with a "salvageable" encoding or collation definition. When the option is `false`, we'd throw a fatal error explaining what happened and that there's an option you can use. It could also be called `strict_mode` or so.

What's "salvageable" is quite arbitrary and it's easier to say what isn't.

For example, an incompatible set of collation rule would lead to a very different application behavior and I'd just throw an error right away. [MySQL Collation doc page](https://dev.mysql.com/doc/refman/8.0/en/charset-collation-names.html) explains different parts of the collation suffix:

```
Suffix | Meaning
-- | --
_ai | Accent-insensitive
_as | Accent-sensitive
_ci | Case-insensitive
_cs | Case-sensitive
_ks | Kana-sensitive
_bin | Binary
```

When a table declares `latin1`, `utf16`, or anything that isn't `utf8`, we'd likely break the app by quietly using utf-8.

On the flip side, I think we're good to rewrite `utf8`, `utf8mb3`, and other similar variations as `utf8mb4` (when the option is set). [Unicode characters sets](https://dev.mysql.com/doc/refman/8.4/en/charset-unicode-sets.html) page lists deprecated character sets and recommends using `utf8mb4` instead. That would still change the application behavior, but I can't imagine a plugin that relies on collating up to 3 bytes from every UTF-8 characters and not the fourth byte. Similarly, `utf8mb4_general_ci` is deprecated in favor of `utf8mb4_unicode_ci` and I think we could treat them as the same charset.

[...]

PR #214, opened 2025-07-24
Better support for `SET` statements and MySQL variables

JanJakes:
This pull request implements **better support for MySQL `SET` statements**.

It also implements additional related functionality and resolves some related issues:

- **Depth-first traversal:** All node querying APIs use depth-first pre-order traversal. This traversal produces a natural ordering that corresponds to the original parsed input. It also aligns with CSS selector behavior. This is useful in situations when we need to reconstruct or retrieve a fragment of the original input with tokens in the correct order. It's used to preserve correct returned MySQL column names (see the next item).
- **Correct result column names:** When no explicit column alias is provided in a select item, the driver will now ensure that the returned column name is equivalent to what MySQL infers from the input. For example, if we translate `CONCAT('a', 'b')` to `('a' || 'b')`, we need to use the original `CONCAT('a', 'b')` string as the column name. In other words, `SELECT CONCAT('a', 'b')` will be translated to `SELECT ('a' || 'b') AS ’CONCAT('a', 'b')’`. This is also needed when selecting user and system variables (see the items below).
- **SET NAMES/CHARSET:**`SET NAMES`, `SET CHARSET`, and `SET CHARACTER SET` are a no-op now, with a `TODO` to implement some validation logic in https://github.com/WordPress/sqlite-database-integration/issues/192.
- **User variables:** Reading and writing of user variables is now implemented (`SET @my_var = ...`, `SELECT @my_var`).
- **System variables:** Reading and writing of session system variables is now implemented (`SET time_zone = ...`, `SET @@time_zone = ...`, , `SET @@session.time_zone = ...`, `SELECT @@time_zone`, `SELECT @@session.time_zone`, etc.). The variables are not yet interpreted in any way, and they are not merged with global system variables at the moment.

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -2089,6 +2089,20 @@ class WP_SQLite_Driver {
 			}
 
 			if (
+				$part instanceof WP_MySQL_Token
+				&& WP_MySQL_Lexer::NAMES_SYMBOL === $part->id
+			) {
+				// "SET NAMES ..." is a no-op for now.
+				// TODO: Validate charset compatibility with UTF-8.
+				//       See: https://github.com/WordPress/sqlite-database-integration/issues/192
+			} elseif (
+				$part instanceof WP_Parser_Node
+				&& 'charsetClause' === $part->rule_name
+			) {
+				// "SET CHARACTER SET ..." is a no-op for now.
+				// TODO: Validate charset compatibility with UTF-8.
+				//       See: https://github.com/WordPress/sqlite-database-integration/issues/192
+			} elseif (
 				$part instanceof WP_Parser_Node
 				&& (
 					'internalVariableName' === $part->rule_name
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -2214,6 +2214,23 @@ class WP_SQLite_Driver {
 			$value = $this->evaluate_expression( $value_node );
 		}
 
+		/*
+		 * Handle ON/OFF values. They are accepted as both strings and keywords.
+		 *
+		 * @TODO: This is actually variable-specific and depends on the its type.
+		 *        For example:
+		 *          SET autocommit = OFF;                   SELECT @@autocommit;                 -> 0
+		 *          SET autocommit = false;                 SELECT @@autocommit;                 -> 0
+		 *          SET session_track_gtids = OFF;          SELECT @@session_track_gtids;        -> OFF
+		 *          SET session_track_gtids = false;        SELECT @@session_track_gtids;        -> OFF
+		 *          SET updatable_views_with_limit = OFF;   ERROR 1231 (42000)
+		 *          SET updatable_views_with_limit = false; SELECT @@updatable_views_with_limit; -> NO
+		 */
+		$lowercase_value = strtolower( $value );
+		if ( 'on' === $lowercase_value || 'off' === $lowercase_value ) {
+			$value = 'on' === $lowercase_value ? 1 : 0;
+		}
+
 		if ( WP_MySQL_Lexer::SESSION_SYMBOL === $type ) {
 			if ( 'sql_mode' === $name ) {
 				$modes                  = explode( ',', strtoupper( $value ) );
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -2365,6 +2365,8 @@ class WP_SQLite_Driver {
 				throw $this->new_not_supported_exception(
 					sprintf( 'data type: %s', $child->get_value() )
 				);
+			case 'selectItem':
+				return $this->translate_select_item( $node );
 			case 'fromClause':
 				// FROM DUAL is MySQL-specific syntax that means "FROM no tables"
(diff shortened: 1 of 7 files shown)


wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2089,17 +2114,39 @@ private function execute_set_statement( WP_Parser_Node $node ): void {
 			}
 
 			if (
+				$part instanceof WP_MySQL_Token
+				&& WP_MySQL_Lexer::NAMES_SYMBOL === $part->id
+			) {
+				// "SET NAMES ..." is a no-op for now.
+				// TODO: Validate charset compatibility with UTF-8.
```
adamziel, line 2121: should we call _doing_it_wrong to log a warning?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2141,15 +2188,55 @@ private function execute_set_system_variable_statement(
 			$type           = $var_ident_type->get_first_child_token()->id;
 		}
 
-		// Get the variable value.
-		$value = $this->translate( $value_node );
-		$value = str_replace( "''", "'", $value );
-		$value = substr( $value, 1, -1 );
+		/*
+		 * Some MySQL system variables values can be set using an unquoted pure
```
adamziel, line 2192: tricky! TIL

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2141,15 +2188,55 @@ private function execute_set_system_variable_statement(
...
+		} elseif ( ! $value_node->has_child_node( 'expr' ) ) {
+			$value = $this->unquote_sqlite_identifier( $this->translate( $value_node ) );
+		} else {
+			$value = $this->evaluate_expression( $value_node );
+		}
+
+		/*
+		 * Handle ON/OFF values. They are accepted as both strings and keywords.
+		 *
+		 * @TODO: This is actually variable-specific and depends on the its type.
```
adamziel, line 2220: tricky! oh, MySQL 🤦

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2162,6 +2249,26 @@ private function execute_set_system_variable_statement(
...
+	 * @throws WP_SQLite_Driver_Exception    When the query execution fails.
+	 */
+	private function execute_set_user_variable_statement(
+		WP_Parser_Node $user_variable,
+		WP_Parser_Node $expr
+	): void {
+		$name  = $this->unquote_sqlite_identifier(
+			$this->translate( $user_variable->get_first_child() )
+		);
+		$name  = strtolower( substr( $name, 1 ) ); // Remove '@', normalize case.
```
adamziel, line 2266: any chance we get a double `@@` here?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2244,6 +2351,33 @@ private function execute_administration_statement( WP_Parser_Node $node ): void
...
+	/**
+	 * Evaluate an expression and return the value, preserving its type.
+	 *
+	 * This is used to support expressions in SET statements for MySQL variables.
+	 *
+	 * @param  WP_Parser_Node $node The "expr" AST node.
+	 * @return mixed                The value of the expression.
+	 */
+	public function evaluate_expression( WP_Parser_Node $node ) {
+		// To support expressions, we'll use a SQLite query.
```
adamziel, line 2363: smart!

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel We'll need to validate it somehow, but I think that should be part of https://github.com/WordPress/sqlite-database-integration/issues/192. If we do it properly (getting a list of all commonly supported charsets and, for each one of them evaluating whether they are UTF-8 compatible or not), then maybe we can even throw an error. Or at least a warning.

We can't use directly `_doing_it_wrong` as that's WP-specific, but we want the driver to be more universal.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel I think that should be excluded by the grammar — single `@` is user variable, double `@` is system variable.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel Yeah... So to support the system variables better in the future, we'll need a list of them, their types, etc. And for some of them, we should implement whatever behavior they modify.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2959,6 +3098,72 @@ private function translate_datetime_literal( string $value ): string {
...
 	}
 
+	/**
+	 * Translate a select item to SQLite.
+	 *
+	 * In some cases, an explicit alias will be added to the select item, so that
+	 * the returned column name is always the same as it would be in MySQL.
+	 *
+	 * @param  WP_Parser_Node $node       The "selectItem" AST node.
+	 * @return string                     The translated value.
```
adamziel: Nitpick:

```suggestion
	 * @return string                     The translated expression.
```

adamziel (approved):
Great work, as always @JanJakes – thank you!

PR #219, opened 2025-07-25
[Query monitor] Tolerate null $value in boot.php

adamziel:
PR #217 added a line `global $wpdb` to `integration/query-monitor/boot.php` which triggered this warning in WordPress Playground multisite tests:

```
+ <b>Warning</b>:  Attempt to read property "option_value" on null in <b>/internal/shared/sqlite-database-integration/integrations/query-monitor/boot.php</b> on line <b>80</b><br />
```

Removing either `global $wpdb` or guarding against a null `$value` solves the failure. I assume there's a reason for adding `global $wpdb` so, in this PR, I'm adding a guard against a null value.

### DIFF
--- a/integrations/query-monitor/boot.php
+++ b/integrations/query-monitor/boot.php
@@ -69,17 +69,20 @@ if ( null === $wpdb->options ) {
 
 $query_monitor_active = false;
 try {
-	$value                = $wpdb->get_row(
+	$value = $wpdb->get_row(
 		$wpdb->prepare(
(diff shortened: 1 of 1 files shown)


JanJakes:
> Removing either `global $wpdb` or guarding against a null `$value` solves the failure. I assume there's a reason for adding `global $wpdb` so, in this PR, I'm adding a guard against a null value.

@adamziel I think the `global $wpdb` part is needed because I'm requiring the script from a function body (WP action). Removing `global` would cause `$wpdb` to always be `null` and thus an early return from the script.

PR #220, opened 2025-07-28
E2E tests

JanJakes:
This PR adds:
1. A job to run WordPress E2E tests with SQLite.
2. A job to run any other custom E2E tests, reusing the WordPress setup.
3. An E2E test for the Query Monitor plugin.

In the future, this can serve as the basis for more E2E tests, more 3rd-party plugins, etc.

### DIFF
--- a/tests/e2e/playwright.config.js
+++ b/tests/e2e/playwright.config.js
@@ -1,0 +1,27 @@
+/**
+ * External dependencies
+ */
+import path from 'node:path';
+import { defineConfig } from '@playwright/test';
+
+/**
+ * WordPress dependencies
+ */
+const baseConfig = require( '@wordpress/scripts/config/playwright.config' );
+
+process.env.WP_ARTIFACTS_PATH ??= path.join( process.cwd(), 'artifacts' );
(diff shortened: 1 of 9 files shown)


tests/e2e/playwright.config.js
```diff
@@ -0,0 +1,27 @@
...
+	process.env.WP_ARTIFACTS_PATH,
+	'storage-states/admin.json'
+);
+
+const config = defineConfig( {
+	...baseConfig,
+	globalSetup: require.resolve( '../../wordpress/tests/e2e/config/global-setup.js' ),
+	webServer: {
+		...baseConfig.webServer,
+		command: 'npm run env:start',
```
adamziel, line 23: It would be useful to run this both with wp-env and Playground CLI. Not a blocker here

adamziel (approved):

tests/e2e/playwright.config.js
JanJakes: @adamziel Should we do that here or in the Playground repo? I guess if what you mean is using Playground CLI for any tests that we write for the SQLite repo, then it probably belongs to the SQLite repo; just making sure. I can create a ticket for that.

PR #221, opened 2025-07-29
Add support for table locking statements and transaction savepoints

JanJakes:
This PR includes the following changes:

- Add support for `LOCK` and `UNLOCK` statements.
- Add support for transaction savepoints (`SAVEPOINT`, `ROLLBACK TO`, `RELEASE SAVEPOINT`)
- Improve verification of internal identifier names

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -812,7 +812,8 @@ class WP_SQLite_Driver {
 			 */
 			$this->execute_sqlite_query( $this->is_readonly ? 'BEGIN' : 'BEGIN IMMEDIATE' );
 		} else {
-			$this->execute_sqlite_query( 'SAVEPOINT LEVEL' . $this->transaction_level );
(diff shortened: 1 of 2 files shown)


wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -1017,6 +1008,121 @@ private function execute_mysql_query( WP_Parser_Node $node ): void {
...
+					break;
+				}
+
+				// COMMIT.
+				if ( WP_MySQL_Lexer::COMMIT_SYMBOL === $token->id ) {
+					$this->commit();
+					break;
+				}
+
+				// Unknown statement. Fall through to the default case.
```
zaerl: Here, the `case` falls through the following case (`savepointStatement`), not the `switch` `default` case. It's better to specify this; otherwise, it can appear as a bug.

JanJakes: @zaerl Ahh, good catch, thanks! I missed this; the linter was nagging me that a fallthrough needs a comment, but this is indeed not what I intended.

zaerl (approved):
LGTM. Great work Jan, as always.

PR #222, opened 2025-07-30
Improve Query Monitor setup to better support multisite installs

JanJakes:
This PR improves the Query Monitor setup to better support multisite installs.

The Query Monitor integration logic tries to eagerly check whether the Query Monitor plugin is active. This is non-trivial for multisites (at that stage, we don't know yet which site is active, etc.).

I created a fix that doesn’t do the eager load on multisites. All is set up correctly a bit later, when the QM plugin loads. Not loading it eagerly means we won't log a couple of queries that WP core does before it loads the QM plugin.

For single sites, the eager load is still in place.

### DIFF
--- a/integrations/query-monitor/boot.php
+++ b/integrations/query-monitor/boot.php
@@ -21,7 +21,7 @@ if ( ! defined( 'QM_DB_SYMLINK' ) ) {
 	define( 'QM_DB_SYMLINK', false );
 }
 
-// 1. Check if we should load Query Monitor (as per the original "db.php" file).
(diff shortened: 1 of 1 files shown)


adamziel:
Cc @johnbillion

JanJakes:
> Cc @johnbillion

For more context, the SQLite & Query Monitor integration is now a working prototype with a few tricks and hacks on our side. It's on my TODO list to propose some changes upstream to https://github.com/johnbillion/query-monitor/ to make the integration more robust.

JanJakes:
I'll go ahead and merge this to tag and test it with v2.2.5.

PR #231, opened 2025-08-26
Resolve ambiguous column names in ORDER BY

adamziel:
🚧  WIP  🚧 

GPT 5-assisted solution for https://github.com/WordPress/sqlite-database-integration/pull/230. The logic seems in tact, this PR needs reformatting and a pass over the comments to better explain the context

## Follow-up work

Explore whether this solution needs to also be applied to column names in LIMIT, HAVING, GROUP BY, etc.

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -227,6 +227,38 @@ class WP_SQLite_Driver {
 	);
 
 	/**
+	 * A stack of maps for resolving ORDER BY unqualified names to select-item expressions.
+	 *
+	 * Motivation:
+	 * - MySQL allows ORDER BY to reference output columns by their column names (or aliases)
+	 *   without qualification (e.g., ORDER BY name), and it resolves the reference against
+	 *   the SELECT list when unambiguous. SQLite requires the term to be either a positional
+	 *   index or a resolvable expression (e.g., t1.name) and will error on ambiguity. To provide
+	 *   MySQL-compatible behavior, we keep track of SELECT output names and their corresponding
+	 *   translated expressions so ORDER BY terms can be rewritten to safe expressions.
+	 *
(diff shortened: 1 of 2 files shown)


adamziel:
@JanJakes It seems like it works but something's off in the CI setup with the TLS certificate errors. I won't be able to address that today – would you be able to take over?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -226,6 +226,24 @@ class WP_SQLite_Driver {
...
+	 *     SELECT name from t1 JOIN t2 ORDER BY name
+  	 *
+	 * Becomes
+  	 *
+	 *     SELECT name from t1 JOIN t2 ORDER BY t1.name
+  	 *
+	 * @see https://github.com/wordpress/sqlite-database-integration/issues/228
+	 * @var array<int, array<string, string|null>> Stack of frames: name(lowercase) → expression|null
+	 */
+	private $select_output_name_to_ordinal_stack = array();
```
adamziel, line 245: I don't like this name. Maybe `$selected_columns_stack`?

JanJakes:
Oh wow, what a complexity (maintaining the stack, etc.) and [dirty code](https://github.com/WordPress/sqlite-database-integration/pull/231/files#diff-f0696e42750cef2f20fda5ea479b1ac09836a9c72d76ca0722f0a7af7879a011R3089-R3114). Nope, GPT-5, not good.

I had something much simpler in mind with a top-down approach. Gave it a quick shot in https://github.com/WordPress/sqlite-database-integration/pull/232. [The first commit](https://github.com/WordPress/sqlite-database-integration/pull/232/commits/f8725354ea74fae0ac8b2f41745609d38216a95b) is all that is needed, the second one covers an advanced edge case. I think it's almost good to go and likely much more correct. I will double-check and finalize it tomorrow.

PR #232, opened 2025-09-01
Implement disambiguation for unqualified columns in `ORDER BY`

### LINKED ISSUE #228: `Query Error: ambiguous column name` in `ORDER BY` clause
Given the following tables and data:

```sql
CREATE TABLE t1 (id INT, name TEXT);
CREATE TABLE t2 (t1_id INT, name TEXT);

INSERT INTO t1 (id, name) VALUES (1, "T1 A");
INSERT INTO t1 (id, name) VALUES (2, "T1 B");
INSERT INTO t2 (t1_id, name) VALUES (1, "T2 B");
INSERT INTO t2 (t1_id, name) VALUES (2, "T2 A");
```

The following queries error in both MySQL and SQLite:
```sql
SELECT t1.id  -- name column not specified here
FROM t1
JOIN t2 ON t2.t1_id = t1.id
ORDER BY name;
-- [MySQL]  Query Error: Column 'name' in order clause is ambiguous
-- [SQLite] Query Error: ambiguous column name: name

SELECT t1.name, t2.name  -- name column ambiguous for ORDER BY even in MySQL
FROM t1
JOIN t2 ON t2.t1_id = t1.id
ORDER BY name;
-- [MySQL]  Query Error: Column 'name' in order clause is ambiguous
-- [SQLite] Query Error: ambiguous column name: name
```

However, the following queries **error in SQLite but not in MySQL**:
```sql
SELECT t1.name  -- name column used for ORDER BY in MySQL
FROM t1
JOIN t2 ON t2.t1_id = t1.id
ORDER BY name;
-- [MySQL]  T1 A, T1 B
-- [SQLite] Query Error: ambiguous column name: name

SELECT t2.name  -- name column used for ORDER BY in MySQL
FROM t1
JOIN t2 ON t2.t1_id = t1.id
ORDER BY name;
-- [MySQL]  T2 A, T2 B
-- [SQLite] Query Error: ambiguous column name: name
```

**Fix:**
When we see a non-fully qualified column in an `ORDER BY` clause and the same column is fully qualified and non-ambiguous in the `SELECT` list, we can use the same full qualifier in the `ORDER BY` clause, and it should fix this issue.

JanJakes:
Resolves https://github.com/WordPress/sqlite-database-integration/issues/228.
Closes https://github.com/WordPress/sqlite-database-integration/pull/231.

### DIFF
--- a/tests/WP_SQLite_Driver_Tests.php
+++ b/tests/WP_SQLite_Driver_Tests.php
@@ -6466,6 +6466,116 @@ END;
 		$this->assertQuery( 'SELECT 1 FROM t1 JOIN t2 ON t2.id = t1.id ORDER BY name' );
 	}
 
+	public function testSelectGroupByAmbiguousColumnResolution(): void {
+		$this->assertQuery( 'CREATE TABLE t1 (id INT, name TEXT)' );
+		$this->assertQuery( 'CREATE TABLE t2 (id INT, name TEXT)' );
+		$this->assertQuery( 'INSERT INTO t1 (id, name) VALUES (1, "A"), (2, "A")' );
+		$this->assertQuery( 'INSERT INTO t2 (id, name) VALUES (1, "B1"), (2, "B2")' );
+
+		// The "name" column will be resolved to "t1.name" as per the SELECT item.
+		$result = $this->assertQuery( 'SELECT t1.name FROM t1 JOIN t2 ON t2.id = t1.id GROUP BY name' );
+		$this->assertEquals(
+			array( (object) array( 'name' => 'A' ) ),
+			$result
+		);
+
+		// The "name" column will be resolved to "t2.name" as per the SELECT item.
+		$result = $this->assertQuery( 'SELECT t2.name FROM t1 JOIN t2 ON t2.id = t1.id GROUP BY name' );
+		$this->assertEquals(
+			array(
+				(object) array( 'name' => 'B1' ),
+				(object) array( 'name' => 'B2' ),
+			),
+			$result
+		);
+
+		// Parenthesized column reference can be used in both SELECT and GROUP BY lists.
+		$result = $this->assertQuery( 'SELECT (t1.name) FROM t1 JOIN t2 ON t2.id = t1.id GROUP BY (((name)))' );
+		$this->assertEquals(
+			array( (object) array( '(t1.name)' => 'A' ) ),
+			$result
+		);
+
+		/*
+		 * The following query fails with "ambiguous column name" in MySQL, but
+		 * in SQLite, it works. It's OK to keep this difference as MySQL behaves
+		 * rather strangely in this case:
+		 *
+		 *   1) This is OK in MySQL:
+		 *        SELECT t1.name AS col, 123 AS col ... GROUP BY col
+		 *   2) This fails in MySQL:
+		 *        SELECT t1.name AS col, t2.name AS col ... GROUP BY col
+		 */
+		$this->assertQuery( 'SELECT t1.name AS col, t2.name AS col FROM t1 JOIN t2 ON t2.id = t1.id GROUP BY col' );
+	}
+
+	public function testSelectGroupByAmbiguousColumnError(): void {
+		$this->assertQuery( 'CREATE TABLE t1 (id INT, name TEXT)' );
+		$this->assertQuery( 'CREATE TABLE t2 (id INT, name TEXT)' );
+
+		$this->expectException( 'WP_SQLite_Driver_Exception' );
+		$this->expectExceptionMessage( 'ambiguous column name: name' );
+		$this->assertQuery( 'SELECT t1.name, t2.name FROM t1 JOIN t2 ON t2.id = t1.id GROUP BY name' );
+	}
+
+	public function testSelectGroupByAmbiguousColumnErrorWithoutSelectList(): void {
+		$this->assertQuery( 'CREATE TABLE t1 (id INT, name TEXT)' );
+		$this->assertQuery( 'CREATE TABLE t2 (id INT, name TEXT)' );
+
+		$this->expectException( 'WP_SQLite_Driver_Exception' );
+		$this->expectExceptionMessage( 'ambiguous column name: name' );
+		$this->assertQuery( 'SELECT 1 FROM t1 JOIN t2 ON t2.id = t1.id GROUP BY name' );
+	}
+
+	public function testSelectHavingAmbiguousColumnResolution(): void {
+		$this->assertQuery( 'CREATE TABLE t1 (id INT, name TEXT)' );
+		$this->assertQuery( 'CREATE TABLE t2 (id INT, name TEXT)' );
+		$this->assertQuery( 'INSERT INTO t1 (id, name) VALUES (1, "A"), (2, "A")' );
+		$this->assertQuery( 'INSERT INTO t2 (id, name) VALUES (1, "B1"), (2, "B2")' );
+
+		// The "name" column will be resolved to "t1.name" as per the SELECT item.
+		$result = $this->assertQuery( 'SELECT t1.name FROM t1 JOIN t2 ON t2.id = t1.id HAVING name' );
+		$this->assertEquals( array(), $result );
+
+		// The "name" column will be resolved to "t2.name" as per the SELECT item.
+		$result = $this->assertQuery( 'SELECT t2.name FROM t1 JOIN t2 ON t2.id = t1.id HAVING name' );
+		$this->assertEquals( array(), $result );
+
+		// Parenthesized column reference can be used in both SELECT and GROUP BY lists.
+		$result = $this->assertQuery( 'SELECT (t1.name) FROM t1 JOIN t2 ON t2.id = t1.id HAVING (((name)))' );
+		$this->assertEquals( array(), $result );
+
+		/*
+		 * The following query fails with "ambiguous column name" in MySQL, but
+		 * in SQLite, it works. It's OK to keep this difference as MySQL behaves
+		 * rather strangely in this case:
+		 *
+		 *   1) This is OK in MySQL:
+		 *        SELECT t1.name AS col, 123 AS col ... HAVING col
+		 *   2) This fails in MySQL:
+		 *        SELECT t1.name AS col, t2.name AS col ... HAVING col
+		 */
+		$this->assertQuery( 'SELECT t1.name AS col, t2.name AS col FROM t1 JOIN t2 ON t2.id = t1.id HAVING col' );
+	}
+
+	public function testSelectHavingAmbiguousColumnError(): void {
+		$this->assertQuery( 'CREATE TABLE t1 (id INT, name TEXT)' );
+		$this->assertQuery( 'CREATE TABLE t2 (id INT, name TEXT)' );
+
+		$this->expectException( 'WP_SQLite_Driver_Exception' );
+		$this->expectExceptionMessage( 'ambiguous column name: name' );
+		$this->assertQuery( 'SELECT t1.name, t2.name FROM t1 JOIN t2 ON t2.id = t1.id HAVING name' );
+	}
+
+	public function testSelectHavingAmbiguousColumnErrorWithoutSelectList(): void {
+		$this->assertQuery( 'CREATE TABLE t1 (id INT, name TEXT)' );
+		$this->assertQuery( 'CREATE TABLE t2 (id INT, name TEXT)' );
+
+		$this->expectException( 'WP_SQLite_Driver_Exception' );
+		$this->expectExceptionMessage( 'ambiguous column name: name' );
+		$this->assertQuery( 'SELECT 1 FROM t1 JOIN t2 ON t2.id = t1.id HAVING name' );
+	}
+
 	public function testRollbackNonExistentTransactionSavepoint(): void {
 		$this->expectException( 'WP_SQLite_Driver_Exception' );
 		$this->expectExceptionMessage( 'no such savepoint: sp1' );
--- a/tests/WP_SQLite_Driver_Tests.php
+++ b/tests/WP_SQLite_Driver_Tests.php
@@ -6323,6 +6323,149 @@ END;
 		$this->assertSame( array(), (array) array_column( $result, 'id' ) );
 	}
 
+	public function testSelectOrderByAmbiguousColumnResolution(): void {
+		$this->assertQuery( 'CREATE TABLE t1 (id INT, name TEXT)' );
+		$this->assertQuery( 'CREATE TABLE t2 (id INT, name TEXT)' );
+		$this->assertQuery( 'INSERT INTO t1 (id, name) VALUES (1, "A1"), (2, "A2")' );
+		$this->assertQuery( 'INSERT INTO t2 (id, name) VALUES (1, "B2"), (2, "B1")' );
+
+		// The "name" column will be resolved to "t1.name" as per the SELECT item.
+		$result = $this->assertQuery( 'SELECT t1.name FROM t1 JOIN t2 ON t2.id = t1.id ORDER BY name DESC' );
+		$this->assertEquals(
+			array(
+				(object) array( 'name' => 'A2' ),
+				(object) array( 'name' => 'A1' ),
+			),
+			$result
+		);
+
+		// The "name" column will be resolved to "t2.name" as per the SELECT item.
+		$result = $this->assertQuery( 'SELECT t2.name FROM t1 JOIN t2 ON t2.id = t1.id ORDER BY name DESC' );
+		$this->assertEquals(
+			array(
+				(object) array( 'name' => 'B2' ),
+				(object) array( 'name' => 'B1' ),
+			),
+			$result
+		);
+
+		// The "name" column will be resolved to "t1.name", the "id" column will be resolved to "t2.id".
+		$this->assertQuery( 'INSERT INTO t1 (id, name) VALUES (3, "A2")' );
+		$this->assertQuery( 'INSERT INTO t2 (id, name) VALUES (3, "A2")' );
+		$result = $this->assertQuery( 'SELECT t2.id, t1.name FROM t1 JOIN t2 ON t2.id = t1.id ORDER BY name, id DESC' );
+		$this->assertEquals(
+			array(
+				(object) array(
+					'id'   => '1',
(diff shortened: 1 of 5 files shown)


adamziel:
That's much simpler than the alternative, thank you @JanJakes

adamziel:
@JanJakes is this one ready for a review?

JanJakes:
@adamziel It should be ready now.

I was also suspecting that we may need to reuse this for `UPDATE` and `DELETE` statements, but it looks like we're good:

> You can also perform [UPDATE](https://dev.mysql.com/doc/refman/8.0/en/update.html) operations covering multiple tables. However, you cannot use ORDER BY or LIMIT with a multiple-table [UPDATE](https://dev.mysql.com/doc/refman/8.0/en/update.html).

From https://dev.mysql.com/doc/refman/8.0/en/update.html.

> You cannot use ORDER BY or LIMIT in a multiple-table DELETE.

From https://dev.mysql.com/doc/refman/9.1/en/delete.html.

tests/WP_SQLite_Driver_Tests.php
```diff
@@ -6323,6 +6323,149 @@ public function testTransactionSavepoints(): void {
...
+		 * in SQLite, it works. It's OK to keep this difference as MySQL behaves
+		 * rather strangely in this case:
+		 *
+		 *   1) This is OK in MySQL:
+		 *        SELECT t1.name AS col, 123 AS col ... ORDER BY col
+		 *   2) This fails in MySQL:
+		 *        SELECT t1.name AS col, t2.name AS col ... ORDER BY col
+		 */
+		$this->assertQuery( 'SELECT t1.name AS col, t2.name AS col FROM t1 JOIN t2 ON t2.id = t1.id ORDER BY col' );
+	}
```
adamziel, line 6448: Let's also test syntactically tricky cases such as:

```sql
WITH temp(name) AS (
  VALUES
    ('a'),
    ('b')
)
SELECT t1.name, (SELECT name from temp WHERE id = 5) FROM t1 ORDER BY name;

WITH temp(name, b) AS (SELECT * FROM t2)
SELECT t1.name, (SELECT name from temp WHERE id = 5) FROM t1 ORDER BY name;

SELECT * FROM (
	SELECT t1.name, t2.name FROM t1 JOIN t2 ON t2.id = t1.id ORDER BY name
) ORDER BY col
```

It's fine if they fail, we need to set the line somewhere. Let's just make sure we account for them in the test suite.

tests/WP_SQLite_Driver_Tests.php
```diff
@@ -6323,6 +6323,149 @@ public function testTransactionSavepoints(): void {
...
+			array(
+				(object) array( 'name' => 'B1-two' ),
+				(object) array( 'name' => 'A2-two' ),
+				(object) array( 'name' => 'B2-two' ),
+			),
+			$result
+		);
+
+		/*
+		 * The following query fails with "ambiguous column name" in MySQL, but
```
adamziel, line 6438: So sometimes it's MySQL that fails with "ambiguous column name" and sometimes it's SQLite? I've noticed there's a different comment that talks about SQLite failing.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2905,6 +2907,204 @@ private function translate_qualified_identifier(
...
+		 * can be disambiguated in the SELECT item list.
+		 *
+		 * For example, with tables "t1" and "t2" both having a "name" column,
+		 * the following query will cause an "ambiguous column name" error in
+		 * SQLite, but not in MySQL:
+		 *
+		 *   SELECT t1.name FROM t1 JOIN t2 ON t2.t1_id = t1.id ORDER BY name
+		 *
+		 * This is because MySQL first considers the "name" column that was used
+		 * in the SELECT list. If it is unambiguous, it will be used in ORDER BY.
```
adamziel: Should we also do this for all SELECT clauses such as JOIN, HAVING, GROUP BY, etc?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2905,6 +2907,204 @@ private function translate_qualified_identifier(
...
+				// Skip when the column reference is not qualified (no "dotIdentifier" node).
+				$dot_identifiers = $select_column_ref->get_descendant_nodes( 'dotIdentifier' );
+				if ( 0 === count( $dot_identifiers ) ) {
+					continue;
+				}
+
+				// Support also parenthesized column references (e.g. "(t.id)").
+				$select_item_expr = $this->unnest_parenthesized_expression( $select_item_expr );
+
+				// Consider only simple and parenthesized column references.
```
adamziel: It should be fine for now. Let's just leave a clear TODO to revisit this once a use-case reveals itself. It may take some months or years and a clear callout will be a huge gift once it happens.

adamziel (approved):
I like the simplicity of it and that it only takes computing cycles when the query needs disambiguation and not pre-emptively like the other PR did. I left a few, rather small notes. This looks great overall – thank you!

wp-setup.sh
```diff
@@ -37,14 +37,16 @@ services:
       - ../:/var/www/src/wp-content/plugins/sqlite-database-integration
 
   php:
-    image: wordpressdevelop/php:8.3-fpm
+    # PHP temporarily pinned to 8.3.10, see: https://github.com/WordPress/wordpress-develop/pull/9602
```
adamziel, line 40: Aha, good catch!

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2905,6 +2907,204 @@ private function translate_qualified_identifier(
...
+				}
+			}
+			return implode( ' ', $parts );
+		}
+
+		return $this->translate_sequence( $node->get_children() );
+	}
+
+	/**
+	 * Unnest parenthesized MySQL expression node.
```
adamziel: I'm really glad this came up. Nice!

tests/WP_SQLite_Driver_Tests.php
JanJakes: Yes, both MySQL and SQLite throw that error, and they slightly differ in edge cases:

1. SQLite doesn't use the SELECT list to try to disambiguate. That's what we're emulating.
2. Both, interestingly, allow multiple identical explicit aliases (`1 AS col`, `2 AS col`, `3 AS col`) — but here they may slightly differ in what they would sort by with `ORDER BY col`.
3. MySQL doesn't allow a special case when multiple columns share the same explicit alias. Even if these are entirely different columns (`id` and `name`, for example). SQLite is fine with that and handles it the same as in point 2.

We're not emulating anything for 2 and 3 — just passing it to SQLite "as is," because it can handle it, just with these edge case variations.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: Oh, what I mean is here "Consider only simple and parenthesized column references, **because that's what MySQL does**." As soon as you do any other expression, e.g. `SELECT col + 1 ... ORDER BY col`, it won't recognize the column and disambiguate. It will only recognize it as `SELECT id`, `SELECT (id)`, etc. The same for the `ORDER BY` clause.

I suspect MySQL may be just normalizing these extra no-op parentheses at some point and that's why it works.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel Oh, good catch! I tried it in `JOIN` and `WHERE` and MySQL didn't disambiguate there. But in `HAVING` and `GROUP BY` it seems to work. Let me check if it's easy to reuse the disambiguation map for these cases.

JanJakes: Got it working. It's nicely reusable for both of these using a private method. I'll polish the code a bit and will push the changes.

adamziel: `WINDOW` might be another (unlikely) one, but I'm not sure

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
adamziel: Ooooh, gotcha! Let's document that, then

tests/WP_SQLite_Driver_Tests.php
JanJakes: @adamziel The first one doesn't seem to be a valid MySQL syntax (something about the `VALUES` part). The last one, we  already have very similar. I'll add a test using the `WITH` syntax.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: Yeah, this one also disambiguates in MySQL:

```sql
SELECT
  t1.name,
  SUM(t1.name) OVER (PARTITION BY name) FROM t1 JOIN t2 ON t2.id = t1.id;
```

There are quite a few nuances with `ORDER BY` and `HAVING` already, so for now, I'd leave this out, I guess.

tests/WP_SQLite_Driver_Tests.php
adamziel: Oh, sorry, the first one is SQLite

aristath (approved):

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
adamziel: Yup, makes sense. We don't have to solve everything at once. The queries we don't fix will fail so we'll get clear feedback if and when that comes up

JanJakes:
@adamziel I pushed the `GROUP BY` and `HAVING` support and some more complex cases like CTEs, UNIONs, etc.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2905,6 +2893,139 @@ private function translate_qualified_identifier(
...
 
+	private function translate_query_specification( WP_Parser_Node $node ): string {
+		$group_by = $node->get_first_child_node( 'groupByClause' );
+		$having   = $node->get_first_child_node( 'havingClause' );
+
+		$group_by_clause = null;
+		$having_clause   = null;
+		if ( $group_by || $having ) {
+			$select_item_list   = $node->get_first_child_node( 'selectItemList' );
+			$disambiguation_map = $this->create_select_item_disambiguation_map( $select_item_list );
```
adamziel: Let's add an inline comment here to say something like "Disambiguate column references. See the `disambiguate_item` method below for more context."

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2905,6 +2893,139 @@ private function translate_qualified_identifier(
...
+				}
+				$having_clause = 'HAVING ' . implode( ' ', $disambiguated_having_list );
+			}
+
+			$parts = array();
+			foreach ( $node->get_children() as $child ) {
+				if ( $child instanceof WP_Parser_Node && 'groupByClause' === $child->rule_name ) {
+					$parts[] = $group_by_clause;
+				} elseif ( $child instanceof WP_Parser_Node && 'havingClause' === $child->rule_name ) {
+					// Translate "HAVING ..." without "GROUP BY ..." to "GROUP BY 1 HAVING ...".
```
adamziel: Let's explain why we're doing that

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
adamziel: A local variable is much simpler than the stack-based approach proposed by GPT, nice!

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2905,6 +2893,139 @@ private function translate_qualified_identifier(
...
+					null !== $order_direction ? ( ' ' . $this->translate( $order_direction ) ) : ''
+				);
+			}
+
+			// Translate the query expression, replacing the ORDER BY list with
+			// the one that was constructed using the disambiguation algorithm.
+			$parts = array();
+			foreach ( $node->get_children() as $child ) {
+				if ( $child instanceof WP_Parser_Node && 'orderClause' === $child->rule_name ) {
+					$parts[] = 'ORDER BY ' . implode( ', ', $disambiguated_order_list );
```
adamziel, line 2960: Just to make sure, complex expressions that don't need disambiguation continue to work here, right? As in `ORDER BY ABS((SELECT 4))` etc.? I assume yes since all the tests pass, but I still want to ask.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -3633,6 +3754,175 @@ private function translate_update_list_in_non_strict_mode( string $table_name, W
...
+		if (
+			1 === count( $children )
+			&& $children[0] instanceof WP_Parser_Node
+			&& in_array( $children[0]->rule_name, array( 'expr', 'boolPri', 'predicate', 'bitExpr', 'simpleExpr' ), true )
+		) {
+			$unnested = $this->unnest_parenthesized_expression( $children[0] );
+			return $unnested === $children[0] ? $node : $unnested;
+		}
+
+		// Unnest "OPEN_PAR_SYMBOL exprList CLOSE_PAR_SYMBOL" to "exprList".
```
adamziel, line 3811: Do we unnest things like `SELECT ABS((SELECT MIN(balance) FROM accounts)) AS smallest_debt;`? Or leave them alone?

tests/WP_SQLite_Driver_Translation_Tests.php
```diff
@@ -1328,6 +1328,348 @@ public function testIndexHints(): void {
...
+			'SELECT t1.name FROM t1 EXCEPT SELECT t2.name FROM t2 ORDER BY name'
+		);
+
+		// With INTERSECT, the column should not be disambiguated.
+		$this->assertQuery(
+			'SELECT `t1`.`name` FROM `t1` INTERSECT SELECT `t2`.`name` FROM `t2` ORDER BY `name`',
+			'SELECT t1.name FROM t1 INTERSECT SELECT t2.name FROM t2 ORDER BY name'
+		);
+
+		// Test a complex query with CTEs.
```
adamziel, line 1443: nice, thank you

tests/WP_SQLite_Driver_Translation_Tests.php
```diff
@@ -1328,6 +1328,348 @@ public function testIndexHints(): void {
...
+		// With EXCEPT, the column should not be disambiguated.
+		$this->assertQuery(
+			'SELECT `t1`.`name` FROM `t1` EXCEPT SELECT `t2`.`name` FROM `t2` ORDER BY `name`',
+			'SELECT t1.name FROM t1 EXCEPT SELECT t2.name FROM t2 ORDER BY name'
+		);
+
+		// With INTERSECT, the column should not be disambiguated.
+		$this->assertQuery(
+			'SELECT `t1`.`name` FROM `t1` INTERSECT SELECT `t2`.`name` FROM `t2` ORDER BY `name`',
+			'SELECT t1.name FROM t1 INTERSECT SELECT t2.name FROM t2 ORDER BY name'
```
adamziel, line 1440: TIL about INTERSECT

adamziel (approved):
I only left minor questions and nitpicks. This PR seems good to go and it's a quality work that covers a lot of bases. It's weird how such a nuance required so much research! Thank you for all the great work on this @JanJakes

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel The unnest logic now only descends some basic expression nodes (`expr`, `boolPri`, `predicate`, 'bitExpr', `simpleExpr`), so in this case, it wouldn't do anything, as it contains some function call node on some level before reaching the `OPEN_PAR_SYMBOL exprList CLOSE_PAR_SYMBOL`.

I'm suspecting that maybe the MySQL parser removes all extra parentheses in some preprocessing, but who knows. We could try to find it in their codebase and then improve accordingly. For example, when you do `SELECT (((name)))` in MySQL, PDO will report that the select column name is just `name`. We don't do this part yet; I will create a ticket.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: Yeah, it's the `??` in https://github.com/WordPress/sqlite-database-integration/pull/232/files/9d2c53660a9b5407828dd4f875ff98c01a113d83#diff-f0696e42750cef2f20fda5ea479b1ac09836a9c72d76ca0722f0a7af7879a011R3008 that handles that. (I also tried that query just to be sure.)

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: If I were to pass here directly the function call arguments node, then something like that could happen, but in the current usage, it should not be possible—a function call will always be under an expression, and the arguments list will always be under a function call node, etc.

JanJakes:
@adamziel Added some more docs, and it should be ready now. Thanks!

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel The `HAVING` behavior made me think about how we could pass translation logic customizations down the recursion (as GPT did) for cases when we want to disambiguate more deeply nested nodes (e.g., column references in `HAVING`). I have an idea for an API that could maybe solve it a bit more universally, without per-case stacks, etc. Something like this:

```php
$having_clause = $this->translate_with_customization(
	$having_expr,
	function ( $node ) use ( $disambiguation_map ) {
		if ( $node instanceof WP_Parser_Node && 'columnRef' === $node->rule_name ) {
			return $this->disambiguate_item( $disambiguation_map, $node );
		}
		return $this->translate( $node );
	}
);
```

This way, we could pass the `$disambiguation_map` deep down the tree while keeping the modification collocated with the code that handles it at the higher level.

PR #238, opened 2025-09-11
Implement `UPDATE` with mutliple tables used (single modified table only)

### LINKED ISSUE #233: Support for `UPDATE ... JOIN` and other multi-table `UPDATE` syntaxes
Action Scheduler is using an `UPDATE ... JOIN` construct which appears to be breaking WooCommerce install on any SQLite backed WP instance.

We already support `DELETE ... JOIN`, let's also support it for `UPDATE`.

JanJakes:
This PR implements support for `UPDATE` statements with multiple tables used.

It only allows a single modified table, as updating multiple tables at once is likely much more complex and may require using temporary tables (e.g., in `UPDATE t1, t2 SET t1.id = t2.id, t2.id = t1.id`, we can't simply update `t1` first and `t2` after that).

This is what I managed to do today—it may require adding some more tests.

Resolves https://github.com/WordPress/sqlite-database-integration/issues/233.

### DIFF
--- a/tests/WP_SQLite_Driver_Tests.php
+++ b/tests/WP_SQLite_Driver_Tests.php
@@ -6693,4 +6693,71 @@ END;
 		$result = $this->assertQuery( 'SELECT * FROM t2' );
 		$this->assertEquals( array( (object) array( 'id' => '0' ) ), $result );
 	}
+
+	public function testUpdateWithJoinedTables(): void {
+		$this->assertQuery( 'CREATE TABLE t1 (id INT)' );
+		$this->assertQuery( 'CREATE TABLE t2 (id INT, name TEXT)' );
+		$this->assertQuery( 'CREATE TABLE t3 (id INT, name TEXT)' );
+
+		$this->assertQuery( 'INSERT INTO t1 (id) VALUES (1), (2), (3)' );
+		$this->assertQuery( 'INSERT INTO t2 (id, name) VALUES (1, "update")' );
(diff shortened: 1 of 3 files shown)


adamziel:
TIL about UPDATE FROM - really cool!

> we can't simply update t1 first and t2 after that

Why not? Couldn't we wrap that in a transaction and make it an atomic operation that's either committed or rolled back as a whole? I mean that for operations that don't involve a swapping cycle of two or more columns. Perhaps it's not worth it as we'd need to compute a graph of dependencies and find cycles in it and that alone is complex 🤔

adamziel:
I wonder if starting n transactions would help us with anything here. MySQL has REPEATABLE READ, if SQLite has something similar then we could set t1.id = t2.id in one transaction, and still have access to original t1.id in another transaction

adamziel:
The more I think about it, the more temporary tables seem like the right approach. Any "target" table would be the regular table, and any "data source" would be a temporary copy of the original table

JanJakes:
@adamziel Unfortunately, it seems that SQLite has [No Isolation Between Operations On The Same Database Connection](https://www.sqlite.org/isolation.html).

If we'd go with temporary tables, I think it's still worth it to keep this simple and more performant path and use the temp tables only for the special case of updating multiple tables at once. It's probably quite an edge case anyway.

adamziel (approved):
I haven't reviewed in depth, but it looks good from a brief read and the tests are passing so it shouldn't break anything.

JanJakes:
@adamziel Thanks! I'll tag this as well, and then we'll work on adding some more test cases for this in a separate PR.

PR #245, opened 2025-09-17
Add support for all binary and hexadecimal literal syntaxes

JanJakes:
Add support for all binary and hexadecimal literal syntaxes. The differences are:
1. In MySQL, `0x…` is a binary literal, while in SQLite it’s a numeric one.
2. In MySQL, `0b...`, `b'...'`, and `B'...'` syntaxes can be used to represent binary literals. SQLite only supports HEX literals (`x'...'`).

This PR implements the following conversions:
1. Convert `0x...` to `x'...'`.
2. Convert `0b...`, `b'...'`, and `B'...'` to `x'...'` with binary to hexadecimal conversion.

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -2860,6 +2860,31 @@ class WP_SQLite_Driver {
 		switch ( $token->id ) {
 			case WP_MySQL_Lexer::EOF:
 				return null;
+			case WP_MySQL_Lexer::BIN_NUMBER:
+				/*
+				 * There are no binary literals in SQLite. We need to convert all
+				 * MySQL binary string values to HEX strings in SQLite (x'...').
+				 */
+				$value = $token->get_value();
+				if ( '0' === $value[0] ) {
+					// 0b...
+					$value = substr( $value, 2 );
+				} else {
+					// b'...' or B'...'
+					$value = substr( $value, 2, -1 );
+				}
+				return sprintf( "x'%s'", base_convert( $value, 2, 16 ) );
+			case WP_MySQL_Lexer::HEX_NUMBER:
+				/*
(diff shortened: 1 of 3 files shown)


wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2860,6 +2860,31 @@ private function translate_token( WP_MySQL_Token $token ): ?string {
...
+				/*
+				 * There are no binary literals in SQLite. We need to convert all
+				 * MySQL binary string values to HEX strings in SQLite (x'...').
+				 */
+				$value = $token->get_value();
+				if ( '0' === $value[0] ) {
+					// 0b...
+					$value = substr( $value, 2 );
+				} else {
+					// b'...' or B'...'
```
adamziel, line 2873: Should we also account for `X'01'`, `x'01'`, `_binary X'01'`, and `_binary x'01'`? Or are these fine because we can just preserve them as they are and we won't see them in this `else`?

adamziel (approved):

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel `X'01'` and `x'01'` will be HEX literals and processed in the case below (`HEX_NUMBER`). So they should never appear under the `WP_MySQL_Lexer::BIN_NUMBER` token. These are valid in SQLite, so that case leaves them as is (only `0x...` is translated).

For `_binary x'01'`, I think the cast syntax will fail in SQLite at the moment, but behavior-wise, it should be fine, as both `x'41'` and `_binary x'41'` represent the same binary bytes — with `SELECT`, you get `A` as a result in both cases.

f1-outsourcing:
@JanJakes is there some install manual?? Nothing seems to work.

https://github.com/WordPress/sqlite-database-integration/issues/247

PR #255, opened 2025-10-01
Pad binary string literals to full bytes as required by SQLite HEX notation

JanJakes:
This fixes the following error:
```bash
SQLSTATE[HY000]: General error: 1 unrecognized token: "x'1'"
```

For queries like the following:
```sql
SELECT 0b1;
SELECT 0b01;
SELECT 0b001;
...
```

SQLite doesn't have binary string literals, so [we need to convert them to HEX string literals](https://github.com/WordPress/sqlite-database-integration/pull/245). However, HEX string literals in SQLite must be aligned to full bytes (their length needs to be even) — `x'01'` is valid, while `x'1'` is invalid.

In MySQL, binary strings don't require such alignment, and `0b1` is valid (and `SELECT 0b1 = 0b01` returns `true`).

Additionally, the `base_convert()` function used in the [original implementation](https://github.com/WordPress/sqlite-database-integration/pull/245) doesn't preserve existing leading `0` padding.

To solve both of these issues, this PR implements the following:
1. Count how many full bytes the original binary string requires.
2. Make sure the resulting HEX string has that number of bytes, using leading `0` padding.

For HEX strings, this fix is not needed, as those require to be byte-aligned also in MySQL, so throwing an error for `x'1'` is actually correct.

### DIFF
--- a/tests/WP_SQLite_Driver_Tests.php
+++ b/tests/WP_SQLite_Driver_Tests.php
@@ -6883,6 +6883,22 @@ END;
 
 		$result = $this->assertQuery( "SELECT B'0100000101111010'" );
 		$this->assertEquals( array( (object) array( "B'0100000101111010'" => 'Az' ) ), $result );
+
+		// Verify correct padding (0b1 === 0b01 === 0b001 ... === 0x00000001).
+		$result = $this->assertQuery( 'SELECT 0b1' );
+		$this->assertEquals( array( (object) array( '0b1' => pack( 'H*', '01' ) ) ), $result );
+
+		$result = $this->assertQuery( 'SELECT 0b01' );
+		$this->assertEquals( array( (object) array( '0b01' => pack( 'H*', '01' ) ) ), $result );
+
+		$result = $this->assertQuery( 'SELECT 0b001' );
+		$this->assertEquals( array( (object) array( '0b001' => pack( 'H*', '01' ) ) ), $result );
+
+		$result = $this->assertQuery( 'SELECT 0b00000001' );
+		$this->assertEquals( array( (object) array( '0b00000001' => pack( 'H*', '01' ) ) ), $result );
+
+		$result = $this->assertQuery( 'SELECT 0b000000001' );
+		$this->assertEquals( array( (object) array( '0b000000001' => pack( 'H*', '0001' ) ) ), $result );
 	}
 
 	public function testHexadecimalLiterals(): void {
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -2873,7 +2873,18 @@ class WP_SQLite_Driver {
 					// b'...' or B'...'
 					$value = substr( $value, 2, -1 );
 				}
-				return sprintf( "x'%s'", base_convert( $value, 2, 16 ) );
+
+				// Convert the binary string to HEX.
+				$hex = base_convert( $value, 2, 16 );
+
+				/*
+				 * The "base_convert()" function doesn't add or preserve padding.
+				 * Let's compute how many bytes we expect and pad the HEX value
+				 * to full bytes (SQLite requires HEX strings of even length).
+				 */
+				$byte_count = (int) ceil( strlen( $value ) / 8 );
+				$hex        = str_pad( $hex, $byte_count * 2, '0', STR_PAD_LEFT );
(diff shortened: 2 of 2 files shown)


wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -2873,7 +2873,18 @@ private function translate_token( WP_MySQL_Token $token ): ?string {
...
+				// Convert the binary string to HEX.
+				$hex = base_convert( $value, 2, 16 );
+
+				/*
+				 * The "base_convert()" function doesn't add or preserve padding.
+				 * Let's compute how many bytes we expect and pad the HEX value
+				 * to full bytes (SQLite requires HEX strings of even length).
+				 */
+				$byte_count = (int) ceil( strlen( $value ) / 8 );
+				$hex        = str_pad( $hex, $byte_count * 2, '0', STR_PAD_LEFT );
```
akirk, line 2886: Is this really necessary or is this a requirement of the notation that starts with x? Quick test run:
```
sqlite> select x'1';
Parse error: unrecognized token: "x'1'"
  select x'1';
         ^--- error here
sqlite> select x'01';

sqlite> select x'02';

sqlite> select 0x1;
1
sqlite> select 0x01;
1
```

akirk (commented):
I think the PR title is confusing because you first talk about hex literals and the change is about binary representation.

tests/WP_SQLite_Driver_Tests.php
```diff
@@ -6883,6 +6883,22 @@ public function testBinaryLiterals(): void {
 
 		$result = $this->assertQuery( "SELECT B'0100000101111010'" );
 		$this->assertEquals( array( (object) array( "B'0100000101111010'" => 'Az' ) ), $result );
+
+		// Verify correct padding (0b1 === 0b01 === 0b001 ... === 0x00000001).
+		$result = $this->assertQuery( 'SELECT 0b1' );
+		$this->assertEquals( array( (object) array( '0b1' => pack( 'H*', '01' ) ) ), $result );
```
adamziel, line 6889: Can we spell out the tested string explicitly without relying on `pack`? Or have two tests?

adamziel, line 6889: Can we spell out the tested string explicitly without relying on `pack`? Or have two tests?

adamziel (approved):
looking great!

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @akirk Unfortunately, in SQLite the `0x1` notation is a numeric literal, while the `x'01'` notation is binary (`SELECT 0x01 = x'01';` is `false`). From the [original PR](https://github.com/WordPress/sqlite-database-integration/pull/245):

1. In MySQL, `0x…` is a binary literal, while in SQLite it’s a numeric one.
2. In MySQL, `0b...`, `b'...'`, and `B'...'` syntaxes can be used to represent binary literals. SQLite only supports HEX literals (`x'...'`).

tests/WP_SQLite_Driver_Tests.php
JanJakes: @adamziel It's hard for things like `0b1`, because that byte represents an invisible character. Converting strings like `00000001` to the actual byte value in PHP seems to be not so straightforward. I used `pack`, maybe there's a better way?

JanJakes:
> I think the PR title is confusing because you first talk about hex literals and the change is about binary representation.

@akirk Thanks! I improved it, hopefully. It's all a bit confusing, as one HEX notation is a binary string in MySQL and a number in SQLite (`0x...`), but another notation is a binary string in both (`x'...'`).

tests/WP_SQLite_Driver_Tests.php
JanJakes: Another way to express it would be char by char:
```php
pack( 'H*', '01' ) === chr(1); // true
pack( 'H*', '0001' ) === chr(0) . chr(1) // true
...
```
Is that more readable? I don't know.

Maybe Unicode codepoints are better?
```php
pack( 'H*', '01' ) === "\u{1}"; // true
pack( 'H*', '0001' ) === "\u{0}\u{1}"; // true
```

adamziel: Gotcha! I suppose pack works

PR #256, opened 2025-10-01
Introduce an unsafe flag for using the new driver with SQLite < `3.37.0`

### LINKED ISSUE #252: SQLite plugin not working on SQLite 3.34.1
## What happened?

When importing a site with this plugin and an existing database on my server with SQLite 3.34.1 I got a fatal error:

```
[18-Sep-2025 17:46:32 UTC] PHP Fatal error:  Uncaught PDOException: SQLSTATE[HY000]: General error: 11 malformed database schema (_wp_sqlite_mysql_information_schema_schemata) - near "STRICT": syntax error in /srv/htdocs/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php:401
Stack trace:
#0 /srv/htdocs/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(401): PDO->query('CREATE TABLE IF...')
#1 /srv/htdocs/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php(348): WP_SQLite_Translator->__construct(Object(PDO))
#2 /srv/htdocs/__wp__/wp-includes/class-wpdb.php(773): WP_SQLite_DB->db_connect()
#3 /srv/htdocs/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php(55): wpdb->__construct('', Object(SensitiveParameterValue), '151359310', '')
#4 /srv/htdocs/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite/db.php(84): WP_SQLite_DB->__construct('151359310')
#5 /srv/htdocs/wp-content/db.php(37): require_once('/srv/htdocs/wp-...')
#6 /srv/htdocs/__wp__/wp-includes/load.php(708): require_once('/srv/htdocs/wp-...')
#7 /srv/htdocs/__wp__/wp-settings.php(134): require_wp_db()
#8 /srv/htdocs/wp-config.php(105): require_once('/srv/htdocs/__w...')
#9 /srv/htdocs/__wp__/wp-load.php(55): require_once('/srv/htdocs/wp-...')
#10 /srv/htdocs/__wp__/wp-blog-header.php(13): require_once('/srv/htdocs/__w...')
#11 /srv/htdocs/__wp__/index.php(17): require('/srv/htdocs/__w...')
#12 {main}
  thrown in /srv/htdocs/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php on line 401
```

Image

## What would I expect?

It would be great if the plugin is compatible with this SQLite version, or it's able to load the SQLite binary version needed, or if at least it fails gracefully explaining why it cannot use these versions.

JanJakes:
Adds a `WP_SQLITE_UNSAFE_ENABLE_UNSUPPORTED_VERSIONS` flag for use with older versions of SQLite.

I came to the conclusion that a special flag with the word `UNSAFE` might be the best solution here.

By default, the new driver would still throw the following error:

> The SQLite version 3.27.2 is not supported. Minimum required version is 3.37.0.

But with the new flag, it could enable accessing a database created with 3.37.0 on older SQLite versions.

**An important consideration or TODO:**
In this form, it would serve a basic use case, which is opening an existing database on an older SQLite for previews, including writing data to the database.

However, any new `CREATE TABLE` statement would fail because of the use of the `STRICT` keyword (`General error: 1 near "STRICT": syntax error`). We could address that by omitting that keyword in this mode, but maybe let's first determine whether there are use cases for it.

Resolves https://github.com/WordPress/sqlite-database-integration/issues/252.

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -483,13 +483,35 @@ class WP_SQLite_Driver {
 		// Check the SQLite version.
 		$sqlite_version = $this->get_sqlite_version();
 		if ( version_compare( $sqlite_version, self::MINIMUM_SQLITE_VERSION, '<' ) ) {
-			throw $this->new_driver_exception(
-				sprintf(
-					'The SQLite version %s is not supported. Minimum required version is %s.',
-					$sqlite_version,
-					self::MINIMUM_SQLITE_VERSION
-				)
-			);
+			if ( defined( 'WP_SQLITE_UNSAFE_ENABLE_UNSUPPORTED_VERSIONS' ) && WP_SQLITE_UNSAFE_ENABLE_UNSUPPORTED_VERSIONS ) {
+				/*
+				 * SQLite versions prior to 3.37.0 do not support STRICT tables.
+				 *
+				 * However, a database created with SQLite >= 3.37.0 can be used
+				 * with SQLite versions < 3.37.0 when "PRAGMA writable_schema" is
+				 * set to "ON", which also enables error-tolerant schema parsing.
+				 *
+				 * This is an unsafe opt-in feature for special back compatibility
+				 * use cases, as it can corrupt the database by allowing incorrect
+				 * types into STRICT tables. Additionally, depending on the legacy
+				 * SQLite version used, there is no guarantee that all features of
+				 * the SQLite driver will work as expected. Use this with caution.
+				 *
+				 * See: https://www.sqlite.org/stricttables.html#accessing_strict_tables_in_earlier_versions_of_sqlite
+				 *
+				 * TODO: Remove this flag when we drop support for PHP 8.0.
+				 *       From PHP 8.1, SQLite 3.46.1 is used by default.
+				 */
+				$this->execute_sqlite_query( 'PRAGMA writable_schema=ON' );
+			} else {
+				throw $this->new_driver_exception(
+					sprintf(
+						'The SQLite version %s is not supported. Minimum required version is %s.',
+						$sqlite_version,
+						self::MINIMUM_SQLITE_VERSION
+					)
+				);
+			}
 		}
 
 		// Load SQLite version to a property used by WordPress health info.
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -484,6 +484,19 @@ class WP_SQLite_Driver {
 		$sqlite_version = $this->get_sqlite_version();
 		if ( version_compare( $sqlite_version, self::MINIMUM_SQLITE_VERSION, '<' ) ) {
 			if ( defined( 'WP_SQLITE_UNSAFE_ENABLE_UNSUPPORTED_VERSIONS' ) && WP_SQLITE_UNSAFE_ENABLE_UNSUPPORTED_VERSIONS ) {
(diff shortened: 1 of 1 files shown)


sejas (commented):
Thanks for trying to make work the STRICT tables. I tested this PR on a machine with SQLite 3.34.1 , but I'm getting a weird error when enabling the new AST driver. `Database directory is not writable: /srv/htdocs/wp-content/database`. There is no more information in debug.log

Screenshot 2025-10-01 at 11 38 25

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -483,13 +483,35 @@ public function __construct( WP_SQLite_Connection $connection, string $database
...
 		$sqlite_version = $this->get_sqlite_version();
 		if ( version_compare( $sqlite_version, self::MINIMUM_SQLITE_VERSION, '<' ) ) {
-			throw $this->new_driver_exception(
-				sprintf(
-					'The SQLite version %s is not supported. Minimum required version is %s.',
-					$sqlite_version,
-					self::MINIMUM_SQLITE_VERSION
-				)
-			);
+			if ( defined( 'WP_SQLITE_UNSAFE_ENABLE_UNSUPPORTED_VERSIONS' ) && WP_SQLITE_UNSAFE_ENABLE_UNSUPPORTED_VERSIONS ) {
```
akirk, line 486: I think we should still keep a (lower than above) minimum sqlite version check here.

adamziel: +1

JanJakes: Good idea! On PHP 7.2, they seem to package `3.27.2` by default:

```
docker run --rm php:7.2 -r "print_r((new PDO('sqlite::memory'))->query('SELECT SQLITE_VERSION();')->fetch());"
```

And given that `3.27.1` and `3.27.2` were bugfix releases, I think we can go with `3.27.0` (February 2019).

JanJakes:
@adamziel @akirk Updated with 3.27.0 as a minimum.

sejas (approved):
Thanks to Jan, I was able to test this PR. I confirm that enabling the correct constants made the WordPress site work on SQLite 3.34.1.

Modified constants:
- `define ( 'WP_SQLITE_AST_DRIVER', true );`
- `define ( 'WP_SQLITE_UNSAFE_ENABLE_UNSUPPORTED_VERSIONS', true );` 
- And setting the correct `DB_NAME` : `define ( 'DB_NAME', 'database_name_here' );

I wonder if defining the `WP_SQLITE_UNSAFE_ENABLE_UNSUPPORTED_VERSIONS` constant is necessary, or if it should be enabled automatically when the SQLite version is older and AST is enabled.

JanJakes:
@sejas

> I wonder if defining the `WP_SQLITE_UNSAFE_ENABLE_UNSUPPORTED_VERSIONS` constant is necessary, or if it should be enabled automatically when the SQLite version is older and AST is enabled.

With this simple implementation, it's probably necessary (as per the `An important consideration or TODO` in the description of this issue). But even if we skip the `STRICT` keyword in all CREATE statements on older versions (= fully implement the mentioned consideration), I would be careful about this hack and would rather caution against it than activate it automatically. It can cause some incompatibilities when writing to the database, and it's technically possible to save incorrect data types, which would make the database file "corrupt" from the perspective of newer SQLite versions.

PR #261, opened 2025-10-03
Add WP_SQLITE_DB_NAME constant to take precedence over DB_NAME

sejas:
- Related to https://github.com/Automattic/studio/pull/1839

Because [filters actually don’t work in db.php](https://github.com/WordPress/sqlite-database-integration/pull/260), I suggest removing the `wp_sqlite_default_db_name` filter and introducing a new constant `WP_SQLITE_DB_NAME` to take precedence over DB_NAME.

* Break your database connection:
  * Edit your wp-config.php
  * Copy the value of your DB_NAME
  * Change it to any value to break your database connection
  * Access the homepage of your site and observe the error: `Error establishing a database connection`
* Fix your database connection using the new constant:
  * Edit your wp-config.php
  * Add the new constant WP_SQLITE_DB_NAME with the value you copied before. Example `define( 'WP_SQLITE_DB_NAME', 'database_name_here' );`
  * Access your site and observe it loads correctly

Credit to @wojtekn for the code. I just [removed the filter](https://github.com/WordPress/sqlite-database-integration/pull/261/commits/8f5de999c950ad4ed76244489613be3b29c1c56c) as it seemed misleading.

PR #262, opened 2025-10-06
Inject database name in queries dynamically

JanJakes:
With the new SQLite driver, we keep running into issues with missing and incorrect database name values (https://github.com/WordPress/sqlite-database-integration/pull/197, https://github.com/WordPress/sqlite-database-integration/pull/203, https://github.com/WordPress/sqlite-database-integration/pull/226, https://github.com/WordPress/sqlite-database-integration/pull/260, https://github.com/WordPress/sqlite-database-integration/pull/261, and other issues).

Thinking through this further, I came to the conclusion that to provide maximum flexibility and portability, it would be best to provide an API in the shape of:

_"Mount `my-sqlite-file.sqlite` as `my-db-name`."_

Even if we consider adding multi-database support one day, it would still be great to allow mounting additional SQLite files under dynamic database names.

I don't see any downsides to doing this, except maybe that we'll have to choose some database name in some scenarios, such as in database admin tools. However, in these cases we can use a reasonable default or the file name.

This is a WIP pull request demonstrating the simplest approach. For it to be merged, I only need to resolve how the existing database name values should be treated.

### DIFF
--- a/tests/WP_SQLite_Driver_Tests.php
+++ b/tests/WP_SQLite_Driver_Tests.php
@@ -9244,4 +9244,81 @@ END;
 			$result[0]->{'Create Table'}
 		);
 	}
+
+	public function testDynamicDatabaseName(): void {
+		// Create a setter for the private property "$db_name".
+		$set_db_name = Closure::bind(
+			function ( $name ) {
+				$this->main_db_name = $name;
+			},
+			$this->engine,
+			WP_SQLite_Driver::class
+		);
+
+		// Default database name.
+		$result = $this->assertQuery( 'SELECT schema_name FROM information_schema.schemata ORDER BY schema_name' );
+		$this->assertEquals(
+			array(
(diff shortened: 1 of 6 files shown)


sejas (commented):
@JanJakes, thanks for working on this issue. If I understand correctly, this change lets the SQLite plugin mount the database with any DB_NAME automatically. The constant is still needed, but its value no longer affects accessing the database.

sejas:
@JanJakes, do you think this is good to merge, or should we go with the [WP_SQLITE_DB_NAME](https://github.com/WordPress/sqlite-database-integration/pull/261) constant in the meantime?

JanJakes:
@sejas I'll try to finish this one now. What it misses is support for wildcards like `SELECT * FROM infromation_schema.tables`, or more complex ones with JOINs, etc. I'll give it a try.

JanJakes:
I'm closing this in favor of https://github.com/WordPress/sqlite-database-integration/pull/266.

PR #266, opened 2025-10-14
Dynamic database name

JanJakes:
This PR replaces https://github.com/WordPress/sqlite-database-integration/pull/262. It is a simpler solution to the same problem:

When a table reference targets an information schema table, we replace it with a subquery, injecting the configured database name dynamically:

```sql
-- The following query:
SELECT *, t.*, t.table_schema FROM information_schema.tables t

-- Will be translated to:
SELECT *, t.*, t.table_schema FROM (
  SELECT
    `TABLE_CATALOG`,
    IIF(`TABLE_SCHEMA` = 'information_schema', `TABLE_SCHEMA`, 'database_name') AS `TABLE_SCHEMA`,
    `TABLE_NAME`,
    ...
  FROM _wp_sqlite_mysql_information_schema_tables AS tables
) t
```
The same logic will be applied to table references in JOIN clauses as well.

---

With the new SQLite driver, we keep running into issues with missing and incorrect database name values (https://github.com/WordPress/sqlite-database-integration/pull/197, https://github.com/WordPress/sqlite-database-integration/pull/203, https://github.com/WordPress/sqlite-database-integration/pull/226, https://github.com/WordPress/sqlite-database-integration/pull/260, https://github.com/WordPress/sqlite-database-integration/pull/261, and other issues).

Thinking through this further, I came to the conclusion that to provide maximum flexibility and portability, it would be best to provide an API in the shape of:

_"Mount `my-sqlite-file.sqlite` as `my-db-name`."_

Even if we consider adding multi-database support one day, it would still be great to allow mounting additional SQLite files under dynamic database names.

I don't see any downsides to doing this, except maybe that we'll have to choose some database name in some scenarios, such as in database admin tools. However, in these cases we can use a reasonable default or the file name.

This is a WIP pull request demonstrating the simplest approach. For it to be merged, I only need to resolve how the existing database name values should be treated.

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -3973,10 +3973,21 @@ class WP_SQLite_Driver {
 
 			// List all columns in the table, replacing columns targeting database
 			// name columns with the configured database name.
+			static $information_schema_db_column_map = array(
+				'SCHEMA_NAME'              => true,
+				'TABLE_SCHEMA'             => true,
+				'VIEW_SCHEMA'              => true,
+				'INDEX_SCHEMA'             => true,
+				'CONSTRAINT_SCHEMA'        => true,
+				'UNIQUE_CONSTRAINT_SCHEMA' => true,
+				'REFERENCED_TABLE_SCHEMA'  => true,
+				'TRIGGER_SCHEMA'           => true,
+			);
+
 			$expanded_list = array();
 			foreach ( $columns as $column ) {
 				$quoted_column = $this->quote_sqlite_identifier( $column );
-				if ( str_contains( strtolower( $column ), 'schema' ) ) {
+				if ( isset( $information_schema_db_column_map[ strtoupper( $column ) ] ) ) {
 					$expanded_list[] = sprintf(
 						"IIF(%s = 'information_schema', %s, %s) AS %s",
 						$quoted_column,
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -3353,7 +3353,7 @@ class WP_SQLite_Driver {
 			);
 			if ( 'information_schema' === strtolower( $schema_name ) ) {
 				$is_information_schema = true;
-			} elseif ( $this->db_name === $schema_name ) {
+			} elseif ( $this->main_db_name === $schema_name ) {
 				$is_information_schema = false;
 			} else {
 				throw $this->new_not_supported_exception(
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -2271,9 +2271,16 @@ class WP_SQLite_Driver {
 
 		$databases = $this->execute_sqlite_query(
 			sprintf(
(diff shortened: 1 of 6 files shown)


wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -3939,6 +3914,93 @@ public function translate_select_item( WP_Parser_Node $node ): string {
...
+				'SELECT name FROM pragma_table_info(?)',
+				array( $sqlite_table_name )
+			)->fetchAll( PDO::FETCH_COLUMN );
+
+			// List all columns in the table, replacing columns targeting database
+			// name columns with the configured database name.
+			$expanded_list = array();
+			foreach ( $columns as $column ) {
+				$quoted_column = $this->quote_sqlite_identifier( $column );
+				if ( str_contains( strtolower( $column ), 'schema' ) ) {
```
adamziel: this is quite permissive – can we explicitly list the tables and columns we're interested in?

adamziel:
@JanJakes this looks great for information schema tables. However, is this enough to support a dynamic database name? What will happen for queries such as `SELECT * FROM my_db_name.my_table`?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel Ah, sure, we can! In this place, we know for sure that we're dealing with an information schema table column, so maybe we could also check if it contains `schema_` or `_schema`, although listing is also easy—this should cover it:

```
'SCHEMA_NAME',
'TABLE_SCHEMA',
'VIEW_SCHEMA',
'INDEX_SCHEMA',
'CONSTRAINT_SCHEMA',
'UNIQUE_CONSTRAINT_SCHEMA',
'REFERENCED_TABLE_SCHEMA',
'TRIGGER_SCHEMA',
```

Or we could even list the exact columns for each table.

JanJakes:
> @JanJakes this looks great for information schema tables. However, is this enough to support a dynamic database name? What will happen for queries such as SELECT * FROM my_db_name.my_table?

@adamziel Using the DB name in the query works. A query like `SELECT * FROM information_schema.tables` will be translated to:
```sql
SELECT *
FROM (
  SELECT
    `TABLE_CATALOG`,
    IIF(`TABLE_SCHEMA` = 'information_schema', `TABLE_SCHEMA`, 'wp') AS `TABLE_SCHEMA`,
    `TABLE_NAME`,
    `TABLE_TYPE`,
    `ENGINE`,
    `VERSION`,
    `ROW_FORMAT`,
    `TABLE_ROWS`,
    `AVG_ROW_LENGTH`,
    `DATA_LENGTH`,
    `MAX_DATA_LENGTH`,
    `INDEX_LENGTH`,
    `DATA_FREE`,
    `AUTO_INCREMENT`,
    `CREATE_TIME`,
    `UPDATE_TIME`,
    `CHECK_TIME`,
    `TABLE_COLLATION`,
    `CHECKSUM`,
    `CREATE_OPTIONS`,
    `TABLE_COMMENT`
  FROM `_wp_sqlite_mysql_information_schema_tables` AS `tables`
)
```

JanJakes:
@adamziel Oh, but I didn't add any tests for this:

```sql
USE information_schema;
SELECT * FROM tables;
SELECT * FROM my_db.tables;
```

Will do and make sure it works.

adamziel (approved):

JanJakes:
@adamziel I pushed 3 more commits, all should be addressed now.

Issue #268, opened 2025-10-15
SQLite driver fails to insert TEXT into longblob column


wojtekn:
When running the Wordfence installation SQL script on SQLite, the INSERT statement for the wp_wfconfig table fails with the following error:

```
WP_SQLite_Driver_Exception: SQLSTATE[23000]: Integrity constraint violation: 19 cannot store TEXT value in BLOB column wp_wfconfig.val
```

The issue occurs because the table defines val as longblob, but the insert statement provides a TEXT value ('2'). SQLite enforces strict typing and disallows implicit TEXT-to-BLOB conversion.

[wp_wfconfig.sql](https://github.com/user-attachments/files/22925024/wp_wfconfig.sql)

PR #273, opened 2025-10-21
Add support for `INSERT INTO ... SET ...` syntax

JanJakes:
Some popular plugins use the `INSERT INTO … SET …` syntax that is specific to MySQL.

In SQLite, this needs to be translated into the standard `INSERT INTO ( … ) VALUES ( … )` syntax.

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -1529,7 +1529,15 @@ class WP_SQLite_Driver {
 
 		$parts = array();
 		foreach ( $node->get_children() as $child ) {
-			if ( $child instanceof WP_MySQL_Token && WP_MySQL_Lexer::IGNORE_SYMBOL === $child->id ) {
+			$is_token = $child instanceof WP_MySQL_Token;
+			$is_node  = $child instanceof WP_Parser_Node;
+
+			// Skip the SET keyword in "INSERT INTO ... SET ..." syntax.
+			if ( $is_token && WP_MySQL_Lexer::SET_SYMBOL === $child->id ) {
+				continue;
+			}
+
+			if ( $is_token && WP_MySQL_Lexer::IGNORE_SYMBOL === $child->id ) {
 				// Translate "UPDATE IGNORE" to "UPDATE OR IGNORE".
 				$parts[] = 'OR IGNORE';
 			} elseif (
@@ -1540,6 +1548,16 @@ class WP_SQLite_Driver {
 				$table_ref  = $node->get_first_child_node( 'tableRef' );
 				$table_name = $this->unquote_sqlite_identifier( $this->translate( $table_ref ) );
 				$parts[]    = $this->translate_insert_or_replace_body_in_non_strict_mode( $table_name, $child );
+			} elseif ( $is_node && 'updateList' === $child->rule_name ) {
+				// Convert "SET c1 = v1, c2 = v2, ... to "(c1, c2, ...) VALUES (v1, v2, ...)".
+				$columns = array();
+				$values  = array();
+				foreach ( $child->get_child_nodes( 'updateElement' ) as $update_element ) {
+					$column_ref = $update_element->get_first_child_node( 'columnRef' );
+					$columns[]  = $this->translate( $column_ref );
+					$values[]   = $this->translate( $update_element->get_first_child_node( 'expr' ) );
+				}
+				$parts[] = '(' . implode( ', ', $columns ) . ') VALUES (' . implode( ', ', $values ) . ')';
 			} else {
 				$parts[] = $this->translate( $child );
 			}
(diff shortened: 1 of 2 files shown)


wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -1529,17 +1529,39 @@ private function execute_insert_or_replace_statement( WP_Parser_Node $node ): vo
 
 		$parts = array();
 		foreach ( $node->get_children() as $child ) {
-			if ( $child instanceof WP_MySQL_Token && WP_MySQL_Lexer::IGNORE_SYMBOL === $child->id ) {
+			$is_token = $child instanceof WP_MySQL_Token;
+			$is_node  = $child instanceof WP_Parser_Node;
+
+			// Skip the SET keyword in "INSERT INTO ... SET ..." syntax.
+			if ( $is_token && WP_MySQL_Lexer::SET_SYMBOL === $child->id ) {
```
adamziel, line 1536: Any chance we ever see a SET symbol in an unexpected place? The tokenizing parser made me hesitant about this pattern of going through all the child nodes up to a symbol, e.g. `INSERT INTO x (SELECT ... SET ...)`. I expect the answer is "no" since a) I don't think there's a way to cram in SET keyword into a select statement and b) that would likely be handled in a different AST node handler and c) get_children() is just for direct children, not descendants, right? Still, I wanted to voice it.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -1529,17 +1529,39 @@ private function execute_insert_or_replace_statement( WP_Parser_Node $node ): vo
...
 				$parts[] = 'OR IGNORE';
 			} elseif (
 				$is_non_strict_mode
-				&& $child instanceof WP_Parser_Node
-				&& ( 'insertFromConstructor' === $child->rule_name || 'insertQueryExpression' === $child->rule_name )
+				&& $is_node
+				&& (
+					'insertFromConstructor' === $child->rule_name
+					|| 'insertQueryExpression' === $child->rule_name
+					|| 'updateList' === $child->rule_name
```
adamziel, line 1549: Oh, so we handle `updateList` both in this else/if branch and in the next one, just in different ways?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -4304,6 +4326,12 @@ private function translate_insert_or_replace_body_in_non_strict_mode(
 			foreach ( $fields_node->get_child_nodes() as $field ) {
 				$insert_list[] = $this->unquote_sqlite_identifier( $this->translate( $field ) );
 			}
+		} elseif ( 'updateList' === $node->rule_name ) {
```
adamziel, line 4329: Aha! so this is for `INSERT INTO table VALUES (val1, val2)` and the other code branch is for `INSERT INTO table (col1, col2) VALUES (val1, val2)`. Do we need to support both? Or could we just go with the one that lists columns explicitly?

adamziel (approved):

adamziel:
I left a few thoughts, but nothing important. Thanks Jan!

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel Yeah, it is safe, because we're iterating direct children and the grammar is:
```
insertStatement:
    INSERT_SYMBOL insertLockOption? IGNORE_SYMBOL? INTO_SYMBOL? tableRef usePartition? (
        insertFromConstructor ({ serverVersion >= 80018}? valuesReference)?
        | SET_SYMBOL updateList ({ serverVersion >= 80018}? valuesReference)?
        | insertQueryExpression
    ) insertUpdateList?
;
```

That said, I may refactor this to something nicer in a subsequent PR with a fix for https://github.com/WordPress/sqlite-database-integration/issues/268.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel At the moment, it is twice—once for the "STRICT" SQL mode and once for the WP-like "non-STRICT" mode.

For the non-STRICT mode, we apply implicit defaults, type casting, etc. based on the data from the information schema. For the STRICT mode, we just pass it "directly" to SQLite.

That said, I will have to apply type casting in both STRICT and non-STRICT mode due to issues like https://github.com/WordPress/sqlite-database-integration/issues/268, so in a subsequent PR, I will try to unify this better. It will also allow us rejecting/trimming values that are too long, based on the current SQL mode.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: In MySQL, you can use all of `INSERT INTO table VALUES (...)`, `INSERT INTO table (col1, col2, ...) VALUES (...)`, and `INSERT INTO table SET col1 = val1, ...`, so we need to support all of them.

JanJakes:
I will go ahead and merge this one and keep the remarks in mind for the subsequent PR.

PR #274, opened 2025-10-22
Fix "Expression tree is too large" for strings with large number of `NULL` bytes

JanJakes:
In SQLite, it is generally difficult to work with `NULL` bytes in strings, because SQLite uses C-style string syntax, which means it considers the first `NULL` byte as the string terminator in some scenarios.

While it is possible to store and read `NULL` bytes within strings (with some limitations), there is no way to express the `NULL` byte in a string literal. It needs to be done in a different way:

1. Concatenation: `str_part1 || CHAR(0) || str_part2 || …`
2. Passing as a parameter to a prepared statement.
3. Using HEX notation and casting to string.

More details: https://sqlite.org/nulinstr.html

At the moment, the SQLite driver uses concatenation (option 1), which seems to fail for strings that contain many `NULL` bytes:

```
SQLSTATE[HY000]: General error: 1 Expression tree is too large (maximum depth 1000)
```

This PR fixes this by using HEX encoding.

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -3339,7 +3339,7 @@ class WP_SQLite_Driver {
 		$value = $token->get_value();
 
 		/*
-		 * 5. Translate datetime literals.
+		 * Translate datetime literals.
 		 *
 		 * Process only strings that could possibly represent a datetime
 		 * literal ("YYYY-MM-DDTHH:MM:SS", "YYYY-MM-DDTHH:MM:SSZ", etc.).
@@ -3349,7 +3349,7 @@ class WP_SQLite_Driver {
 		}
 
 		/*
-		 * 6. Handle null characters.
+		 * Handle null characters.
 		 *
 		 * SQLite doesn't fully support null characters (\u0000) in strings.
 		 * However, it can store them and read them, with some limitations.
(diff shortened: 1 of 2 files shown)


wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -3358,25 +3358,26 @@ private function translate_string_literal( WP_Parser_Node $node ): string {
...
-			// Escape and quote each segment.
-			$parts[] = "'" . str_replace( "'", "''", $segment ) . "'";
-		}
-		if ( count( $parts ) > 1 ) {
-			return '(' . implode( ' || CHAR(0) || ', $parts ) . ')';
+		if ( strpos( $value, "\0" ) !== false ) {
+			return sprintf( "CAST(x'%s' AS TEXT)", bin2hex( $value ) );
 		}
-		return $parts[0];
+		return sprintf( "'%s'", str_replace( "'", "''", $value ) );
```
adamziel: don't we have some helper for escaping string values?

JanJakes: @adamziel Ahh, good point. This is a string value so the native PDO\Sqlite::quote() should work here. Pushed in https://github.com/WordPress/sqlite-database-integration/pull/274/commits/455a139d23221ba8820ad46090212a4696b4285f.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -3358,25 +3358,26 @@ private function translate_string_literal( WP_Parser_Node $node ): string {
...
 		 *   https://www.sqlite.org/nulinstr.html
 		 */
-		$parts = array();
-		foreach ( explode( "\0", $value ) as $segment ) {
-			// Escape and quote each segment.
-			$parts[] = "'" . str_replace( "'", "''", $segment ) . "'";
-		}
-		if ( count( $parts ) > 1 ) {
-			return '(' . implode( ' || CHAR(0) || ', $parts ) . ')';
+		if ( strpos( $value, "\0" ) !== false ) {
```
adamziel, line 3377: I like that you're only doing that optionally, this way we most queries remain in a human-readable form for SQL monitor and such. Good call!

adamziel (approved):

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel Yeah. Also, the HEX notation makes the values longer.

PR #276, opened 2025-10-23
Implement `INSERT` and `UPDATE` value type casting

### LINKED ISSUE #268: SQLite driver fails to insert TEXT into longblob column

JanJakes:
This PR originates from the following error (https://github.com/WordPress/sqlite-database-integration/issues/268):
```
SQLSTATE[23000]: Integrity constraint violation:
19 cannot store TEXT value in BLOB column wp_wfconfig.val
```

However, there isn't a simple fix. It requires correctly casting values for `INSERT`, `REPLACE`, and `UPDATE` statements, and that's what this PR implements. Type casting logic, together with "implicit defaults", was already implemented for MySQL's "non-strict" mode. This PR extends the type casting part to be used with both strict and non-strict modes.

Generally, it does the following:
```sql
-- Rewrite INSERT statement from:
INSERT INTO t (col1, col2, ...) VALUES (v1, v2, ...);

-- To:
INSERT INTO t (col1, col2, ...)
SELECT <apply-type-casting-to-select-list> FROM (VALUES (v1, v2, ...)) WHERE true;

-- Rewrite UPDATE statement from:
UPDATE t SET col1 = value;

-- To:
UPDATE t SET col1 = <apply-type-casting-to-value>;
```

The PR comes with a comprehensive test suite, but it doesn't implement all tiny MySQL nuances and acknowledges some differences in TODO comments. It should increase the type compatibility when saving values significantly. We can address more edge cases separately, and it should also enable us checking for column lengths and trimming or enforcing the values based on the SQL mode.

Fixes https://github.com/WordPress/sqlite-database-integration/issues/268.

### DIFF
--- a/tests/WP_SQLite_Driver_Tests.php
+++ b/tests/WP_SQLite_Driver_Tests.php
@@ -28,7 +28,7 @@ class WP_SQLite_Driver_Tests extends TestCase {
 			"CREATE TABLE _dates (
 					ID INTEGER PRIMARY KEY AUTO_INCREMENT NOT NULL,
 					option_name TEXT NOT NULL default '',
-					option_value DATE NOT NULL
+					option_value DATETIME NOT NULL
 				);"
 		);
 	}
@@ -37,6 +37,17 @@ class WP_SQLite_Driver_Tests extends TestCase {
 		$retval = $this->engine->query( $sql );
 		$this->assertNotFalse( $retval );
 		return $retval;
+	}
+
+	private function assertQueryError( $sql, $error_message ) {
+		$exception = null;
+		try {
+			$this->engine->query( $sql );
+		} catch ( WP_SQLite_Driver_Exception $e ) {
+			$exception = $e;
+		}
+		$this->assertNotNull( $exception, 'An exception was expected, but none was thrown.' );
+		$this->assertSame( $error_message, $exception->getMessage() );
 	}
 
 	public function testRegexp() {
@@ -2342,6 +2353,8 @@ class WP_SQLite_Driver_Tests extends TestCase {
 	}
 
 	public function testTruncatesInvalidDates() {
+		$this->assertQuery( "SET sql_mode = ''" );
+
 		$this->assertQuery( "INSERT INTO _dates (option_value) VALUES ('2022-01-01 14:24:12');" );
 		$this->assertQuery( "INSERT INTO _dates (option_value) VALUES ('2022-31-01 14:24:12');" );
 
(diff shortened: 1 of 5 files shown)


wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -4423,55 +4390,102 @@ private function translate_show_like_or_where_condition( WP_Parser_Node $like_or
...
-				$insert_list[] = $column_name;
+				$insert_list[] = strtolower( $column_name );
 			}
 		}
 
-		// 3. Filter out omitted columns that will get a value from the SQLite engine.
-		//    That is, nullable columns, columns with defaults, and generated columns.
+		// Check if all listed columns exist.
+		$unkwnown_columns = array_diff( $insert_list, array_column( $columns, 'COLUMN_NAME' ) );
+		if ( count( $unkwnown_columns ) > 0 ) {
```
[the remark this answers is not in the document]
JanJakes: @copilot I don't like this change. A lot has been said and written about how `empty()` is problematic, and in this case, perhaps also slower.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -4423,55 +4390,102 @@ private function translate_show_like_or_where_condition( WP_Parser_Node $like_or
...
+					$is_omitted = ! isset( $insert_map[ $column['COLUMN_NAME'] ] );
 					if ( ! $is_omitted ) {
 						return true;
 					}
+					if ( $is_strict_mode ) {
+						return false;
+					}
 					$is_nullable  = 'YES' === $column['IS_NULLABLE'];
 					$has_default  = $column['COLUMN_DEFAULT'];
 					$is_generated = str_contains( $column['EXTRA'], 'auto_increment' );
```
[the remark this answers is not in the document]
JanJakes: @copilot This project has polyfills for these str functions.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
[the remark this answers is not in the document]
JanJakes: Can I discard this suggestion?

tests/WP_SQLite_Driver_Tests.php
```diff
@@ -10053,4 +10069,1069 @@ public function testEmptyColumnMeta(): void {
...
+		$this->assertQuery( 'INSERT INTO t VALUES (NULL)' );
+		$this->assertQuery( 'INSERT INTO t VALUES (FALSE)' );
+		$this->assertQuery( 'INSERT INTO t VALUES (TRUE)' );
+		$this->assertQuery( 'INSERT INTO t VALUES (0)' );
+		$this->assertQuery( 'INSERT INTO t VALUES (1)' );
+		$this->assertQuery( "INSERT INTO t VALUES ('2')" );
+		$this->assertQuery( "INSERT INTO t VALUES ('3.0')" );
+
+		// TODO: These are supported in MySQL:
+		$this->assertQueryError( "INSERT INTO t VALUES ('4.5')", 'SQLSTATE[23000]: Integrity constraint violation: 19 cannot store REAL value in INTEGER column t.value' );
```
adamziel, line 10085: Good call describing known limitations of the system with tests

wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
```diff
@@ -88,6 +89,18 @@ public static function register_for( PDO $pdo ): self {
...
 
+	/**
+	 * A helper function to throw an error from SQLite expressions.
+	 *
+	 * @param string $message The error message.
+	 *
+	 * @throws Exception The error message.
+	 * @return void
+	 */
+	public function throw( $message ): void {
```
adamziel, line 100: I'm surprised this is not a reserved keyword and makes a valid method name. [PHP docs](https://www.php.net/manual/en/reserved.keywords.php) confirms it's safe:

> They are, however, allowed as property, constant, and method names of classes, interfaces and traits, except that class may not be used as constant name.

TIL

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -4423,55 +4390,102 @@ private function translate_show_like_or_where_condition( WP_Parser_Node $like_or
...
-			array( $this->get_saved_db_name(), $table_name )
+			array( $database, $table_name )
 		)->fetchAll( PDO::FETCH_ASSOC );
 
-		// 2. Get the list of fields explicitly defined in the INSERT statement.
+		// Check if the table exists.
+		if ( 0 === count( $columns ) ) {
+			throw $this->new_driver_exception(
+				sprintf(
+					"SQLSTATE[42S02]: Base table or view not found: 1146 Table '%s' doesn't exist",
```
adamziel, line 4427: MySQL error codes! Nice

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -4527,15 +4570,11 @@ function ( $column ) use ( $insert_list ) {
 				// When a column value is included, we need to apply type casting.
 				$position   = array_search( $column['COLUMN_NAME'], $insert_list, true );
 				$identifier = $this->quote_sqlite_identifier( $select_list[ $position ] );
-				$fragment  .= sprintf(
-					'%s AS %s',
-					$this->cast_value_in_non_strict_mode( $column['DATA_TYPE'], $identifier ),
-					$identifier
-				);
+				$fragment  .= $this->cast_value_for_insert_or_update( $column['DATA_TYPE'], $identifier );
```
adamziel: `cast_value_for_insert_or_update` – Any other statements that need casting? E.g. `REPLACE INTO`?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -4568,25 +4607,45 @@ function ( $column ) use ( $insert_list ) {
...
 	 * To a list with the following structure:
-	 *   UPDATE table SET <non-null-column> = COALESCE(<value>, <implicit-default>)
+	 *   UPDATE table SET <column> = <adjusted-value>
 	 *
 	 * In MySQL, the behavior of INSERT and UPDATE statements depends on whether
 	 * the STRICT_TRANS_TABLES (InnoDB) or STRICT_ALL_TABLES SQL mode is enabled.
 	 *
-	 * When the strict mode is not enabled, executing an UPDATE statement that
-	 * sets a NOT NULL column value to NULL saves an IMPLICIT DEFAULT instead.
+	 * This method applies relevant type casting and emulates IMPLICIT DEFAULT
```
adamziel, line 4624: Let's drop a couple of links to the relevant MySQL doc page in a few places as a reference:

https://dev.mysql.com/doc/refman/8.4/en/data-type-defaults.html

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -4927,23 +5009,27 @@ private function create_table_reference_map( WP_Parser_Node $node ): array {
...
 	 * @param  string $translated_value The original translated value.
 	 * @return string                   The translated value.
 	 */
-	private function cast_value_in_non_strict_mode(
+	private function cast_value_for_insert_or_update(
 		string $mysql_data_type,
 		string $translated_value
 	): string {
-		$sqlite_data_type = self::DATA_TYPE_STRING_MAP[ $mysql_data_type ];
+		// TODO: This is also a good place to implement checks for maximum column
```
adamziel, line 5032: 👍

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -4977,21 +5063,72 @@ private function cast_value_in_non_strict_mode(
 				} elseif ( 'datetime' === $mysql_data_type || 'timestamp' === $mysql_data_type ) {
 					$function_call = sprintf( 'DATETIME(%s)', $translated_value );
 				} elseif ( 'year' === $mysql_data_type ) {
-					$function_call = sprintf( "STRFTIME('%%Y', %s)", $translated_value );
+					/*
+					 * The YEAR type in MySQL only uses 1 byte and therefore
```
adamziel, line 5077: good comment

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -4927,23 +5009,27 @@ private function create_table_reference_map( WP_Parser_Node $node ): array {
...
 	/**
-	 * Emulate MySQL type casting for INSERT or UPDATE value in non-strict mode.
+	 * Emulate MySQL type casting for INSERT or UPDATE values.
 	 *
 	 * @param  string $mysql_data_type  The MySQL data type.
 	 * @param  string $translated_value The original translated value.
 	 * @return string                   The translated value.
 	 */
-	private function cast_value_in_non_strict_mode(
+	private function cast_value_for_insert_or_update(
```
adamziel: Do we need to consider any potential interferences with SQLite's type affinity?

```sql
sqlite> create table t(a int);
sqlite> insert into t(a) values (22), ('33');
sqlite> select a from t where a = '22' or a = 33;
22
33
```

https://www.sqlite.org/datatype3.html

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -4977,21 +5063,72 @@ private function cast_value_in_non_strict_mode(
...
+					);
 				}
 
-				// When the function call evaluates to NULL (invalid date/time),
-				// we need to fallback to the IMPLICIT DEFAULT value.
+				// In strict mode, invalid date/time values are rejected.
+				// In non-strict mode, they get an IMPLICIT DEFAULT value.
+				if ( $is_strict_mode ) {
+					$fallback = sprintf(
+						"THROW('Incorrect %s value: ''' || %s || '''')",
```
adamziel, line 5106: Such a nice use of callbacks 🤯 ! At the same time – could we just throw in place? Do we need to delay it? What's the benefit of doing this vs just throwing a new exception right here?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel Yeah, `REPLACE` as well. Maybe we could use a more generic name like `cast_value_for_saving` or something along those lines?

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel This particular example [behaves the same in MySQL](https://onecompiler.com/mysql/4436wgz8b), but generally, it can be a problem when an expression evaluates types differently, or comparisons are automatically type casted in one DB but not in the other one.

```sql
CREATE TABLE t(a DATE);
INSERT INTO t(a) VALUES ('2021-01-01');
SELECT * FROM t WHERE a = '20210101'
```
This returns one row in MySQL but no rows in SQLite. It's probably difficult to solve—it would require inferring the correct type of every part of every expression. Fortunately, the most common numeric string cast seems to work in both DBs.

JanJakes: @adamziel Checking further, the situation is not that good:

```sql
SELECT 1 = '1';         -- true in MySQL, false in SQLite
SELECT 1 WHERE 1 = '1'; -- one row in MySQL, no rows in SQLite
```
So it appears that the numeric string type cast is only applied when comparing against a numeric column.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
JanJakes: @adamziel We need to throw the error from SQLite, because in PHP we don't know the value. You can insert known values (`INSERT INTO ... VALUES (1, 2, 3)`), as well as unknown ones `INSERT INTO ... SELECT ...`. For that reason, we need SQLite to evaluate this for every inserted/updated row, as we can't do that in PHP. In PHP the "value" can be a column name, an expression, etc.

JanJakes:
@adamziel I added two commits with small fixes.

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
adamziel: We can probably come up with a sqlite expression equivalent to `MYSQL_TYPECAST( value, column_type ) => value_in_column_type` that pattern-matches and transforms the value in the same way as mysql-server would. Maybe we can even source the logic from the actual mysql-server codebase

JanJakes: @adamziel I think that's basically what the `cast_value_for_saving()` method does. It knows the column type, and it has a value expression, and then it wraps it in a casting expression. We can make it more exhaustive and handle all edge cases, but the core logic is there.

What will be a problem, though, is expressions where we have no column type at all—things like `WHERE '1' = 1`, but with arbitrary complexity. Anyway, that's probably a topic for another PR.

adamziel: Separate pr sounds good! The transform there would be more like `(lhs value, lhs type, comparison, rhs value, rhs type)` -> expression that evaluates to true or false using mysql rules.

JanJakes: @adamziel Oh, right. Seems like it's definitely better to tackle this separately when needed. Rewriting all expressions on every nesting level seems like a bigger topic and it could possibly cause some performance issues (like index not being used due to inline casting, etc.).

Otherwise, do you think this is good to merge?

adamziel: Yup!

PR #283, opened 2025-11-05
Make it work with PHP built with static-php-cli

prashanthca:
Thank you so much for the work on the plugin, it's a treat to setup and use!

The issue this PR is trying to solve surfaced while using a static build of PHP and PHP-FPM (v8.4.14) built using @crazywhalecc/static-php-cli. I'm using the "bulk" pre-built version of the PHP and PHP-FPM binaries, which contains pretty-much all of the essential PHP extensions built-in, including SQLite.

But unfortunately the wordpress setup crashes with "PDO Driver for SQLite is missing", after installing the sqlite-database-integration plugin (and following through with the necessary steps).

Checking through the code, I found that the SQLite PHP extension is detected by checking for the pdo_sqlite extension, while the SQLite extension built-into the static PHP build is called: sqlite3. 

Replacing all occurrences of "pdo_sqlite" with "sqlite3" in the extension source code fixes the problem. But instead of replacing strings, this PR changes the logic, wherever used, to crash only if neither of the PHP extensions are installed

### DIFF
--- a/activate.php
+++ b/activate.php
@@ -76,7 +76,7 @@ add_filter(
  */
 function sqlite_plugin_copy_db_file() {
 	// Bail early if the PDO SQLite extension is not loaded.
-	if ( ! extension_loaded( 'pdo_sqlite' ) ) {
+	if ( ! extension_loaded( 'pdo_sqlite' ) && ! extension_loaded( 'sqlite3' ) ) {
 		return;
 	}
 
--- a/admin-notices.php
+++ b/admin-notices.php
@@ -20,7 +20,7 @@ function sqlite_plugin_admin_notice() {
 	}
 
 	// If PDO SQLite is not loaded, bail early.
-	if ( ! extension_loaded( 'pdo_sqlite' )  && ! extension_loaded( 'sqlite3' )) {
+	if ( ! extension_loaded( 'pdo_sqlite' ) && ! extension_loaded( 'sqlite3' ) ) {
 		printf(
 			'<div class="notice notice-error"><p>%s</p></div>',
 			esc_html__( 'The SQLite Integration plugin is active, but the PDO SQLite extension is missing from your server. Please make sure that PDO SQLite is enabled in your PHP installation.', 'sqlite-database-integration' )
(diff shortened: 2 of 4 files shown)


JanJakes:
Hi @prashanthca!

This project uses the `pdo_sqlite` extensions, not `sqlite3`. These are different, independent APIs. See:
- https://www.php.net/manual/en/book.sqlite3.php
- https://www.php.net/manual/en/ref.pdo-sqlite.php

> The issue this PR is trying to solve surfaced while using a static build of PHP and PHP-FPM (v8.4.14) built using @crazywhalecc/static-php-cli. I'm using the "bulk" pre-built version of the PHP and PHP-FPM binaries, which contains pretty-much all of the essential PHP extensions built-in, including SQLite.

I just checked this and downloaded a macOS build for the "bulk" PHP 8.4 bundle version. Doing some tests:

```bash
./php -r "var_dump(extension_loaded('pdo'));" // true
./php -r "var_dump(extension_loaded('pdo_mysql'));" // true
./php -r "var_dump(extension_loaded('pdo_sqlite'));" // false
./php -r "var_dump(extension_loaded('sqlite3'));" // true
```

It would appear that the "bulk" bundle doesn't include the `pdo_sqlite` extension.

However, as you mention, I can see that this works:

```php
./php -r "var_dump((new PDO('sqlite::memory:'))->query('SELECT sqlite_version()')->fetch());"
```

It looks like the extension is included, but it's not being correctly reported. Indeed:

```php
./php -r "phpinfo();" | grep pdo_sqlite
-- no results
```

That is strange. It looks like an issue with the `static-php-cli` project. Looking closer, it seems that the `pdo_sqlite` extension is not meant to be included in the bulk bundle, as it is not listed in https://static-php-cli.fra1.digitaloceanspaces.com/static-php-cli/bulk/build-extensions.json. I don't know why it seems to work then, but I can only recommend opening an issue in the `static-php-cli` project and checking with them.

Please let me know if we can close the issue on our side.

prashanthca:
@JanJakes Thank you for the analysis and the explanation. It was my bad, not checking the difference between pdo_sqlite and sqlite3 extensions. I suppose this PR can be closed in favor of an issue on the `static-php-cli` repo. I'll look into it.

Closing this PR for now. Thanks!

JanJakes:
@prashanthca No problem, and please don't hesitate to add any updates here or reopen the issue in case it turns out that I was wrong in my research. I'm curious to learn what the root cause is.

PR #285, opened 2025-11-06
Fix `INSERT ... SELECT ...` edge case in non-strict mode

JanJakes:
In MySQL non-STRICT mode, when inserting from a `SELECT` query:

When a column is declared as `NOT NULL`, inserting a `NULL` value saves an IMPLICIT DEFAULT value instead. This behavior only applies to the `INSERT ... SELECT` syntax (not `VALUES` or `SET`).

In other words:
```sql
CREATE TABLE t (value INT NOT NULL);

-- Strict mode:
INSERT INTO t VALUES (NULL); -- error
INSERT INTO t SET value = NULL; -- error
INSERT INTO t SELECT NULL; -- error
INSERT INTO t VALUES ((SELECT NULL)); -- error

-- Non-strict mode:
INSERT INTO t VALUES (NULL); -- error
INSERT INTO t SET value = NULL; -- error
INSERT INTO t SELECT NULL; -- OK, saves 0 (the implicit default for integer)
INSERT INTO t VALUES ((SELECT NULL)); -- error
```

This fixes some failures in the activation of the top 100 plugins.

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-driver.php
@@ -4580,7 +4580,23 @@ class WP_SQLite_Driver {
 				// When a column value is included, we need to apply type casting.
 				$position   = array_search( $column['COLUMN_NAME'], $insert_list, true );
 				$identifier = $this->quote_sqlite_identifier( $select_list[ $position ] );
-				$fragment  .= $this->cast_value_for_saving( $column['DATA_TYPE'], $identifier );
+				$value      = $this->cast_value_for_saving( $column['DATA_TYPE'], $identifier );
+
+				/*
+				 * In MySQL non-STRICT mode, when inserting from a SELECT query:
+				 *
+				 * When a column is declared as NOT NULL, inserting a NULL value
+				 * saves an IMPLICIT DEFAULT value instead. This behavior only
+				 * applies to the INSERT ... SELECT syntax (not VALUES or SET).
+				 */
+				$is_insert_from_select = 'insertQueryExpression' === $node->rule_name;
+				if ( ! $is_strict_mode && $is_insert_from_select && 'NO' === $column['IS_NULLABLE'] ) {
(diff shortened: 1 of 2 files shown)


adamziel:
How are you even coming up with these nuances. 🤯

wp-includes/sqlite-ast/class-wp-sqlite-driver.php
```diff
@@ -4580,7 +4580,23 @@ function ( $column ) use ( $is_strict_mode, $insert_map ) {
...
+				 *
+				 * When a column is declared as NOT NULL, inserting a NULL value
+				 * saves an IMPLICIT DEFAULT value instead. This behavior only
+				 * applies to the INSERT ... SELECT syntax (not VALUES or SET).
+				 */
+				$is_insert_from_select = 'insertQueryExpression' === $node->rule_name;
+				if ( ! $is_strict_mode && $is_insert_from_select && 'NO' === $column['IS_NULLABLE'] ) {
+					$implicit_default = self::DATA_TYPE_IMPLICIT_DEFAULT_MAP[ $column['DATA_TYPE'] ] ?? null;
+					if ( null !== $implicit_default ) {
+						$value = sprintf( 'COALESCE(%s, %s)', $value, $this->connection->quote( $implicit_default ) );
```
adamziel, line 4596: Is that SQLite coalesce? Will it give us the result we need?

JanJakes: @adamziel Yes. There is a very similar logic for `UPDATE SET ... = NULL` already:

https://github.com/WordPress/sqlite-database-integration/blob/d982450ef97fe63d08b6d339846b165f7e270b50/wp-includes/sqlite-ast/class-wp-sqlite-driver.php#L4727-L4738

JanJakes:
> How are you even coming up with these nuances. 🤯

@adamziel Going through the few remaining errors from the top 100 plugins 🙂

PR #289, opened 2025-11-14
MySQL proxy improvements

JanJakes:
This PR further improves and refactors the original MySQL proxy prototype. Notably:
- Add relevant docs and manual links to all packet manipulation methods.
- Refactor packet building, fix packet sequence ID handling.
- Refactor packet parsing and authentication.
- Use `pack()` and `unpack()` when possible.
- Fix connecting with no password specified.
- Support specifying database name during the authentication phase (support the `MySQL_Protocol::CLIENT_CONNECT_WITH_DB` capability).
- Correctly support the `MySQL_Protocol::CLIENT_DEPRECATE_EOF` capability.
- Add MySQL CLI tests.

I've tried the proxy with the following tools, which are working now:
- MySQL CLI
- Adminer
- phpMyAdmin
- TablePlus

I think this is a good point to stop at for this PR. The refactor may go further, and there are more things to implement (prepared statements for `mysqli`, full authentication support, and more), but I'd leave those to separate issues.

I'm also considering moving the packet building methods from `MySQL_Protocol` to `MySQL_Session`, as currently, these need to pass many function parameters in some cases, while in the `MySQL_Session` class, private state properties could be used directly for some attributes.

### DIFF
--- a/packages/wp-mysql-proxy/src/class-mysql-protocol.php
+++ b/packages/wp-mysql-proxy/src/class-mysql-protocol.php
@@ -7,8 +7,16 @@ namespace WP_MySQL_Proxy;
  */
 class MySQL_Protocol {
 	/**
+	 * MySQL protocol version.
+	 *
+	 * The current version 10 is used since MySQL 3.21.0.
+	 */
+	const PROTOCOL_VERSION = 10;
+
+	/**
 	 * MySQL client capability flags.
 	 *
+	 * @see https://dev.mysql.com/doc/dev/mysql-server/latest/group__group__cs__capabilities__flags.html
 	 * @see https://github.com/mysql/mysql-server/blob/056a391cdc1af9b17b5415aee243483d1bac532d/include/mysql_com.h#L260
 	 */
 	const CLIENT_LONG_PASSWORD                  = 1 << 0;  // [NOT USED] Use improved version of old authentication.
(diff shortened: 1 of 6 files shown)


packages/wp-mysql-proxy/src/class-mysql-protocol.php
```diff
@@ -184,198 +196,439 @@ class MySQL_Protocol {
...
+	 *
+	 * @see https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_dt_strings.html
+	 *
+	 * @param  string $payload A payload of bytes to read from.
+	 * @param  int    $offset  And offset to start reading from within the payload.
+	 *                         The value will be advanced by the number of bytes read.
+	 * @return string          The decoded string value.
+	 */
+	public static function read_null_terminated_string( string $payload, int &$offset ): string {
+		$value   = unpack( 'Z*', $payload, $offset )[1];
```
adamziel, line 630: should we check all unpack calls for `false` return value ?

adamziel (approved):

packages/wp-mysql-proxy/src/class-mysql-protocol.php
JanJakes: @adamziel Ah, good point, definitely!

adamziel: This PR Already adds value as it is so I'll go ahead and merge. Let's iterate on the improvements in a follow-up.

PR #291, opened 2025-12-05
PDO API foundations

JanJakes:
This PR implements foundational PDO APIs and uses them in the existing `WP_SQLite_Driver` class.

The changes are best understood commit-by-commit, and they include the following:
1. Make `WP_PDO_MySQL_On_SQLite` extend `PDO`.
2. Implement `PDO::exec()` API.
3. Implement `PDO::query()` API.
4. Implement `PDOStatement` API for in-memory data (operating on PHP arrays) as `WP_PDO_Synthetic_Statement`. This is required for `PDO::query()` that must return a `PDOStatement` instance.
5. Implement `PDOStatement::fetch()` and `PDOStatement::fetchAll()` with the most common PDO fetch modes.
6. Implement `PDO` and `PDOStatement` `getAttribute()` and `setAttribute()` methods.
7. Implement `PDOStatement::setFetchMode()`.
8. Fix some PDO compatibility issues across all supported PHP versions.

#### Synthetic vs. proxy PDO statement
This initial work implements only a `WP_PDO_Synthetic_Statement` that requires the full PDO statement result set to be loaded in memory. This makes it easier to implement a gradual transition to the PDO API (the current driver loads all result sets in memory as well) and it can support all statement types, including those that are composed on the PHP side.

In a follow-up work, it should be possible to transition most statement types to a proxy-like PDO statement that would internally use PDO SQLite statements directly, without eagerly fetching all data. The proxy will be required to address incompatibilities between SQLite and MySQL statements.

#### Next steps
Subsequent PRs will focus on the following:
- Implement more of the `PDO` and `PDOStatement` APIs.
- Implement a `WP_PDO_Proxy_Statement` as described above.
- Simplify the SQLite driver state to only store a "last result statement" instead of the current list of properties.
- Use the PDO API directly and remove the temporary proxy to the "old" driver API.


tests/WP_PDO_MySQL_On_SQLite_PDO_API_Tests.php
```diff
@@ -0,0 +1,298 @@
...
+	/** @var WP_PDO_MySQL_On_SQLite */
+	private $driver;
+
+	public function setUp(): void {
+		$this->driver = new WP_PDO_MySQL_On_SQLite( 'mysql-on-sqlite:path=:memory:;dbname=wp;' );
+		$this->driver->setAttribute( PDO::ATTR_STRINGIFY_FETCHES, false );
+	}
+
+	public function test_connection(): void {
+		$driver = new WP_PDO_MySQL_On_SQLite( 'mysql-on-sqlite:path=:memory:;dbname=WordPress;' );
```
adamziel, line 19: nice :heart:

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -0,0 +1,6462 @@
...
+/**
+ * SQLite driver for MySQL.
+ *
+ * This class emulates a MySQL database server on top of an SQLite database.
+ * It translates queries written in MySQL SQL dialect to an SQLite SQL dialect,
+ * maintains necessary metadata, and executes the translated queries in SQLite.
+ *
+ * The driver requires PDO with the SQLite driver, and the PCRE engine.
+ */
+class WP_PDO_MySQL_On_SQLite extends PDO {
```
adamziel, line 17: Oh, so the connection driver is just this now and the other class only wraps it and translates PDO output into WordPress information. Really nice!

adamziel: Can we intentionally split this into two PRs where one would make this change directly in the `class-wp-sqlite-connection.php` and move the actual WP SQLite connection class to another file? Just so that we get to review what changes in the driver logic. Right now this file is 6,5k new lines and I'm not sure which ones overlap exactly with the previous driver and which ones were modified or added. Once such a PR is merged, we could swap the file names in a follow-up PR to end up with the useful naming scheme this PR proposes. That's a bit of GitHub hacking, but it should support us here.

JanJakes: @adamziel Good idea. Now it can be viewed in a commit-by-commit fashion:
- [Rename WP_SQLite_Driver to WP_PDO_MySQL_On_SQLite](https://github.com/WordPress/sqlite-database-integration/pull/291/changes/a9b17488302ff0030423d563af64acb8dfc78cdd)
- [Reintroduce WP_SQLite_Driver as a temporary proxy exposing legacy API](https://github.com/WordPress/sqlite-database-integration/pull/291/changes/f704b1a3d0647f34d560edc8598d59e334a99597)

But making an extra PR is even better. I already started extracting this PR into multiple smaller ones:
- https://github.com/WordPress/sqlite-database-integration/pull/294
- https://github.com/WordPress/sqlite-database-integration/pull/295
- https://github.com/WordPress/sqlite-database-integration/pull/296

So I'll prepare even more.

JanJakes: @adamziel Extracted into https://github.com/WordPress/sqlite-database-integration/pull/297.

adamziel: I'm still getting some reasonably recent search results when I google for "PHP Fatal error:  Class 'PDO' not found". We might want to polyfill the core PDO class when it's missing.

wp-includes/sqlite-ast/class-wp-pdo-proxy-statement.php
```diff
@@ -0,0 +1,342 @@
...
+ * PDO uses $class as a variable name, enable it:
+ *   phpcs:disable Universal.NamingConventions.NoReservedKeywordParameterNames.classFound
+ *
+ * We use traits to support different PHP versions with incompatible PDO statement
+ * method signatures. For that, enable multiple object structures in one file:
+ *   phpcs:disable Generic.Files.OneObjectStructurePerFile.MultipleFound
+ */
+
+/**
+ * Some PDOStatement methods are not compatible across different PHP versions.
```
adamziel, line 24: oh wow

wp-includes/sqlite-ast/class-wp-pdo-synthetic-statement.php
```diff
@@ -0,0 +1,342 @@
...
+	 */
+	public function __construct(
+		PDO $pdo,
+		array $sqlite_column_metadata,
+		array $rows,
+		int $affected_rows = 0
+	) {
+		$this->pdo                = $pdo;
+		$this->sqlite_column_meta = $sqlite_column_metadata;
+		$this->rows               = $rows;
```
adamziel: This implies we're always loading the full result set into memory. Do we have to do that? I worry we could blow up memory on some large queries.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
adamziel: It's not the blocker here by the way, we could follow up with this. I don't think we'll get any problem reports early on but let's still take this into consideration

wp-includes/sqlite-ast/class-wp-pdo-synthetic-statement.php
JanJakes: >  Do we have to do that?

@adamziel In most cases, fortunately not. But in some, when we construct the result in PHP (things like `SHOW GRANTS`, etc.), we need to have a "synthetic" statement that is implementing the PDO API on top of PHP arrays.

For all statements where a MySQL statement corresponds to an SQLite statement, it will be better to use a "proxy" statement. We can't use the SQLite statement directly because some things like column metadata formats differ, but we can use a proxy to avoid loading all results into memory. I actually have a [class-wp-pdo-proxy-statement.php‎](https://github.com/WordPress/sqlite-database-integration/pull/291/changes#diff-7e96a9ea61970a541b4d1f930592a9c7e3ab2d3f39764e8b62a190bf4b9ded23) in a WIP commit, but I'll probably extract it to a follow-up PR. Here's why I think it's fine:

In the first step, we can use the synthetic statement for everything because 1) the current driver already stores everything in memory (`$this->last_result`, etc.), and 2) `$wpdb` loads everything in memory as well. So I think I can complete this PR with the "synthetic" statement, and then in the next PR, I would replace all the `$this->last_something` props with a single `$this->last_result_statement` that we would use as the proxy target.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: > I'm still getting some reasonably recent search results when I google for "PHP Fatal error: Class 'PDO' not found". We might want to polyfill the core PDO class when it's missing.

@adamziel At the moment, we actually require PDO, because that's what we're using for the SQLite database. If a need for that ever comes up, we could consider supporting the PHP SQLite3 extension as well, although I'm afraid it provides less metadata like column info, etc.

JanJakes:
@adamziel I rebased this on all the extracted changes, and made the changeset a bit clearer and easier to understand. More details in the updated PR description.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -443,6 +443,17 @@ class WP_PDO_MySQL_On_SQLite {
...
 
+	/**
+	 * PDO API: The PDO attributes of the connection.
+	 *
+	 * TODO: Add PDO default attribute values.
+	 *
+	 * @var array<int, mixed>
+	 */
+	private $pdo_attributes = array(
+		PDO::ATTR_STRINGIFY_FETCHES => PHP_VERSION_ID < 80100 ? true : false,
```
adamziel: let's document the rationale behind this condition

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -579,24 +590,51 @@ class WP_PDO_MySQL_On_SQLite {
...
+			throw new PDOException( 'invalid data source name' );
+		}
+
+		$driver = $dsn_parts[0];
+		if ( 'mysql-on-sqlite' !== $driver ) {
+			throw new PDOException( 'could not find driver' );
+		}
+
+		$args = array();
+		foreach ( explode( ';', $dsn_parts[1] ) as $arg ) {
```
adamziel: let's scrutinize this and make double sure we're doing it in the same way as php-src does, e.g. I wonder if `mysql:dbname=testdb; host=127.0.0.1` with an extra space is also a correct dsn.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -579,24 +590,51 @@ class WP_PDO_MySQL_On_SQLite {
...
+			$arg_parts             = explode( '=', $arg, 2 );
+			$args[ $arg_parts[0] ] = $arg_parts[1] ?? null;
+		}
+
+		$path    = $args['path'] ?? ':memory:';
```
adamziel, line 654: We're exploding by `:`, so for a DSN such as `new PDO('mysql-on-sqlite::memory:')` we'll get an empty string as `$dsn_parts[1]`. Let's add a thorough test suite for parsing the DSN string, it's such a small part that's so important to get right.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -696,17 +734,100 @@ function ( string $sql, array $params ) {
...
+			}
+			if ( ! is_int( $fetch_mode_args[0] ) ) {
+				throw new TypeError(
+					sprintf( 'PDO::query(): Argument #3 must be of type int, %s given', $get_type( $fetch_mode_args[0] ) )
+				);
+			}
+			$arg_colno = $fetch_mode_args[0];
+		} elseif ( PDO::FETCH_CLASS === $fetch_mode ) {
+			if ( $arg_count < 3 ) {
+				throw new ArgumentCountError(
```
adamziel, line 815: `pdo_stmt.c` [doesn't seem to check the number of args for this mode](https://github.com/php/php-src/blob/3e87cfa9c8befa963e3a1a02cb92a479fc078df0/ext/pdo/pdo_stmt.c#L1178) – would it tolerate more? Or is it either 3 or 4 args still?

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -696,17 +734,100 @@ function ( string $sql, array $params ) {
...
+				throw new TypeError( 'PDO::query(): Argument #3 must be a valid class' );
+			}
+			if ( 4 === $arg_count && ! is_array( $fetch_mode_args[1] ) ) {
+				throw new TypeError(
+					sprintf( 'PDO::query(): Argument #4 must be of type ?array, %s given', $get_type( $fetch_mode_args[1] ) )
+				);
+			}
+			$arg_class            = $fetch_mode_args[0];
+			$arg_constructor_args = $fetch_mode_args[1] ?? array();
+		} elseif ( PDO::FETCH_INTO === $fetch_mode ) {
```
adamziel, line 839: I've stumbled upon [this weird comment in pdo_stmt.c](https://github.com/php/php-src/blob/3e87cfa9c8befa963e3a1a02cb92a479fc078df0/ext/pdo/pdo_stmt.c#L1247):

```c
 		default:
			/* No support for PDO_FETCH_INTO which takes 2 args??? */
```

I'm not sure if we need to do anything with it. Does PHP actually support FETCH_INTO? 😅

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -696,17 +734,100 @@ function ( string $sql, array $params ) {
...
+			}
+			return $type;
+		};
+
+		if ( null === $fetch_mode ) {
+			// When the default FETCH_BOTH is not set explicitly, additional
+			// arguments are ignored, and the argument count is not validated.
+			$fetch_mode = PDO::FETCH_BOTH;
+		} elseif ( PDO::FETCH_COLUMN === $fetch_mode ) {
+			if ( 3 !== $arg_count ) {
```
adamziel, line 802: Is it 3 because of the extra `query` arg at the beginning? I've noticed `fetchAll()` [expects up to 2](https://github.com/php/php-src/blob/3e87cfa9c8befa963e3a1a02cb92a479fc078df0/ext/pdo/pdo_stmt.c#L1222C1-L1228C5):

```c
			if (ZEND_NUM_ARGS() > 2) {
				zend_string *func = get_active_function_or_method_name();
				zend_argument_count_error("%s() expects at most 2 argument for the fetch mode provided, %d given",
					ZSTR_VAL(func), ZEND_NUM_ARGS());
				zend_string_release(func);
				RETURN_THROWS();
			}
```

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -696,17 +734,100 @@ function ( string $sql, array $params ) {
...
+				return 'bool';
+			} elseif ( 'integer' === $type ) {
+				return 'int';
+			} elseif ( 'double' === $type ) {
+				return 'float';
+			}
+			return $type;
+		};
+
+		if ( null === $fetch_mode ) {
```
adamziel, line 788: There's also `PDO::FETCH_DEFAULT` that [pdo_stmt.c seems to rely on](https://github.com/php/php-src/blob/3e87cfa9c8befa963e3a1a02cb92a479fc078df0/ext/pdo/pdo_stmt.c#L1257-L1261)

adamziel: Also [PDO::FETCH_KEY_PAIR](https://github.com/php/php-src/blob/3e87cfa9c8befa963e3a1a02cb92a479fc078df0/ext/pdo/pdo_stmt.c#L1270C20-L1270C38). Let's identify all of those and at least document the unsupported ones (and throw an explicit error when they're used). Ditto for flags.

wp-includes/sqlite-ast/class-wp-pdo-proxy-statement.php
```diff
@@ -0,0 +1,502 @@
...
+	 * @param  int|null $cursorOrientation The cursor orientation. Controls which row is returned.
+	 *                                     Default: PDO::FETCH_ORI_NEXT (null for PHP < 8.0)
+	 * @param  int|null $cursorOffset      The cursor offset. Controls which row is returned.
+	 *                                     Default: 0 (null for PHP < 8.0)
+	 * @return mixed                       The row data formatted according to the fetch mode;
+	 *                                     false if there are no more rows or a failure occurs.
+	 */
+	#[ReturnTypeWillChange]
+	public function fetch(
+		$mode = 0, // PDO::FETCH_DEFAULT (available from PHP 8.0)
```
adamziel, line 184: Ah, that's why we don't use that constant by name in the other file. Maybe we can refer to our own set of constants that's always consistently defined across all PHP versions? That way we won't need to wonder whether a specific `0` has a special constant-related meaning.

wp-includes/sqlite-ast/class-wp-pdo-synthetic-statement.php
```diff
@@ -0,0 +1,502 @@
...
+		$cursorOrientation = 0,
+		$cursorOffset = 0
+	) {
+		if ( 0 === $mode || null === $mode ) {
+			$mode = $this->fetch_mode;
+		}
+		if ( null === $cursorOrientation ) {
+			$cursorOrientation = PDO::FETCH_ORI_NEXT;
+		}
+		if ( null === $cursorOffset ) {
```
adamziel: This will only be null if someone explicitly overrides the default `0` value. It would read slightly more natural if the signature said `$cursorOffset = null` and then we'd correct nulls to `0`s here. I suppose we can't do that because of the parent method signature?

wp-includes/sqlite-ast/class-wp-pdo-synthetic-statement.php
```diff
@@ -0,0 +1,502 @@
...
+
+		/*
+		 * TODO: Support scrollable cursor ($cursorOrientation and $cursorOffset).
+		 *       This only has works for with statements that were prepared with
+		 *       the PDO::ATTR_CURSOR attribute set to PDO::CURSOR_SCROLL value.
+		 *       Without it, these parameters have no effect.
+		 */
+
+		/**
+		 * With PHP < 8.1, the "PDO::ATTR_STRINGIFY_FETCHES" value of "false"
```
adamziel: Great find!

wp-includes/sqlite-ast/class-wp-pdo-synthetic-statement.php
```diff
@@ -0,0 +1,502 @@
...
+						$values[ $name ] = $value;
+					}
+				}
+				return $values;
+			case PDO::FETCH_OBJ:
+				return (object) array_combine( $column_names, $row );
+			case PDO::FETCH_CLASS:
+				throw new RuntimeException( "'PDO::FETCH_CLASS' mode is not supported" );
+			case PDO::FETCH_INTO:
+				throw new RuntimeException( "'PDO::FETCH_INTO' mode is not supported" );
```
adamziel: Ah, there we go throwing for unsupported modes. That takes care of my other comment.

wp-includes/sqlite-ast/class-wp-pdo-synthetic-statement.php
adamziel: > (things like SHOW GRANTS, etc.),

I don't suppose we would ever see a 100 megabyte result set in those so that seems fine. 

> In the first step, we can use the synthetic statement for everything because 1) the current driver already stores everything in memory ($this->last_result, etc.), and 2) $wpdb loads everything in memory as well. So I think I can complete this PR with the "synthetic" statement, and then in the next PR, I would replace all the $this->last_something props with a single $this->last_result_statement that we would use as the proxy target.

That sounds good. For Playground, the database is in memory anyway so it's not a big deal. It will matter when a larger site wants to adapt this driver, and perhaps we're reasonably close to having a technical ability to do it with all the MySQL queries we support. Not a huge priority, but it would be nice to briefly document in a docstring that yes, this is possible and here's what needs to happen technically.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel It's actually `explode( ':', $dsn, 2 )`, which means it will only split by the first `:` (the `driver:` prefix). The second explode is by `;`, so I think we're safe in this case.

JanJakes: Digging deeper, I can confirm searching for the first `:` is what PDO does as well: https://github.com/php/php-src/blob/cdd5aa21d8ca3d774447c6ff02ba6518ce2aece7/ext/pdo/pdo_dbh.c#L328

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel Good point. This one is, indeed, a valid DSN. I checked what PDO does, and this is the only place where they allow (= ignore) the whitespace: https://github.com/php/php-src/blob/cdd5aa21d8ca3d774447c6ff02ba6518ce2aece7/ext/pdo/pdo.c#L247-L249

Apart from that, they only seem to implement semicolon quoting by `;;` sequences: https://github.com/php/php-src/blob/cdd5aa21d8ca3d774447c6ff02ba6518ce2aece7/ext/pdo/pdo.c#L196-L198

So I'm adding support and tests for both of these.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel Seems to be 3 or 4 still:

```php
php -a
Interactive shell

php > $pdo = new PDO( 'mysql:dbname=wordpress;host=127.0.0.1;port=8306', 'wordpress', 'wordpress' );
php > $pdo->query( 'SELECT 1', PDO::FETCH_CLASS, 'stdClass', array(), 1 );
PHP Warning:  Uncaught ArgumentCountError: PDO::query() expects at most 4 arguments for the fetch mode provided, 5 given in php shell code:1
Stack trace:
#0 php shell code(1): PDO->query('SELECT 1', 8, 'stdClass', Array, 1)
#1 {main}
  thrown in php shell code on line 1

Warning: Uncaught ArgumentCountError: PDO::query() expects at most 4 arguments for the fetch mode provided, 5 given in php shell code:1
Stack trace:
#0 php shell code(1): PDO->query('SELECT 1', 8, 'stdClass', Array, 1)
#1 {main}
  thrown in php shell code on line 1
```

Although now I too wonder where  this is checked in the PDO code 🤔

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel Haha, it seems to support it:

```php
php -a
Interactive shell

php > $pdo = new PDO( 'mysql:dbname=wordpress;host=127.0.0.1;port=8306', 'wordpress', 'wordpress' );
php > $out = new class() { public $test = null; };
php > $pdo->query( 'SELECT 1 AS test', PDO::FETCH_INTO, $out )->fetch();
php > var_dump( $out );
object(class@anonymous)#2 (1) {
  ["test"]=>
  int(1)
}
```

But in our case, it's not implemented in this PR.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel Yes, it seems so. In this case, the arguments are 1) query, 2) fetch mode, 3) the column number; all required.

wp-includes/sqlite-ast/class-wp-pdo-proxy-statement.php
JanJakes: @adamziel Actually, `PDO::FETCH_DEFAULT` seems to be the only one that's available from PHP 8.0 only. All the other constants are available in the whole range of PHP versions that we support. In that case, I think it's not worth it creating our own set for just that one constant.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel Oh, I missed some, it seems (`PDO::FETCH_KEY_PAIR`). Listing them all in a comment just to be sure.

However, the number and complexity of all the fetch mode variations make me wonder...

**Do we need to reimplement them all?** What if we just **use PDO SQLite statements in all cases**, even when we have a fixed dataset in memory? Something like:

```php
$stmt = $sqlite_pdo->query( 'SELECT <col-names> FROM VALUES <my-fixed-dataset>' );

// Now the SQLite statement should support all the fetch modes and stuff out of the box.
// We may still need to proxy it for other reasons, but it could save us a lot of emulation.
$stmt->fetchAll( ... );
```

Maybe I should give this approach a quick try.

adamziel: That sounds really good

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
adamziel: Btw I had no idea you could do this in PHP

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
adamziel: Oh, good!

wp-includes/sqlite-ast/class-wp-pdo-proxy-statement.php
adamziel: Makes sense. Let’s make sure to document it and we’re good

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel 

Works like a charm: https://github.com/WordPress/sqlite-database-integration/pull/291/changes/ba5b0cece2321c91547b23922b58e97e2d7cd7eb

Next step: Use actual SQLite PDOStatement everywhere we can and leave this dummy select just for the in-memory case when we have no SQLite statement at hand. (But it is a good confidence check that the tests are passing when _all_ data go through the dummy select.)

This is great news—we can always proxy the statement methods to an SQLite PDOStatement and only modify or emulate what will be needed to be modified or emulated. PDO does most of the heavy lifting here for us.

wp-includes/sqlite-ast/class-wp-pdo-synthetic-statement.php
JanJakes: @adamziel The current newer version leaves this to PDO and proxies set/get attributes to the SQLite PDO.

And just for the record, this was actually inaccurate 😬 It turns out that:
1. `PDO::ATTR_STRINGIFY_FETCHES` always defaults to `false` on all PHP versions.
2. On PHP < 8.1, it's buggy. That is, when `PDO::ATTR_EMULATE_PREPARES` is `true` (which is the default), the above option has no effect.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: And the next step wasn't that hard either: https://github.com/WordPress/sqlite-database-integration/pull/291/changes/f7248a356ba9e258c824d4655d9e3fba8cb6bae4

We still do load all data in the legacy temporary wrapper class and will need to continue doing so in WPDB, but that's another story.

wp-includes/sqlite-ast/class-wp-pdo-synthetic-statement.php
JanJakes: Resolved in https://github.com/WordPress/sqlite-database-integration/pull/291/changes/f7248a356ba9e258c824d4655d9e3fba8cb6bae4, see also https://github.com/WordPress/sqlite-database-integration/pull/291#discussion_r2698644574

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: Same same. PDO has many interesting modes, some of which could bring good perf to ORMs (but as I know Doctrine, they don't use any of these).

JanJakes:
@adamziel I think this one is ready for a final round. In the end, it only needs a `WP_PDO_Proxy_Statement` that always wraps a real or a "dummy select" PDO SQLite statement.

I left many methods with `throw new RuntimeException( 'Not implemented' );`. This is on purpose—I know that just proxying to the underlying statement can implement most of them, but I want to add that with a test coverage and in a follow-up PR so that it's easier to review.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
adamziel: I really like where this is going.

tests/WP_PDO_MySQL_On_SQLite_PDO_API_Tests.php
```diff
@@ -7,8 +7,222 @@ class WP_PDO_MySQL_On_SQLite_PDO_API_Tests extends TestCase {
 	private $driver;
 
 	public function setUp(): void {
-		$connection   = new WP_SQLite_Connection( array( 'path' => ':memory:' ) );
-		$this->driver = new WP_PDO_MySQL_On_SQLite( $connection, 'wp' );
+		$this->driver = new WP_PDO_MySQL_On_SQLite( 'mysql-on-sqlite:path=:memory:;dbname=wp;' );
```
adamziel, line 10: I just realized PHP probably doesn't support paths such as `./mydb/one,two,three.sqlite` (and so doesn't this implementation, which is absolutely fine).

tests/WP_PDO_MySQL_On_SQLite_PDO_API_Tests.php
```diff
@@ -7,8 +7,222 @@ class WP_PDO_MySQL_On_SQLite_PDO_API_Tests extends TestCase {
...
+
+		// Run all tests with stringified fetch mode results, so we can use
+		// assertions that are consistent across all tested PHP versions.
+		// The "PDO::ATTR_STRINGIFY_FETCHES" mode is tested in separately.
```
adamziel: nitpick

```suggestion
		// The "PDO::ATTR_STRINGIFY_FETCHES" mode is tested separately.
```

tests/WP_PDO_MySQL_On_SQLite_PDO_API_Tests.php
```diff
@@ -7,8 +7,222 @@ class WP_PDO_MySQL_On_SQLite_PDO_API_Tests extends TestCase {
...
+		$this->driver->setAttribute( PDO::ATTR_STRINGIFY_FETCHES, true );
+	}
+
+	public function test_connection(): void {
+		$driver = new WP_PDO_MySQL_On_SQLite( 'mysql-on-sqlite:path=:memory:;dbname=WordPress;' );
+		$this->assertInstanceOf( PDO::class, $driver );
+	}
+
+	public function test_dsn_parsing(): void {
```
adamziel, line 23: thank you for these tests!

tests/WP_PDO_MySQL_On_SQLite_PDO_API_Tests.php
```diff
@@ -7,8 +7,222 @@ class WP_PDO_MySQL_On_SQLite_PDO_API_Tests extends TestCase {
...
+
+	public function test_query_fetch_mode_invalid_arg_count(): void {
+		$this->expectException( ArgumentCountError::class );
+		$this->expectExceptionMessage( 'PDO::query() expects exactly 2 arguments for the fetch mode provided, 3 given' );
+		$this->driver->query( 'SELECT 1', PDO::FETCH_ASSOC, 0 );
+	}
+
+	public function test_query_fetch_default_mode_allow_any_args(): void {
+		if ( PHP_VERSION_ID < 80100 ) {
+			// On PHP < 8.1, fetch mode value of NULL is not allowed.
```
adamziel, line 134: til

tests/WP_PDO_MySQL_On_SQLite_PDO_API_Tests.php
```diff
@@ -7,8 +7,222 @@ class WP_PDO_MySQL_On_SQLite_PDO_API_Tests extends TestCase {
...
+	}
+
+	public function test_query_fetch_into_invalid_object_type(): void {
+		$this->expectException( TypeError::class );
+		$this->expectExceptionMessage( 'PDO::query(): Argument #3 must be of type object, int given' );
+		$this->driver->query( 'SELECT 1', PDO::FETCH_INTO, 1 );
+	}
+
+	public function test_exec(): void {
+		$result = $this->driver->exec( 'SELECT 1' );
```
adamziel, line 209: It's really cool how easy to use this is becoming.

tests/WP_PDO_MySQL_On_SQLite_PDO_API_Tests.php
```diff
@@ -50,4 +264,162 @@ public function test_rollback_no_active_transaction(): void {
...
+		$result = $this->driver->query( "SELECT 123, 1.23, 'abc', true, false" );
+		$this->assertSame(
+			/*
+			 * On PHP < 8.1, "PDO::ATTR_STRINGIFY_FETCHES" set to "false" has no
+			 * effect when "PDO::ATTR_EMULATE_PREPARES" is "true" (the default).
+			 *
+			 * TODO: Consider supporting non-string values on PHP < 8.1 when both
+			 *       "PDO::ATTR_STRINGIFY_FETCHES" and "PDO::ATTR_EMULATE_PREPARES"
+			 *       are set to "false". This would require emulating the behavior,
+			 *       as PDO SQLite on PHP < 8.1 seems to always return strings.
```
adamziel, line 365: Would that be a real problem? It seems to only affect programs that perform this exact PHP_VERSION check to react differently to PDO outputs in different PHP versions.

tests/WP_PDO_MySQL_On_SQLite_PDO_API_Tests.php
```diff
@@ -50,4 +264,162 @@ public function test_rollback_no_active_transaction(): void {
...
+			 */
+			PHP_VERSION_ID < 80100
+				? array( '123', '1.23', 'abc', '1', '0' )
+				: array( 123, 1.23, 'abc', 1, 0 ),
+			$result->fetch( PDO::FETCH_NUM )
+		);
+	}
+
+	public function data_pdo_fetch_methods(): Generator {
```
adamziel, line 374: oh! you can use a Generator as a data provider. This. Is. Amazing. Hey, @dmsnell @sirreal, I think Jan has found a solution to all our PHPCS problems in large data providers.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -579,24 +583,63 @@ class WP_PDO_MySQL_On_SQLite {
...
+		if ( count( $dsn_parts ) < 2 ) {
+			throw new PDOException( 'invalid data source name' );
+		}
+
+		$driver = $dsn_parts[0];
+		if ( 'mysql-on-sqlite' !== $driver ) {
+			throw new PDOException( 'could not find driver' );
+		}
+
+		// PDO DSN supports semicolon quoting using double semicolon sequences.
```
adamziel: did you mean semicolon escaping?

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -579,24 +583,63 @@ class WP_PDO_MySQL_On_SQLite {
...
+		// Replace ";;" with "\0" to preserve quoted semicolons in "explode()".
```
adamziel: This would conflate double semicolons with legitimate `\0`-s in the data. Replacing explode with strspn/strcspn would take more effort but yield more reliable results in those corner cases.

Also, is `;;;` treated as `;<end of data>` or `<end of data>;`?

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -579,24 +583,63 @@ class WP_PDO_MySQL_On_SQLite {
...
+		$args_string = str_replace( ';;', "\0", $dsn_parts[1] );
+		$args        = array();
+		foreach ( explode( ';', $args_string ) as $arg ) {
+			// Restore quoted semicolons that were replaced with "\0".
+			$arg = str_replace( "\0", ';', $arg );
+
+			// PDO DSN allows whitespace before argument name.
```
adamziel: any specific whitespace characters? do they overlap with what `ltrim()` removes, or do we need to list them all explicitly?

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -579,24 +583,63 @@ class WP_PDO_MySQL_On_SQLite {
...
+			$arg_parts             = explode( '=', $arg, 2 );
+			$args[ $arg_parts[0] ] = $arg_parts[1] ?? null;
+		}
+
+		$path    = $args['path'] ?? ':memory:';
+		$db_name = $args['dbname'] ?? 'sqlite_database';
+
+		// Create a new SQLite connection.
+		if ( isset( $options['pdo'] ) ) {
+			$this->connection = new WP_SQLite_Connection( array( 'pdo' => $options['pdo'] ) );
```
adamziel, line 659: Hm, so we can pass both the `dsn` and `$options['pdo']`? And the dsn is still validated and parsed in that case? Shouldn't we throw an error when both are passed here?

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -696,15 +739,98 @@ function ( string $sql, array $params ) {
...
+				return 'bool';
+			} elseif ( 'integer' === $type ) {
+				return 'int';
+			} elseif ( 'double' === $type ) {
+				return 'float';
+			}
+			return $type;
+		};
+
+		if ( null === $fetch_mode ) {
```
adamziel, line 788: Sooo, now that we delegate the logic to PDO, do we need to validate all of that here? Or could we also rely on PDO validation?

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -4210,10 +4341,21 @@ private function translate_function_call( WP_Parser_Node $node ): string {
...
-					$found_rows = count( $this->last_result );
-				}
-				return $found_rows;
+
+				/*
+				 * TODO: Handle case when "null === $found_rows".
+				 *
+				 * From MySQL documentation:
+				 *
+				 * In the absence of the SQL_CALC_FOUND_ROWS option in the most
```
adamziel: I hope noone out there actually relies on that behavior, but some plugins probably do. Obligatory xkcd: xkcd

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -6289,41 +6431,90 @@ private function quote_mysql_utf8_string_literal( string $utf8_literal ): string
...
+		 *   UNION ALL
+		 *
+		 *   -- The actual data rows.
+		 *   VALUES
+		 *     (val11, val12, ...),
+		 *     (val21, val22, ...),
+		 *     ...
+		 */
+
+		// Construct column header row ("SELECT <column-list> WHERE FALSE").
```
adamziel, line 6567: very clever!

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -6289,41 +6431,90 @@ private function quote_mysql_utf8_string_literal( string $utf8_literal ): string
...
-	private function set_result_from_affected_rows( ?int $override = null ): void {
+	private function create_result_statement_from_data( array $columns, array $rows ): PDOStatement {
+		$pdo = $this->connection->get_pdo();
+
 		/*
-		 * SELECT CHANGES() is a workaround for the fact that $stmt->rowCount()
-		 * returns "0" (zero) with the SQLite driver at all times.
-		 * See: https://www.php.net/manual/en/pdostatement.rowcount.php
+		 * With 0 columns, we need to create a PDO statement that has no columns.
+		 * This can be done using a noop INSERT statement that modifies no data.
```
adamziel, line 6540: wow, clever

adamziel (approved):
I only have nitpicks. Really good work here, it's exciting to see how this is coming together. If you asked me a few years ago, I would have never guessed we'd be doing anything like this and yet here we are :heart:

tests/WP_PDO_MySQL_On_SQLite_PDO_API_Tests.php
JanJakes: @adamziel This works:

```
new WP_PDO_MySQL_On_SQLite( 'mysql-on-sqlite:path=./mydb/one,two,three.sqlite;dbname=WordPress;' );
```

The only thing is that the `mydb` directory needs to exist, as it won't be automatically created.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: > This would conflate double semicolons with legitimate \0-s in the data.

Oh, I should've documented this better together with some test coverage. PDO itself doesn't support `\0` in the DSN string, and it will end parsing at the first `\0`. Fixing in https://github.com/WordPress/sqlite-database-integration/pull/291/changes/b6368247cee806189a291cb189ca7bf2d8547934.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel Unfortunately, not here. We delegate the `PDOStatement` API to SQLite PDO, but not the `PDO::query()`, where we need more granular control over what queries and how many (if any) we execute.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: Good point, let's list them explicitly: https://github.com/WordPress/sqlite-database-integration/pull/291/changes/3f9c9673a76b842c49e5d08d89d0c7dc147157ad

They are derived from the `isspace` C function: https://github.com/php/php-src/blob/a3576bddc52310e8ed6c545e38104e884713ceef/ext/pdo/pdo.c#L247

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel True true 😬😂 I took a stab at it here: https://github.com/WordPress/sqlite-database-integration/pull/291/changes/3c8543bc7a82966c95020d1c052c4720ddac4f16

I think that implementation does the > 99% case. 

It's not 100%, because calling `FOUND_ROWS()` multiple times in a row will not necessarily give correct results. And then I also discovered this:

I also found this strange nuance:
Screenshot 2026-01-19 at 11 24 55

But we never supported these edge cases anyway.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel This allows passing the underlying PDO SQLite instance, not the one we're parsing the options for.

While in theory, it can make some sense to support it (`new WP_PDO_MySQL_On_SQLite( 'mysql-on-sqlite:dbname=test', [ 'pdo' => $my_pdo ] )`), I'm not sure if we should keep it, and if we do, a better name for that option would probably make sense.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel Yeah, fixing the comments. Thanks!

tests/WP_PDO_MySQL_On_SQLite_PDO_API_Tests.php
JanJakes: Yeah, it's pretty cool. There was an older P2 post about this. I'll try to find it and send it to you.

tests/WP_PDO_MySQL_On_SQLite_PDO_API_Tests.php
JanJakes: @adamziel Oh, that's a good point! Realistically, we may not run into this at all.

It's a potential problem in case someone wants to use the new PDO API directly and force it to return typed values across all PHP versions. In native PDO, the trick to do that would be to set both of these attributes to `false`, but with our driver, it would return typed values only from PHP 8.1. I hope we'll bump to PHP 8.1 sooner than we run into this 🙂

JanJakes:
@adamziel Thanks for the great review, as always. Another round should be ready now ✅

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
adamziel: Oh wow. Good find!

tests/WP_PDO_MySQL_On_SQLite_PDO_API_Tests.php
adamziel: Oh, sorry, I've meant `./mydb/one;two;three.sqlite` which is a valid path but it also contains the `;` separator.

adamziel: Aha, we'd probably have to write it as `one;;two;;three.sqlite`

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -2875,11 +2896,16 @@ private function execute_show_index_statement( WP_Parser_Node $node ): void {
 					ROWID,
 					SEQ_IN_INDEX
 			",
-			array( $this->get_saved_db_name( $database ), $table_name )
+			$this->quote_sqlite_identifier( $statistics_table ),
+			$this->connection->quote( $this->get_saved_db_name( $database ) ),
```
adamziel: interesting, why do we need two quoting functions?

adamziel: aha, one is for identifiers and another is for string values, and also this is no longer a prepared statement? Why ditch prepared statements here?

adamziel: This would communicate the data flow right away:

```php
$this->quote_sqlite_identifier( $statistics_table ),
$this->quote_sqlite_value( $this->get_saved_db_name( $database ) );
```

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
adamziel: Aha, lovely!

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel I guess it evolved into this state over time, and it's a bit confusing indeed:
- Identifier: Private method on the driver class.
- Value: Public method on the connection class.

What you suggest is much clearer 👍, and it has additional advantages:
- It will clearly differentiate the methods from `WP_PDO_MySQL_On_SQLite::quote()`, which is required by the PDO API, but it's intended to be MySQL-like quoting, not SQLite.
- We may consider getting rid of `WP_SQLite_Connection` in the future. I'm not sure if such a small wrapper around the PDO SQLite connection is necessary, probably not anymore.

So I can add a commit with these changes 👍

JanJakes: > Why ditch prepared statements here?

@adamziel This was a result of a quick attempt to solve the `FOUND_ROWS()` without `SQL_CALC_FOUND_ROWS` functionality. I needed to store the last SQLite query that returns some rows in order to be able to wrap it with `SELECT COUNT ...` in case `FOUND_ROWS()` is used. For `SELECT` it was straightforward, but then I realized we also need `SHOW` and `DESCRIBE` and the quick attempt was to make all the queries just strings that I can then use in a subquery.

That said, it's not strictly necessary to ditch prepared statements—I could simply store both the query string and the parameters. Do you think I should revert those changes and do it this way instead?

adamziel: I'm always nervous about formatting strings instead of using prepared statements, considering how different encodings represent bytes. We may have a watertight implementation here, but it still stands out to me as a risky spot in the system. I wouldn't say it's blocking this PR, but if it's an easy change I'd go for it.

JanJakes: @adamziel I already did (last two commits), but forgot to note it here.

The test failures are unrelated, I hotfixed that in https://github.com/WordPress/sqlite-database-integration/pull/306.

adamziel:
Lovely work @JanJakes! This one looks great to me, feel free to merge anytime.

PR #296, opened 2025-12-18
Fix incorrect column names for string literals without alias

JanJakes:
It turns out that when selecting a string literal, the resulting column name differs between MySQL and SQLite.

For example, with a query such as `SELECT 'abc'`:
- In MySQL, the returned column name is `abc` (the unquoted string value).
- In SQLite, the returned column name is `'abc'` (the quoted string).

To address that, we need to use "SELECT 'abc' AS \`abc\`" in these cases.


adamziel:
Oh wow, this is such a good find.

PR #297, opened 2025-12-18
Bootstrap for PDO API implementation

JanJakes:
Extracting smaller chunks from https://github.com/WordPress/sqlite-database-integration/pull/291 so we can review and merge them independently.

This PR renames the current `WP_SQLite_Driver` to `WP_PDO_MySQL_On_SQLite` and reintroduces the `WP_SQLite_Driver` as a simple proxy over the new renamed class. It's done in these two steps (first rename, then reintroduce) so that Git understands the rename and presents history (hopefully) accurately.

The changes are better understood in a commit-by-commit view.

#### `WP_PDO_MySQL_On_SQLite` vs `WP_SQLite_Driver`
The "reintroduced" `WP_SQLite_Driver` is not meant to be permanent. It is a temporary proxy so we can gradually modify `WP_PDO_MySQL_On_SQLite` to support PDO APIs while not touching the driver API just yet. Once the basics of PDO API are in place, we can make all dependencies use the new class directly and then remove the `WP_SQLite_Driver`.

That is, in the future, the `WP_SQLite_DB extends wpdb` will use directly the new `WP_PDO_MySQL_On_SQLite` class.

### DIFF
--- a/packages/wp-mysql-proxy/src/Adapter/class-sqlite-adapter.php
+++ b/packages/wp-mysql-proxy/src/Adapter/class-sqlite-adapter.php
@@ -9,24 +9,7 @@ use WP_SQLite_Connection;
 use WP_SQLite_Driver;
 use WP_MySQL_Proxy\MySQL_Protocol;
 
-define( 'SQLITE_DRIVER_PATH', __DIR__ . '/../../../..' );
-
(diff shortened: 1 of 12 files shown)


adamziel (approved):
reviewing this commit by commit it seems like a fairly small PR despite what the files changed tab says. Looking good!

PR #302, opened 2026-01-07
Improve compatibility with legacy SQLite versions

JanJakes:
## SQLite Version Compatibility Fixes

This PR adds support for older SQLite versions by addressing various compatibility issues. It also updates the CI workflow to **test against multiple SQLite versions**.

**Compatibility fixes:**

1. **UPSERT / ON CONFLICT clause (SQLite < 3.35.0)** — The generic `ON CONFLICT DO UPDATE` clause without an explicit column list is only supported from SQLite 3.35.0. For older versions, we now catch the constraint violation error, parse the conflicting column names from the error message, and re-execute with an explicit column list.

2. **STRICT tables (SQLite < 3.37.0)** — The `STRICT` keyword for table creation is only supported from SQLite 3.37.0. It is now conditionally omitted on older versions.

3. **VALUES list `columnN` naming (SQLite < 3.33.0)** — Automatic column naming for VALUES lists (`column1`, `column2`, etc.) is only supported from SQLite 3.33.0. For older versions, we now emulate this by prepending a dummy `SELECT NULL AS column1, ... WHERE FALSE UNION ALL ...` header.

4. **`UPDATE ... FROM` syntax (SQLite < 3.33.0)** — The `UPDATE ... FROM` syntax is only supported from SQLite 3.33.0. We now avoid using it internally in the information schema builder.

5. **`IIF()` function (all versions)** — The `IIF()` function was added in SQLite 3.32.0. We now use `CASE WHEN ... THEN ... ELSE ... END` instead for broader compatibility.

6. **`SUBSTRING()` function (all versions)** — The `SUBSTRING()` function is not supported in older SQLite versions. We now translate it to `SUBSTR()`.

7. **`sqlite_schema` table (SQLite < 3.33.0)** — The `sqlite_schema` table alias was added in SQLite 3.33.0. We now use the older `sqlite_master` name for compatibility.

8. **SQLite version detection** — Changed from using a `SELECT SQLITE_VERSION()` query to using the `PDO::ATTR_SERVER_VERSION` attribute for retrieving the SQLite version, which avoids extra queries by using SQLite C API directly.

9. **CI workflow** — Added support for testing against multiple SQLite versions.


wojtekn:
I didn't test each case individually, but I tested the changes on the Studio and Studio preview sites. It worked well in all cases.

Code-wise looks good.

It's great that we are adding support for older SQLite versions, as it will improve support on some LTS images.

wojtekn (approved):

wp-includes/sqlite-ast/class-wp-sqlite-information-schema-builder.php
```diff
@@ -1058,18 +1058,17 @@ private function record_drop_column(
...
 			sprintf(
-				'UPDATE %s AS statistics
-				SET seq_in_index = renumbered.seq_in_index
-				FROM (
+				'WITH renumbered AS (
 					SELECT
 						rowid,
 						row_number() OVER (PARTITION BY index_name ORDER BY seq_in_index) AS seq_in_index
-					FROM %s
+					FROM %s x
```
JanJakes: Oh, this `x` is a leftover from a quick fix when I was probably nagged by SQLite to use an alias. I'll rename it to something more reasonable, in case it is required.

PR #304, opened 2026-01-08
Bump "Tested up to" plugin header

JanJakes:
This is to fix the following warning:

> This plugin **hasn’t been tested with the latest 3 major releases of WordPress**. It may no longer be maintained or supported and may have compatibility issues when used with more recent versions of WordPress.

Screenshot 2026-01-08 at 17 29 18


JanJakes:
Merging this one as there is nothing to review or test.

PR #306, opened 2026-01-15
Add support for `DEFAULT (expression)` in column definitions

### LINKED ISSUE #300: Create table with `DEFAULT (now())` fails in AST driver
Here is the SQL:

```
CREATE TABLE `test_now_default` (
  `id` int NOT NULL,
  `updated` timestamp NOT NULL DEFAULT (now()) ON UPDATE CURRENT_TIMESTAMP
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_520_ci;
```

It should work in MySQL 8.0.

wojtekn:
## Summary
Adds full support for `DEFAULT (expression)` in `CREATE TABLE` statement column definitions.

Fixes #300 - Support for `DEFAULT (now())`.

## Problem

The MySQL 8.0 syntax `DEFAULT (function())` was failing because the translator only consumed a single token after the DEFAULT keyword. For expressions like `DEFAULT (now())`, it would only capture the opening parenthesis `(`, causing the query translation to fail.

## Solution

1. Full `DEFAULT (expression)` support in the AST driver.

2. Enhanced the `DEFAULT` clause handler in `WP_SQLite_Translator` to:
- Detect when a DEFAULT value starts with an opening parenthesis
- Track parenthesis depth to consume the complete expression
- Handle nested parentheses correctly


JanJakes:
@wojtekn Thank you for looking into this! Good job working with the AST and even fixing the old driver 👍 

Generally, it looks good to me, and it is the right approach to work with the new driver. On the AST side, I spotted one issue that is related to only implementing a subset of the expression functionality, and that is that for expressions such as this one:

```sql
DEFAULT (DATE_ADD(NOW(), INTERVAL 1 YEAR))
```

It would detect and incorrectly return the value of `NOW()`. That's because the implementation checks whether there is that function call, but it can't easily determine that there is nothing else. Fixing that is not straightforward with arbitrary nesting, e.g. for `DEFAULT (((now())))`, and doing the full expression support is easy, so I think we could even go all the way. Let me see if I can push my idea somehow. I probably can't modify this branch as it's from a fork, but I can push another one.

Additionally, in the MySQL column metadata, there needs to be an indication that it has a `DEFAULT_GENERATED` value, but I can see that [we already seem to be doing that](https://github.com/WordPress/sqlite-database-integration/blob/5a5c7b99dbc97a4f31c8169d74301a34ba5f6f37/wp-includes/sqlite-ast/class-wp-sqlite-information-schema-builder.php#L2194), so I think we're safe on this one. Perhaps it needs a test in case there isn't one yet.

JanJakes:
@wojtekn It turns out I can push to this branch. I just have no idea how to manage PRs from forks 😆

I added one commit with a quick implementation of the full `DEFAULT (expression)` on the AST side. I will check it again after the weekend and see if it needs some polishing and/or more tests.

TODO:
- [x] Check if we need to test more expressions.
- [x] Add a test for metadata (`DEFAULT_GENERATED `).
- [x] Check if we can improve stored expression formatting.

Thanks again for pushing this forward!

JanJakes:
No more feedback here, so let's get it in.

wojtekn:
@JanJakes, thanks for taking it over and finishing it.

PR #308, opened 2026-01-16
Add support for `DEFAULT (expression)` in column definitions

### LINKED ISSUE #300: Create table with `DEFAULT (now())` fails in AST driver

JanJakes:
Add full support for `DEFAULT (expression)` in column definitions.

This PR builds on https://github.com/WordPress/sqlite-database-integration/pull/306 and expands it to add full support for expressions in `DEFAULT` column values.

Fixes https://github.com/WordPress/sqlite-database-integration/issues/300.
Closes https://github.com/WordPress/sqlite-database-integration/pull/306.

### DIFF
--- a/tests/WP_SQLite_Driver_Tests.php
+++ b/tests/WP_SQLite_Driver_Tests.php
@@ -11261,5 +11261,58 @@ END;
 
 		// Verify the updated timestamp was set (should match YYYY-MM-DD HH:MM:SS format)
 		$this->assertRegExp( '/\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d/', $result[0]->updated );
+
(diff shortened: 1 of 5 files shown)


JanJakes:
It appears that I actually can edit https://github.com/WordPress/sqlite-database-integration/pull/306, so I'll continue there.

PR #309, opened 2026-01-19
Add support for `SHOW FULL COLUMNS`

JanJakes:
The new AST driver already supports `SHOW COLUMNS` statement.

This PR adds support for `SHOW FULL COLUMNS`.


adamziel:
Looking great, thank you @JanJakes! My only not is that a single WordPress core unit test failed on this PR. I'm not sure how much that's related to this change and how much to something in WP trunk.

JanJakes:
> The following tests were expected to fail but passed

That's a lovely CI failure. I had no idea WPDB used `SHOW FULL COLUMNS` internally: https://github.com/WordPress/wordpress-develop/blob/16aa10f04dfaa8b9084f05213a9371526dc3e79b/src/wp-includes/class-wpdb.php#L3234

Fixed.

PR #311, opened 2026-01-27
Add a require statement for back compatibility

JanJakes:


JanJakes:
This is just a small fix, merging.

PR #312, opened 2026-01-28
Handle incorrectly stored MySQL metadata in the legacy-to-AST driver migrator

JanJakes:
When migrating from the legacy SQLite driver to the new AST-based driver, the information schema reconstructor reads MySQL column type metadata from the `_mysql_data_types_cache` table. Some older versions of the legacy driver stored invalid type definitions in this table, causing migration failures.

This PR detects and handles two types of invalid data:

**1. Truncated decimal definitions** — Before [PR #126](https://github.com/WordPress/sqlite-database-integration/pull/126), columns with multiple type arguments like `decimal(26, 8)` were incorrectly stored as `decimal(26,` (missing the second argument and closing parenthesis). For WooCommerce tables, the fix restores the correct decimal precision based on known WooCommerce column patterns. For other tables, these invalid definitions are discarded and the column type is inferred from SQLite.

**2. Index definitions mistaken for columns** — Before [commit b5a9fba](https://github.com/WordPress/sqlite-database-integration/commit/b5a9fbaed4d0d843f792aaa959e3d00f193ff1ee), index definitions like `KEY timestamp (timestamp)` were incorrectly parsed and stored as column `KEY` with type `timestamp(timestamp)`. The fix validates that type arguments are numeric, rejecting these malformed entries and falling back to the SQLite type inference.

Fixes https://github.com/WordPress/wordpress-playground/issues/3050.

### DIFF
--- a/tests/WP_SQLite_Information_Schema_Reconstructor_Tests.php
+++ b/tests/WP_SQLite_Information_Schema_Reconstructor_Tests.php
@@ -310,6 +310,88 @@ class WP_SQLite_Information_Schema_Reconstructor_Tests extends TestCase {
 		);
 	}
 
+	public function testInvalidDataTypeCacheDataForDecimalDefinition(): void {
+		// Recreate the invalid database state before the following fix:
+		//   https://github.com/WordPress/sqlite-database-integration/pull/126
+		$connection = $this->engine->get_connection();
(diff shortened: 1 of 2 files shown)


adamziel (approved):

adamziel:
LGTM, thank you for fixing this!

PR #314, opened 2026-01-30
Fix implicit result column name/alias for `SELECT` item literals with `NULL` bytes

JanJakes:
Fixes `SELECT` queries that have string literals with `NULL` bytes in them and no column alias.

E.g.:
```sql
SELECT `a<null-byte>b`;

-- result column name should be: 'a'
```

This is because in MySQL:
- There can’t be `\0` bytes in identifiers (like column names). The lexer will reject that.
- But string literals can have them.
- And there is a way to make a string literal “identifier-like” (implicit alias = the result “column name”): `SELECT 'abc\0xyz'` The result column name will be: `abc`

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
+++ b/wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
@@ -4566,6 +4566,13 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 		$is_text_string_literal = $text_string_literal && $item === $this->translate( $text_string_literal );
 		if ( $is_text_string_literal ) {
 			$alias = $text_string_literal->get_first_child_token()->get_value();
+
+			// When the literal value contains a NULL byte, MySQL truncates the
+			// resulting identifier at the position of the first one of them.
+			$fist_null_byte_pos = strpos( $alias, "\0" );
+			if ( false !== $fist_null_byte_pos ) {
+				$alias = substr( $alias, 0, $fist_null_byte_pos );
+			}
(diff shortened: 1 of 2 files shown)


wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -4566,6 +4566,13 @@ public function translate_select_item( WP_Parser_Node $node ): string {
 		$is_text_string_literal = $text_string_literal && $item === $this->translate( $text_string_literal );
 		if ( $is_text_string_literal ) {
 			$alias = $text_string_literal->get_first_child_token()->get_value();
+
+			// When the literal value contains a NULL byte, MySQL truncates the
+			// resulting identifier at the position of the first one of them.
+			$fist_null_byte_pos = strpos( $alias, "\0" );
```
adamziel, line 4572: Does this matter in more places than just `translate_select_item`?

adamziel (approved):
LGTM and solves a problem so looks good for merging. We may want to follow up with a more general solution if that matters beyond the SELECT columns list.

wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @adamziel I will check that a bit deeper in a follow-up PR. There is the case where you type this explicitly (`SELECT 1 AS 'a\0b'`), and we need a follow-up to tackle that. Then, I think there is no other way where a string literal gets "converted" to an alias, **but**—I found some inconsistencies in how we process string literals vs. quoted identifiers and want to improve that too. We seem to treat them the same, but they shouldn't. It might be an interesting edge case bug going down to the lexer.

PR #321, opened 2026-02-13
Improve identifier and string escaping in the legacy driver

JanJakes:
Improve escaping consistency in the legacy SQLite translator:

- Use `quote_identifier()` for all identifier interpolations (table names, column names, index names, trigger names) in SQL queries, token values, and DDL reconstruction output.
- Use parameterized queries and `PDO::quote()` for all string literals processed by the translator.

See added tests for more details.

### DIFF
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -3602,6 +3602,7 @@ QUERY
 		foreach ( $this->engine->executed_sqlite_queries as $q ) {
 			if ( stripos( $q['sql'], 'INSERT' ) !== false && stripos( $q['sql'], '_options' ) !== false ) {
 				$insert_query = $q;
+				break;
 			}
 		}
 		$this->assertNotNull( $insert_query );
@@ -3635,6 +3636,7 @@ QUERY
 		foreach ( $this->engine->executed_sqlite_queries as $q ) {
 			if ( stripos( $q['sql'], 'INSERT' ) !== false && stripos( $q['sql'], '_options' ) !== false ) {
 				$insert_query = $q;
+				break;
 			}
 		}
 		$this->assertNotNull( $insert_query );
--- a/tests/WP_SQLite_Translator_Tests.php
+++ b/tests/WP_SQLite_Translator_Tests.php
@@ -1170,14 +1170,14 @@ class WP_SQLite_Translator_Tests extends TestCase {
 					'name'     => '___tmp_table_created_at_on_update__',
 					'tbl_name' => '_tmp_table',
 					'rootpage' => '0',
-					'sql'      => "CREATE TRIGGER \"___tmp_table_created_at_on_update__\"\n\t\t\tAFTER UPDATE ON \"_tmp_table\"\n\t\t\tFOR EACH ROW\n\t\t\tBEGIN\n\t\t\t  UPDATE \"_tmp_table\" SET \"created_at\" = CURRENT_TIMESTAMP WHERE rowid = NEW.rowid;\n\t\t\tEND",
+					'sql'      => "CREATE TRIGGER `___tmp_table_created_at_on_update__`\n\t\t\tAFTER UPDATE ON `_tmp_table`\n\t\t\tFOR EACH ROW\n\t\t\tBEGIN\n\t\t\t  UPDATE `_tmp_table` SET `created_at` = CURRENT_TIMESTAMP WHERE rowid = NEW.rowid;\n\t\t\tEND",
 				),
 				(object) array(
 					'type'     => 'trigger',
 					'name'     => '___tmp_table_updated_at_on_update__',
 					'tbl_name' => '_tmp_table',
 					'rootpage' => '0',
-					'sql'      => "CREATE TRIGGER \"___tmp_table_updated_at_on_update__\"\n\t\t\tAFTER UPDATE ON \"_tmp_table\"\n\t\t\tFOR EACH ROW\n\t\t\tBEGIN\n\t\t\t  UPDATE \"_tmp_table\" SET \"updated_at\" = CURRENT_TIMESTAMP WHERE rowid = NEW.rowid;\n\t\t\tEND",
+					'sql'      => "CREATE TRIGGER `___tmp_table_updated_at_on_update__`\n\t\t\tAFTER UPDATE ON `_tmp_table`\n\t\t\tFOR EACH ROW\n\t\t\tBEGIN\n\t\t\t  UPDATE `_tmp_table` SET `updated_at` = CURRENT_TIMESTAMP WHERE rowid = NEW.rowid;\n\t\t\tEND",
 				),
 			),
 			$results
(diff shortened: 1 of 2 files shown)


tests/WP_SQLite_Translator_Tests.php
```diff
@@ -3557,4 +3557,149 @@ public function testCreateTableWithDefaultNowFunction() {
...
+		// With parameterization, the value is safely bound as a parameter.
+		$this->assertQuery(
+			'INSERT INTO _options (option_name, option_value) VALUES ("safe_key", "admin\" OR 1=1--")'
+		);
+
+		// Verify the injection payload is not present in the SQL sent to SQLite.
+		$insert_query = null;
+		foreach ( $this->engine->executed_sqlite_queries as $q ) {
+			if ( stripos( $q['sql'], 'INSERT' ) !== false && stripos( $q['sql'], '_options' ) !== false ) {
+				$insert_query = $q;
```
brandonpayton, line 3638: It's probably workable as-is for this test, but to match our intent, should it break after the first assignment?

wp-includes/sqlite/class-wp-sqlite-translator.php
```diff
@@ -3710,7 +3706,7 @@ private function generate_create_statement() {
 	 * @return stdClass[]
 	 */
 	protected function get_table_columns( $table_name ) {
-		return $this->execute_sqlite_query( "PRAGMA table_info(\"$table_name\");" )
+		return $this->execute_sqlite_query( 'PRAGMA table_info(' . $this->pdo->quote( $table_name ) . ');' )
```
brandonpayton, line 3710: What is the reason for using `$this->pdo->quote()` here and `$this->quote_identifier()` on line 3687? What is the difference?

brandonpayton (approved):
@JanJakes I scanned these changes, and they generally look good to me. I left a couple of questions.

I also asked Claude for feedback. You probably have already done so, but some of the items looked possibly legitimate, so I'll paste the feedback here. The following is from Claude Code (Opus 4.6):

----------

## Summary

This PR introduces two systemic improvements to the legacy SQLite translator:

1. **A new `quote_identifier()` method** that properly wraps identifiers in backticks and escapes internal backticks by doubling them.
2. **Consistent use of `PDO::quote()` and parameterized binding** for string literals instead of manual quoting.

It touches ~30 locations across `class-wp-sqlite-translator.php` and adds 7 well-targeted tests. The changes are clearly security-motivated and address real escaping gaps.

---

## What's Good

**Centralized identifier quoting.** The new `quote_identifier()` method eliminates a scattered mix of `"\"$var\""`, `'"' . $var . '"'`, and `` '`' . $var . '`' `` patterns, replacing them all with a single correct implementation. This is exactly the right approach.

**Proper `normalize_column_name()`.** The old `trim($column_name, '`\'"')` was lossy — it would mangle an identifier like `` `col``name` `` by stripping outer backticks but leaving the escaped inner ones. The new version correctly strips matching quote pairs and unescapes doubled quote characters.

**`extract_bound_parameter()` broadened to all strings.** Removing the `FLAG_STRING_SINGLE_QUOTES` restriction means double-quoted strings (which MySQL treats as string literals in non-ANSI mode) now get parameterized instead of being inlined into the SQL. This directly addresses SQL injection vectors through double-quoted strings — the `testDoubleQuotedStringWithBackslashEscapeDoesNotCauseInjection` test demonstrates this well.

**`escape_like_to_glob()` cleanup.** The old version did `trim($pattern, "'\"")` then manually wrapped in single quotes — a classic escaping gap if the pattern contains a literal single quote. Using `$this->pdo->quote()` fixes this properly.

**`get_autoincrement_column()` regex fix.** Now handles both backtick-quoted and double-quote-quoted identifiers, with proper support for escaped internal quotes (`` [^`]|`` ``+ instead of `[^"]+`). This is necessary since the PR itself changes DDL output to use backticks.

**Tests are well-chosen.** They cover the specific edge cases that motivate the changes: backtick-in-identifier, double-quoted string parameterization, injection via backslash-escaped double quotes, single quotes in LIKE BINARY patterns, etc.

---

## Issues / Questions

### 1. `extract_bound_parameter` — double-quoted identifiers risk

By broadening from `FLAG_STRING_SINGLE_QUOTES` to `TYPE_STRING`, any double-quoted token classified as `TYPE_STRING` will now be extracted as a bound parameter. In MySQL, `"foo"` is a string literal (unless `ANSI_QUOTES` is enabled). But SQLite treats `"foo"` as an identifier.

The question: does the lexer ever emit double-quoted **identifiers** with `TYPE_STRING`? If the MySQL lexer classifies `` `table_name` `` as `TYPE_SYMBOL` but `"table_name"` as `TYPE_STRING`, then this change could accidentally parameterize identifiers in queries like:

```sql
SELECT * FROM "my_table"
```

Could you confirm how the lexer classifies double-quoted identifiers vs. double-quoted string literals? The `AS` keyword guard only protects alias contexts.

### 2. Inconsistency: `escape_like_to_glob` receives unquoted value but old code expected quoted

The old code did `trim($pattern, "'\"")` to strip quotes, implying `$pattern_token->value` included surrounding quotes. The new code skips the trim and goes straight to `str_replace` + `pdo->quote()`.

In `translate_like_binary()`, the pattern comes from `$pattern_token->value`. If the token value still includes surrounding quotes (e.g., `'it''s%'`), then `pdo->quote()` will double-quote it, producing something like `''it''s%''`. Could you verify that `$pattern_token->value` is unquoted at this point? The test `testLikeBinaryWithSingleQuoteInFormat` passes, which is encouraging, but I want to be sure the token's `.value` property is consistently unquoted.

### 3. `execute_create_table` — extra space before `$if_not_exists`

Line 943 in the diff:
```php
'CREATE ' . $unique . 'INDEX ' . $if_not_exists . ' ' . $this->quote_identifier( $index_name ) . ...
```

If `$if_not_exists` is empty string, this produces `CREATE INDEX  `` `index_name` `` ...` with a double space. The old code had the same issue since `$if_not_exists` was interpolated inline, but since you're touching this line, it might be worth a trim or conditional space. Minor nit.

### 4. Missing `normalize_column_name` on `execute_delete` table name (line 1420/1434)

The PR adds `normalize_column_name()` calls when extracting table names in `execute_delete()`:

```php
$this->table_name = $this->normalize_column_name( $rewriter->skip()->value );
```

This is good — but should the same treatment be applied in other places where `$this->table_name` is set from token values? A quick grep shows `$this->table_name` is assigned in several other methods. If those tokens can also be quoted, they'd need the same normalization. Though that may be out of scope for this PR.

### 5. Token type for quoted identifiers in rewriter

In several places, the PR changes token types when adding quoted identifiers:

```php
// Old:
new WP_SQLite_Token( $this->table_name, WP_SQLite_Token::TYPE_KEYWORD, WP_SQLite_Token::FLAG_KEYWORD_RESERVED )
// New:
new WP_SQLite_Token( $this->quote_identifier( $this->table_name ), WP_SQLite_Token::TYPE_KEYWORD, WP_SQLite_Token::FLAG_KEYWORD_KEY )
```

The flag change from `FLAG_KEYWORD_RESERVED` to `FLAG_KEYWORD_KEY` seems intentional and correct (these are identifiers, not reserved words). Just want to confirm the rewriter doesn't filter or treat tokens differently based on these flags.

### 6. `$ids_to_delete` still not parameterized

In `execute_delete()` around the diff's line 263-268:

```php
"DELETE FROM {$quoted_table} WHERE {$quoted_pk} IN (" . implode( ',', $ids_to_delete ) . ')'
```

The `$ids_to_delete` values come from a prior query result (`$id['id_1']`), so they're not user-controlled. But for defense-in-depth consistency with the rest of the PR's approach, these could also use parameterized binding. Not a blocker, just noting.

---

## Overall Assessment

This is a solid, well-scoped security hardening PR. The `quote_identifier()` method and consistent `PDO::quote()` usage are clear improvements. The tests cover the key edge cases. The main thing I'd want clarified is item #1 (how the lexer classifies double-quoted tokens) since that determines whether broadening `extract_bound_parameter` is safe or could accidentally parameterize identifiers.

wp-includes/sqlite/class-wp-sqlite-translator.php
JanJakes: @brandonpayton Drivers like PDO and mysqli provide a `quote()` function for string values, but unfortunately, no analogous function for identifiers. I guess the reasoning is that usually, one types identifiers such as table and column names in the code as constant values. It's the same thing for prepared statements—there is no way to parametrize identifiers.

Now back to the original question. The difference is that strings and identifiers have different syntax and apply different escaping and quoting rules. Which one belongs where is defined by the SQL grammar, and I think in this case it is `CREATE TABLE <identifier>` vs. `FUNCTION_CALL(<string>)`.

To confirm, I tried this scenario:
```sql
-- OK: the argument is a string literal
SELECT * FROM pragma_table_info('t');

-- ERROR: no such column: t (these are identifier syntaxes)
SELECT * FROM pragma_table_info(t);
SELECT * FROM pragma_table_info(`t`);
SELECT * FROM pragma_table_info("t");
SELECT * FROM pragma_table_info([t]);
```

tests/WP_SQLite_Translator_Tests.php
JanJakes: 👍 Fixed in https://github.com/WordPress/sqlite-database-integration/pull/321/changes/ece1e39fafaacb6a40df0b39014c8839b3eeeb26.

JanJakes:
Thanks for the thorough review, @brandonpayton! I went through all the comments and addressed them when needed.

For the Claude-generated items:

**1. `extract_bound_parameter` — double-quoted identifiers risk:** I verified that this is safe. In the legacy lexer, a double-quoted literal is always classified as `TYPE_STRING`, because [`parse_string` runs before `parse_symbol`](https://github.com/WordPress/sqlite-database-integration/blob/065098384942af69af20de89f95ee0a99fd1170b/wp-includes/sqlite/class-wp-sqlite-lexer.php#L66).

**2. `escape_like_to_glob` — unquoted value:** This is correct, because `$token->value` is already unquoted by [the token's `extract()` method](https://github.com/WordPress/sqlite-database-integration/blob/065098384942af69af20de89f95ee0a99fd1170b/wp-includes/sqlite/class-wp-sqlite-token.php#L278). The old `trim()` was usually a no-op, and in fact could corrupt values with quotes in them.

**3. Double space when `$if_not_exists` is empty:** Since this is pre-existing, I'd leave it as-is in the legacy driver.

**4. Missing `normalize_column_name` in other methods:** Good point to make this consistent. Done in https://github.com/WordPress/sqlite-database-integration/pull/321/changes/84f2455afed57ab707872ba131d05bf02d9924c0.

**5. Token flags:** These are newly built output tokens. I can confirm this has no effect on how the output SQL is built.

**6. `$ids_to_delete` not parameterized:** Good point. Addressed in https://github.com/WordPress/sqlite-database-integration/pull/321/commits/08f0f15cbf07bb9437c2f2ad933b55d92f4eecc4.

wp-includes/sqlite/class-wp-sqlite-translator.php
brandonpayton: @JanJakes I think this should have been obvious to me. Thank you for the thoughtful explanation!

brandonpayton (approved):
Thanks for the updates and explanations, @JanJakes. This looks good to go.

PR #323, opened 2026-02-25
Add AI agent configuration and devcontainer

JanJakes:
## Summary

This PR adds configuration for AI coding agents and a development container:

- **AI agent guidance:** Add `AGENTS.md` with project architecture overview, commands, coding conventions, security and compatibility notes, and git practices. `CLAUDE.md` references it.
- **Claude Code settings:** Add `.claude/settings.json` with pre-approved permissions for common development commands.
- **Devcontainer:** Add root-level `.devcontainer/` with PHP 8.4, pdo_sqlite, Node.js 20, and Docker-in-Docker — suitable for developing the SQLite driver and running all test suites.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -15,7 +15,7 @@
 			"Bash(npm --prefix wordpress *)",
 			"Bash(npm --prefix tests/e2e *)",
 			"WebFetch(domain:developer.wordpress.org)",
-			"WebFetch(domain:www.sqlite.org)"
+			"WebFetch(domain:sqlite.org)"
 		]
 	}
 }
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -4,6 +4,7 @@
 		"allow": [
 			"Bash(composer install *)",
 			"Bash(composer update *)",
+			"Bash(composer run *)",
 			"Bash(composer check-cs *)",
 			"Bash(composer fix-cs *)",
 			"Bash(composer test *)",
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,21 @@
+{
+	"$schema": "https://json.schemastore.org/claude-code-settings.json",
+	"permissions": {
+		"allow": [
+			"Bash(composer install *)",
+			"Bash(composer update *)",
+			"Bash(composer check-cs *)",
+			"Bash(composer fix-cs *)",
+			"Bash(composer test *)",
+			"Bash(composer wp-*)",
+			"Bash(vendor/bin/phpunit *)",
+			"Bash(vendor/bin/phpcs *)",
+			"Bash(vendor/bin/phpcbf *)",
+			"Bash(vendor/bin/parallel-lint *)",
+			"Bash(npm --prefix wordpress *)",
+			"Bash(npm --prefix tests/e2e *)",
+			"WebFetch(domain:developer.wordpress.org)",
+			"WebFetch(domain:www.sqlite.org)"
+		]
+	}
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,13 @@
+{
+    "name": "SQLite Database Integration",
+    "image": "mcr.microsoft.com/devcontainers/php:8.4",
+    "features": {
+        "ghcr.io/devcontainers/features/node:1": {
+            "version": "20"
+        },
+        "ghcr.io/devcontainers/features/docker-in-docker:2": {
+            "moby": false
+        }
+    },
+    "postCreateCommand": "composer install"
+}
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ composer.lock
 .DS_Store
 ._*
 /wordpress
+/.claude/settings.local.json
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,161 @@
+<!--
+MAINTENANCE: Update this file when:
+- Adding/removing composer scripts
+- Changing the directory structure (new modules, major refactors)
+- Modifying build/test workflows
+- Adding new architectural patterns or conventions
+-->
+
+# AGENTS.md
+This file provides guidance to AI coding agents when working with code in this repository.
+
+## Project overview
+This project implements SQLite database support for MySQL-based projects.
+
+It is a monorepo that includes the following components:
+- **MySQL lexer** — A fast MySQL lexer with multi-version support.
+- **MySQL parser** — An exhaustive MySQL parser with multi-version support.
+- **SQLite driver** — A MySQL emulation layer on top of SQLite with a PDO-compatible API.
+- **MySQL proxy** — A MySQL binary protocol implementation to support MySQL-based projects beyond PHP.
+- **WordPress plugin** — A plugin that adds SQLite support to WordPress.
+- **Test suites** — A set of extensive test suites to cover MySQL syntax and functionality.
+
+The codebase is pure PHP with zero dependencies. It supports PHP 7.2 through 8.5, MySQL syntax from version 5.7 onward, and requires SQLite 3.37.0 or newer (with legacy mode down to 3.27.0).
+
+### New and old driver
+At the moment, the project includes two MySQL-on-SQLite driver implementations:
+1. A new, AST-based MySQL-on-SQLite driver (`class-wp-pdo-mysql-on-sqlite.php`).
+2. A legacy, token-based MySQL-to-SQLite translator (`class-wp-sqlite-translator.php`).
+
+This state is temporary. The new driver will fully replace the legacy one. New features
+must always be implemented in the new driver. The legacy driver can receive small fixes.
+The new driver is under a `WP_SQLITE_AST_DRIVER` feature flag, but it is widely used.
+
+## Commands
+The codebase is written in PHP and Composer is used to manage the project.
+The following commands are useful for development and testing:
+
+```bash
+composer install                        # Install dependencies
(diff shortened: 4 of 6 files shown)


.claude/settings.json
```diff
@@ -0,0 +1,21 @@
...
+			"Bash(composer test *)",
+			"Bash(composer wp-*)",
+			"Bash(vendor/bin/phpunit *)",
+			"Bash(vendor/bin/phpcs *)",
+			"Bash(vendor/bin/phpcbf *)",
+			"Bash(vendor/bin/parallel-lint *)",
+			"Bash(npm --prefix wordpress *)",
+			"Bash(npm --prefix tests/e2e *)",
+			"WebFetch(domain:developer.wordpress.org)",
+			"WebFetch(domain:www.sqlite.org)"
```
ashfame: I think it's primarily `sqlite.org` (without non-www) though seems accessible on both without canonical meta tag. Bold.

.gitignore
```diff
@@ -6,3 +6,4 @@ composer.lock
 .DS_Store
 ._*
 /wordpress
+/.claude/settings.local.json
```
ashfame, line 9: Curious: This seems like a path under the container, but how come it has your `settings.local.json` in there?

AGENTS.md
```diff
@@ -0,0 +1,161 @@
+<!--
+MAINTENANCE: Update this file when:
+- Adding/removing composer scripts
+- Changing the directory structure (new modules, major refactors)
+- Modifying build/test workflows
+- Adding new architectural patterns or conventions
+-->
+
+# AGENTS.md
+This file provides guidance to AI coding agents when working with code in this repository.
```
ashfame: Think we can skip this line.
```suggestion

```

AGENTS.md
```diff
@@ -0,0 +1,161 @@
...
+composer run test -- --filter testName  # Run specific unit test class/method
+composer run test-e2e                   # Run E2E tests (Playwright via WP env)
+
+# WordPress tests
+composer run wp-setup                   # Set up WordPress with SQLite for tests
+composer run wp-test-start              # Start WordPress environment (Docker)
+composer run wp-test-php                # Run WordPress PHPUnit tests
+composer run wp-test-e2e                # Run WordPress E2E tests (Playwright)
+composer run wp-test-clean              # Clean up WordPress environment (Docker and DB)
+```
```
ashfame, line 57: I don't think any of this is needed. Would unnecessarily eat some tokens with every inference.

AGENTS.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Architecture
+The project consists of multiple components providing different APIs that funnel
+into the SQLite driver to support diverse use cases both inside and outside the
+PHP ecosystem.
+
+### Component overview
```
ashfame, line 64: I asked Opus 4.6 to explain this diagram & looks like it hallucinated:

```
This is an architecture diagram showing how different consumers (on the left) connect to a SQLite database (on the right) through intermediate components.
There are two paths:

WordPress path: WordPress Playground (Studio, wp-env) connects through a wpdb drop-in, which feeds into a SQLite driver (using the PDO API), which talks to the actual SQLite database.
MySQL client path: MySQL CLI and desktop clients connect through a MySQL proxy, which also feeds into the same SQLite driver.

The vertical line (│) represents a shared junction point — both the wpdb drop-in and the MySQL proxy converge there before going through the single SQLite driver to the database.
The top group (plain PHP apps like Adminer, phpMyAdmin, and WordPress with plugins) are listed as consumers but don't have explicit arrows — implying they connect through one of these two paths depending on context.
The core idea: This project provides a compatibility layer that lets MySQL-expecting tools (WordPress, MySQL clients, PHP apps) work against a SQLite backend instead of a real MySQL server — which is exactly what the WordPress Playground SQLite integration does.
```

Got the facts right but the interpretation of first line wrong. I think we change it to just text indented lines and repeat the right side of the diagram. But either way, the diagram is not super-relevant for LLM inference, I think.

AGENTS.md
```diff
@@ -0,0 +1,161 @@
...
+Always scrutinize implemented logic for security vulnerabilities and verify any
+assumptions. Never take shortcuts.
+
+## Compatibility
+This project must support a range of PHP and SQLite versions, and it must evolve
+the public APIs responsibly, following semantic versioning practices.
+
+In particular:
+- **Public APIs:** It's possible to evolve the public API, but this must always be
+  surfaced to the developer so versioning decisions can be made.
```
ashfame, line 150: <3

ashfame (approved):
Left a few comments to consider but good to land. Nicely done!

.gitignore
JanJakes: It's for local indeed, not the container.

AGENTS.md
JanJakes: This was copy-pasted from Playground 😆 But I think you're right.

AGENTS.md
JanJakes: Oh, really? I actually thought this was very useful, because the only other way to know this is by reading `composer.json`, which adds a delay and uses more tokens.

AGENTS.md
JanJakes: Interesting. It was largely co-created with the exact same model 😄

Maybe I should put this one in a README for humans.

In a textual form, it could perhaps be somewhat useful for the model to know what components the query goes through in each use case.

AGENTS.md
```diff
@@ -0,0 +1,161 @@
...
+
+It is a monorepo that includes the following components:
+- **MySQL lexer** — A fast MySQL lexer with multi-version support.
+- **MySQL parser** — An exhaustive MySQL parser with multi-version support.
+- **SQLite driver** — A MySQL emulation layer on top of SQLite with a PDO-compatible API.
+- **MySQL proxy** — A MySQL binary protocol implementation to support MySQL-based projects beyond PHP.
+- **WordPress plugin** — A plugin that adds SQLite support to WordPress.
+- **Test suites** — A set of extensive test suites to cover MySQL syntax and functionality.
+
+The codebase is pure PHP with zero dependencies. It supports PHP 7.2 through 8.5, MySQL syntax from version 5.7 onward, and requires SQLite 3.37.0 or newer (with legacy mode down to 3.27.0).
```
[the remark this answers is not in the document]
JanJakes: Oh, Copilot 🤦

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,13 @@
...
+    "image": "mcr.microsoft.com/devcontainers/php:8.4",
+    "features": {
+        "ghcr.io/devcontainers/features/node:1": {
+            "version": "20"
+        },
+        "ghcr.io/devcontainers/features/docker-in-docker:2": {
+            "moby": false
+        }
+    },
+    "postCreateCommand": "composer install"
```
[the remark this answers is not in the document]
JanJakes: The Docker image already includes `pdo_sqlite`.

AGENTS.md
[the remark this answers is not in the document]
JanJakes: Addressed in https://github.com/WordPress/sqlite-database-integration/pull/323/commits/7f082167f1d9e88efbabce8b35c48229cd006445.

AGENTS.md
JanJakes: Addressed in https://github.com/WordPress/sqlite-database-integration/pull/323/commits/7f082167f1d9e88efbabce8b35c48229cd006445.

AGENTS.md
JanJakes: Addressed in https://github.com/WordPress/sqlite-database-integration/pull/323/commits/7f082167f1d9e88efbabce8b35c48229cd006445. I also added a README with human-readable diagrams in https://github.com/WordPress/sqlite-database-integration/pull/323/changes/c7a0d9ace11306ca15df7429c0fc40beddce6aa5.

.claude/settings.json
JanJakes: Addressed in https://github.com/WordPress/sqlite-database-integration/pull/323/commits/df285125de8546102c3beb5f7222fcac4aece056.

.gitignore
ashfame: But its `/.claude`

AGENTS.md
ashfame: Wouldn't it always read those kind of files? It can't rely on our guard-rails to act as source of truth, I believe.

.gitignore
JanJakes: @ashfame That means it ignores the path in the repository root only, but not recursively. Since the path is specific enough, both options would work fine here.

ashfame: oh yes, I completely forgot about that bit. I kept thinking its on root, so must be a path in container. Thanks!

AGENTS.md
JanJakes: There's a difference.

1) On this branch — **477 tokens** and faster:
    Screenshot 2026-02-26 at 11 05 29

2) On trunk — **697 tokens** and slower:
    Screenshot 2026-02-26 at 11 06 20

ashfame: Cool that you checked. But would it always do that when not explicitly invoked? Something to think about. Anyway, let's keep it if you think is better. General advice by Claude Code creator is "wipe it and start fresh when it gets too big", so :)

JanJakes: @ashfame The field guide now recommends listing commands too, so I guess keep it?

(The trouble with Composer scripts specifically is that they are not documented at all in `composer.json` itself. When you see what it executes, you may get an idea of what it does, or you may not.)

ashfame: Yes sure 👍

PR #325, opened 2026-03-06
fix: query() returned null instead of 0 for statements affecting 0 rows

### LINKED ISSUE #322: `null` returned instead of `0` for DELETE queries with no matching rows in `WP_SQLite_Driver`
According to the DocBlocks, the `query()` method of `wpdb` should return either an `int` or a `bool`: https://github.com/WordPress/wordpress-develop/blob/0c3e42c238fc2ca907bbe649160598f39267d812/src/wp-includes/class-wpdb.php#L2205-L2217

However, I believe, there is at least one case where `null` is returned where `0` would be expected: when a DELETE statement matches no rows. This seems to affect only the `WP_SQLite_Driver`.

From an initial look, the `get_last_return_value()` returns `null` at this point:
https://github.com/WordPress/sqlite-database-integration/blob/535b42a935a778740387a8223c788f8d6155d5f8/wp-includes/sqlite/class-wp-sqlite-db.php#L452

Even if returning `null` is intentional in some internal context, I think, it should be coerced to `0` before being returned from `query()`, to match the stated behavior.

I tried to put together a minimal test case, which hopefully it's a helpful one:

```php
<?php

class Test_Delete_Returns_Zero_Affected_Rows_When_No_Matching_Row extends \PHPUnit\Framework\TestCase
{
	public function test_wp_sqlite_translator()
	{
		$driver = new WP_SQLite_Translator();

		$driver->query('CREATE TABLE t (id INT)');
		$driver->query('INSERT INTO t (id) VALUES (1)');
		$driver->query('DELETE FROM t WHERE id = 1');

		$this->assertSame(1, $driver->get_affected_rows());

		$driver->query('DELETE FROM t WHERE id = 1');

		$this->assertSame(0, $driver->get_affected_rows());
	}

	public function test_wp_sqlite_driver()
	{
		$driver = new WP_SQLite_Driver(
			new WP_SQLite_Connection(array('path' => ':memory:')),
			'wp'
		);

		$driver->query('CREATE TABLE t (id INT)');
		$driver->query('INSERT INTO t (id) VALUES (1)');
		$driver->query('DELETE FROM t WHERE id = 1');

		$this->assertSame(1, $driver->get_last_return_value());

		$driver->query('DELETE FROM t WHERE id = 1');

		// This fails, as NULL is returned.
		$this->assertSame(0, $driver->get_last_return_value());
	}
}
```

After installing the PHP dependencies, I was able to run this with:

```shell
./vendor/bin/phpunit tests/Test_Delete_Returns_Zero_Affected_Rows_When_No_Matching_Row.php
```

chloe-pomegranate:
### Summary

Fixes #322 

WP_SQLite_Driver::query() returned null instead of 0 for statements affecting zero rows (e.g., DELETE FROM x WHERE id = 1 when no matching row exists). This caused wpdb::query() to also return null, breaking its documented contract of returning int|bool.

### Problem

In WP_SQLite_Driver::query(), the return logic for non-SELECT queries was:

1. rowCount() > 0 - return row count
2. Otherwise - return null

This meant any DELETE, UPDATE, INSERT statement with 0 affected rows returned null instead of 0.

### Fix

Simplified the logic to always return rowCount() for non-SELECT statements. 

DELETE / UPDATE / INSERT will now correctly returns 0 instead of null when no rows affected.
CREATE / ALTER / DROP / TRUNCATE now returns 0 instead of null, but wp-includes/sqlite/class-wp-sqlite-db.php never reads this value for these queries (it returns true unconditionally at line 448-449)

### Testing

- Added testDeleteReturnsZeroAffectedRowsWhenNoMatchingRows - verifies DELETE returns 0 when no rows match.
- Added testUpdateReturnsZeroAffectedRowsWhenNoMatchingRows - verifies UPDATE returns 0 when no rows match.
- Updated existing test assertions from assertNull to assertSame( 0, ... ) to reflect the corrected behaviour 
- Full test suite passes (777 tests)

### DIFF
--- a/tests/WP_SQLite_Driver_Metadata_Tests.php
+++ b/tests/WP_SQLite_Driver_Metadata_Tests.php
@@ -719,7 +719,7 @@ class WP_SQLite_Driver_Metadata_Tests extends TestCase {
 			'TRUNCATE TABLE wp_comments;'
 		);
 		$actual = $this->engine->get_query_results();
-		$this->assertNull( $actual );
+		$this->assertSame( 0, $actual );
 		$this->assertTableEmpty( 'wp_comments', true );
 	}
 
--- a/tests/WP_SQLite_Driver_Tests.php
+++ b/tests/WP_SQLite_Driver_Tests.php
@@ -920,7 +920,7 @@ class WP_SQLite_Driver_Tests extends TestCase {
 				KEY user_email (user_email)
 			) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_520_ci"
 		);
-		$this->assertNull( $result );
+		$this->assertSame( 0, $result );
 
 		$this->assertQuery( 'DESCRIBE wptests_users;' );
 		$results = $this->engine->get_query_results();
@@ -1018,7 +1018,7 @@ class WP_SQLite_Driver_Tests extends TestCase {
 				PRIMARY KEY  (ID)
 			) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_520_ci'
 		);
-		$this->assertNull( $result );
+		$this->assertSame( 0, $result );
 	}
 
 	public function testCreateTableSpatialIndex() {
@@ -1028,7 +1028,7 @@ class WP_SQLite_Driver_Tests extends TestCase {
 				UNIQUE KEY (ID)
 			)'
 		);
-		$this->assertNull( $result );
+		$this->assertSame( 0, $result );
 	}
 
 	public function testCreateTableWithMultiValueColumnTypeModifiers() {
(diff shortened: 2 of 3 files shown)


JanJakes:
Thanks for fixing this, @chloe-pomegranate!

I'm just double-checking [what WPDB expects](https://github.com/WordPress/wordpress-develop/blob/4d83a5bf761c8b783294a6a78103be751389f871/src/wp-includes/class-wpdb.php#L2306-L2331).

For `create|alter|truncate|drop`, it seems to just use `$return_val = $this->result;`, which seems to be obtained as `$this->result = mysqli_query( $this->dbh, $query );`, meaning that for these statements, WPDB expects `true` rather than `0` ([PHP docs](https://www.php.net/manual/en/mysqli.query.php)).

For a moment, I thought we had it wrong, but this `true` value [is actually already accounted for](https://github.com/WordPress/sqlite-database-integration/blob/d87bc8f53f5f26090200fcf039e47ebbd1f6102b/wp-includes/sqlite/class-wp-sqlite-db.php#L449) in the WPDB override by closely mirroring WPDB logic, so I think we're all good. Thanks!

PR #329, opened 2026-03-18
Add SQLite database application ID and new consistent file extension

zaerl:
This PR adds the `3948349` integer as the [SQLite Application ID](https://sqlite.org/fileformat.html#application_id) that is standard procedure to take when a SQLite database is used as an application file format[^1]. A new consistent file extension has been chosen.

1. `PRAGMA application_id = SQLITE_DB_APPLICATION_ID` is set for new and old databases
2. New databases are created with the new file extension
3. Legacy files are renamed for backward compatibility

[^1]: https://sqlite.org/appfileformat.html

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-sqlite-connection.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-connection.php
@@ -96,6 +96,9 @@ class WP_SQLite_Connection {
 		if ( $journal_mode && in_array( $journal_mode, self::SQLITE_JOURNAL_MODES, true ) ) {
 			$this->query( 'PRAGMA journal_mode = ' . $journal_mode );
 		}
+
+		// Set the application ID to identify this as a WordPress SQLite database.
+		$this->query( 'PRAGMA application_id = ' . SQLITE_DB_APPLICATION_ID );
 	}
 
 	/**
--- a/wp-includes/sqlite-ast/class-wp-sqlite-connection.php
+++ b/wp-includes/sqlite-ast/class-wp-sqlite-connection.php
@@ -63,6 +63,7 @@ class WP_SQLite_Connection {
 	 *     @type int|null    $timeout      Optional. SQLite timeout in seconds.
 	 *                                     The time to wait for a writable lock.
 	 *     @type string|null $journal_mode Optional. SQLite journal mode.
+	 *     @type int|null $application_id  Optional. SQLite application ID.
 	 * }
 	 *
 	 * @throws InvalidArgumentException When some connection options are invalid.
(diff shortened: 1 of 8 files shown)


wp-includes/sqlite/class-wp-sqlite-db.php
```diff
@@ -306,6 +306,24 @@ public function db_connect( $allow_bail = true ) {
...
+		// default file name and the new file does not already exist.
+		if ( ! defined( 'DB_FILE' ) && ! file_exists( FQDB ) ) {
+			$old_db_path = FQDBDIR . '.ht.sqlite';
+
+			if ( file_exists( $old_db_path ) ) {
+				rename( $old_db_path, FQDB );
+
+				foreach ( array( '-wal', '-shm' ) as $suffix ) {
+					if ( file_exists( $old_db_path . $suffix ) ) {
+						rename( $old_db_path . $suffix, FQDB . $suffix );
```
JanJakes: Should we check whether `rename` succeeded, and if not, then something like `wp_die( 'Failed to rename database file.', 'Error!' );`?

wp-includes/sqlite-ast/class-wp-sqlite-connection.php
```diff
@@ -96,6 +96,9 @@ public function __construct( array $options ) {
 		if ( $journal_mode && in_array( $journal_mode, self::SQLITE_JOURNAL_MODES, true ) ) {
 			$this->query( 'PRAGMA journal_mode = ' . $journal_mode );
 		}
+
+		// Set the application ID to identify this as a WordPress SQLite database.
+		$this->query( 'PRAGMA application_id = ' . SQLITE_DB_APPLICATION_ID );
```
JanJakes: This application ID is specific to the WP plugin rather than to the driver itself, right? Just wondering if we should move this pragma call to `WP_SQLite_DB` (and possibly making the constant a `WP_SQLite_DB` class constant). That said, I know it's not straightforward to call the SQLite pragma from `WP_SQLite_DB`.

What if we pass it via `$options`? That seems like the best solution here to me.

JanJakes (commented):
@zaerl This looks great, thanks! I have just two small ideas.

wp-includes/sqlite/class-wp-sqlite-db.php
zaerl: Yes, good idea. Added.

wp-includes/sqlite-ast/class-wp-sqlite-connection.php
zaerl: Yes, it's the best solution. I've moved the constant to `WP_SQLite_DB` and now the connection reads `$options['application_id']`. Thanks. It's this plugin that asks for such an application ID, not the connection itself.

JanJakes (approved):
Looks great, thank you!

Issue #320, opened 2026-02-13
Address remaining WordPress PHPUnit test failures


JanJakes:
The WordPress PHPUnit test suite passes over 99% tests with SQLite, but some of them are still failing.

Let's look into these failures and see if we can address them:

```bash
# Errors
Tests_DB_Charset::test_invalid_characters_in_query
Tests_DB_Charset::test_set_charset_changes_the_connection_collation

# Failures
Tests_Admin_wpSiteHealth::test_object_cache_thresholds with data set #2
Tests_Admin_wpSiteHealth::test_object_cache_thresholds with data set #3
Tests_Comment::test_wp_new_comment_respects_comment_field_lengths
Tests_Comment::test_wp_update_comment
Tests_DB_Charset::test_get_column_charset with data set #0
Tests_DB_Charset::test_get_column_charset with data set #1
Tests_DB_Charset::test_get_column_charset with data set #2
Tests_DB_Charset::test_get_column_charset with data set #3
Tests_DB_Charset::test_get_column_charset with data set #4
Tests_DB_Charset::test_get_column_charset with data set #5
Tests_DB_Charset::test_get_column_charset with data set #6
Tests_DB_Charset::test_get_column_charset with data set #7
Tests_DB_Charset::test_get_column_charset_is_mysql_undefined with data set #0
Tests_DB_Charset::test_get_column_charset_is_mysql_undefined with data set #1
Tests_DB_Charset::test_get_column_charset_is_mysql_undefined with data set #2
Tests_DB_Charset::test_get_column_charset_is_mysql_undefined with data set #3
Tests_DB_Charset::test_get_column_charset_is_mysql_undefined with data set #4
Tests_DB_Charset::test_get_column_charset_is_mysql_undefined with data set #5
Tests_DB_Charset::test_get_column_charset_is_mysql_undefined with data set #6
Tests_DB_Charset::test_get_column_charset_is_mysql_undefined with data set #7
Tests_DB_Charset::test_get_column_charset_non_mysql with data set #0
Tests_DB_Charset::test_get_column_charset_non_mysql with data set #1
Tests_DB_Charset::test_get_column_charset_non_mysql with data set #2
Tests_DB_Charset::test_get_column_charset_non_mysql with data set #3
Tests_DB_Charset::test_get_column_charset_non_mysql with data set #4
Tests_DB_Charset::test_get_column_charset_non_mysql with data set #5
[...]

PR #331, opened 2026-03-20
Address WordPress PHPUnit test fails: charset detection, length validation etc

chloe-pomegranate:
## Summary
Fixes (some of) #320 - WordPress PHPUnit test failures related to charset/collation handling, column length validation, and db capabilities 

### New Methods
- `get_col_charset($table, $column)` - Returns charset for a table column by parsing the Collation field from SHOW FULL COLUMNS output
- `get_col_length($table, $column)` - Returns column length information for validation, supporting char/varchar/binary/varbinary and TEXT/BLOB 
- `has_cap($db_cap)` - Reports supported capabilities: collation, group_concat, subqueries, set_charset, utf8mb4, utf8mb4_520, identifier_placeholders
- `determine_charset($charset, $collate)` - Upgrades utf8 to utf8mb4 and handles collation mapping

### Existing Methods Changed
- `query()` - Added strip_invalid_text_from_query() validation to reject queries with invalid data for the column charset
- `close()` - Properly sets ready=false and has_connected=false when closing the connection

## Tests
Fixes from the issue:
- Charset detection tests
- Length validation tests
- Capability tests
- Charset tests
- Data handling tests 

## Other
Haven't touched these ones as they're generally all MySQL specific or SQLite handles them differently:
- `test_replace` - MySQL REPLACE returns affected rows differently than SQLite
- `test_close` - Similar to above
- `test_mysqli_flush_sync` - MySQL-specific, don't think we can do anything here
- `test_spatial_indices` - SQLite doesn't support SPATIAL KEY
- `test_wp_update_comment` 
- `test_strip_invalid_text` 
- `test_set_charset_changes_the_connection_collation`

## Notes
- For new methods I've used `@since 2.3.0` - can change to whatever :)


packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php
```diff
@@ -612,18 +814,31 @@ protected function load_col_info() {
...
 	 *
 	 * @param string $db_cap The feature to check for. Accepts 'collation',
 	 *                       'group_concat', 'subqueries', 'set_charset',
-	 *                       'utf8mb4', or 'utf8mb4_520'.
+	 *                       'utf8mb4', 'utf8mb4_520', or 'identifier_placeholders'.
 	 *
 	 * @return bool Whether the database feature is supported, false otherwise.
 	 */
 	public function has_cap( $db_cap ) {
-		return 'subqueries' === strtolower( $db_cap );
```
JanJakes, line 576: Do we still need this method override at all? The capabilities are derived from the MySQL version, so they depend on the `db_version()` implementation (currently hardcoded to `8.0`).

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php
```diff
@@ -612,18 +814,31 @@ protected function load_col_info() {
...
+
+		switch ( $db_cap ) {
+			case 'collation':
+			case 'group_concat':
+			case 'subqueries':
+			case 'set_charset':
+			case 'utf8mb4':
+			case 'utf8mb4_520':
+			case 'identifier_placeholders':
+				return true;
```
JanJakes: This one [seems to be dependent on WordPress version](https://github.com/WordPress/wordpress-develop/blob/e9afdf8981b78327a76025498d10c13c332ee070/src/wp-includes/class-wpdb.php#L4096).

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php
```diff
@@ -453,16 +644,27 @@ public function query( $query ) {
...
-		/*
-		 * @TODO: WPDB uses "$this->check_current_query" to check table/column
-		 *        charset and strip all invalid characters from the query.
-		 *        This is an involved process that we can bypass for SQLite,
-		 *        if we simply strip all invalid UTF-8 characters from the query.
-		 *
-		 *        To do so, mb_convert_encoding can be used with an optional
-		 *        fallback to a htmlspecialchars method. E.g.:
-		 *          https://github.com/nette/utils/blob/be534713c227aeef57ce1883fc17bc9f9e29eca2/src/Utils/Strings.php#L42
-		 */
```
JanJakes, line 431: The `TODO` explains what should be done for SQLite. However, the implementation doesn't seem to follow it, and it removes all of that context. The `TODO` is there for a reason, and we should follow it.

wp-includes/sqlite/class-wp-sqlite-db.php
```diff
@@ -145,10 +282,64 @@ public function set_sql_mode( $modes = array() ) {
...
+		$this->has_connected = false;
+
 		return true;
 	}
 
+	/**
+	 * Determines the best charset and collation for the database connection.
+	 *
+	 * This overrides wpdb::determine_charset() to handle SQLite's lack of mysqli.
+	 * WordPress expects utf8 to be upgraded to utf8mb4 when supported.
```
JanJakes: Let's just state that this closely mirrors the implementation of the parent method, [like we do for other methods](https://github.com/WordPress/sqlite-database-integration/blob/783790c58ddbe605d38f69d8a7a29858471237e7/wp-includes/sqlite/class-wp-sqlite-db.php#L92), keeping the parent method docs otherwise.

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php
```diff
@@ -145,10 +282,64 @@ public function set_sql_mode( $modes = array() ) {
 	 *
 	 * @return bool True to indicate the connection was successfully closed.
 	 */
+	/**
+	 * Close the database connection.
+	 *
+	 * @since 2.3.0
+	 *
+	 * @return bool True if connection was closed successfully.
+	 */
```
JanJakes, line 189: Let's just state that this closely mirrors the implementation of the parent method, like we do for other methods, keeping the parent method docs otherwise.

wp-includes/sqlite/class-wp-sqlite-db.php
```diff
@@ -78,17 +78,154 @@ public function set_charset( $dbh, $charset = null, $collate = null ) {
...
+	 *
+	 *     @type string $type   One of 'byte' or 'char'.
+	 *     @type int    $length The column length.
+	 * }
+	 */
+	public function get_col_length( $table, $column ) {
+		$table_key  = strtolower( $table );
+		$column_key = strtolower( $column );
+
+		// Check cached column metadata first.
```
JanJakes: The parent method ensures `$this->col_meta` is populated, but here it's missing.

Alternatively, we could also consider just calling the parent:
```php
try {
	$this->is_mysql = true; // We'd need to explain this hack.
	return parent::get_col_length( $table, $column );
} finally {
	$this->is_mysql = false;
}
```

wp-includes/sqlite/class-wp-sqlite-db.php
```diff
@@ -78,17 +78,154 @@ public function set_charset( $dbh, $charset = null, $collate = null ) {
...
+		 * @param string            $column  The name of the column being checked.
+		 */
+		$charset = apply_filters( 'pre_get_col_charset', null, $table, $column );
+		if ( null !== $charset ) {
+			return $charset;
+		}
+
+		// Use SHOW FULL COLUMNS to get column metadata with collation info.
+		// This works because the driver translates this to query the information schema.
+		$results = $this->get_results( "SHOW FULL COLUMNS FROM `{$table_key}`" );
```
JanJakes: Is there a reason to implement this differently than WPDB does?

Alternatively, we could also consider just calling the parent:
```php
try {
	$this->is_mysql = true; // We'd need to explain this hack.
	return parent::get_col_charset( $table, $column );
} finally {
	$this->is_mysql = false;
}
```

JanJakes:
Hi @chloe-pomegranate, thanks for your contribution! It demonstrates that we're only a few small fixes away from improving the WP unit test compatibility. I left some feedback, mostly with ideas of better aligning with WPDB and what makes sense for SQLite. Feel free to address them, or I can take it over later on. Thanks again for bringing these to our attention!

chloe-pomegranate:
hey @JanJakes no prob, would be happy to! I can't see the review at the moment though, am I missing something?

JanJakes (commented):
@chloe-pomegranate Ups, sorry, missed the submit review button.

wp-includes/sqlite/class-wp-sqlite-db.php
chloe-pomegranate: Amended to go with this method

wp-includes/sqlite/class-wp-sqlite-db.php
chloe-pomegranate: Amended to go with this method

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php
chloe-pomegranate: Added this

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php
chloe-pomegranate: Added a version_compare for this one

chloe-pomegranate:
@JanJakes Made a few updates from your review!

JanJakes:
I gave this another go and added some improvements and fixes. Merging now, thanks for your contribution!

PR #333, opened 2026-03-23
Add constant for SQLite driver loader path

JanJakes:
## Summary
Define `WP_MYSQL_ON_SQLITE_PATH` constant in `wp-pdo-mysql-on-sqlite.php` so that dependent projects can reliably determine the path to the SQLite driver loader file.

### DIFF
--- a/wp-pdo-mysql-on-sqlite.php
+++ b/wp-pdo-mysql-on-sqlite.php
@@ -1,5 +1,7 @@
 <?php
 
+define( 'WP_MYSQL_ON_SQLITE_LOADER_PATH', __FILE__ );
+
 /**
  * Load the PDO MySQL-on-SQLite driver and its dependencies.
  */


wojtekn (approved):

wp-pdo-mysql-on-sqlite.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+define( 'WP_MYSQL_ON_SQLITE_PATH', __FILE__ );
```
adamziel: Would it make sense to reflect the file name in the constant?

```suggestion
define( 'WP_PDO_MYSQL_ON_SQLITE_PHP_PATH', __FILE__ );
```

JanJakes: @adamziel I was looking at existing constants in WP, and there is no such convention (although there is also no PHP-path constant). If `WP_MYSQL_ON_SQLITE_PATH` may sound a bit like directory, then what about sth like `WP_MYSQL_ON_SQLITE_LOADER_PATH`?

JanJakes: @adamziel Ah, and for reflecting the filename itself, I wouldn't do that. It's more about "a path to a file that will load the driver". The filename might change by 3.0 (e.g., `wp-mysql-on-sqlite/load.php`).

JanJakes: Updated with `WP_MYSQL_ON_SQLITE_LOADER_PATH`.

JanJakes:
I'll go ahead with `WP_MYSQL_ON_SQLITE_LOADER_PATH`.

wp-pdo-mysql-on-sqlite.php
adamziel: Makes sense, thank you!

PR #341, opened 2026-03-30
Implement deterministic RAND(N) and improve unseeded RAND() translation

wp-fuse:
### Summary
This PR improves the MySQL-compatible `RAND()` support in the AST-based SQLite driver. It replaces the previous generic implementation with a bit-exact MySQL Linear Congruential Generator (LCG) and a robust SQLite-level translation for unseeded calls.

### Technical Changes
1.  **Seeded RAND(N)**: 
    - Refactored the `rand()` UDF in `WP_SQLite_PDO_User_Defined_Functions` to use an instance-local LCG.
    - Implemented MySQL’s specific 30-bit seed initialization constants (`0x10001`, `55555555`, `0x10000001`) to ensure bit-exact deterministic output for given seeds, as requested in the previous review.
    - Removed `mt_srand()` and `mt_rand()` usage to prevent PHP global state pollution.
2.  **Unseeded RAND()**:
    - Updated the AST node translation in `WP_PDO_MySQL_On_SQLite` to use: 
      `((RANDOM() & 0x7FFFFFFFFFFFFFFF) / 9223372036854775808.0)`.
    - This bitwise approach avoids integer overflow issues inherent to `ABS(-9223372036854775808)` in SQLite and ensures an unbiased distribution in the `[0, 1)` range.
3.  **Cleanup**: This is a standalone PR following the feedback to separate the `RAND()` fix from other compatibility layers. All string-replacement hacks and non-AST layers have been excluded.

### Validation
- **PHPUnit**: All 795 tests in the official test suite are passing.
- **CS**: Verified against WordPress Coding Standards via `phpcs`.

### DIFF
--- a/tests/WP_SQLite_Driver_Tests.php
+++ b/tests/WP_SQLite_Driver_Tests.php
@@ -9068,14 +9068,13 @@ END;
 					'mysqli:type'      => 8,
 				),
 				array(
-					// TODO: Fix custom "RAND()" function to behave like in MySQL.
-					'native_type'      => 'LONGLONG',          // DOUBLE in MySQL.
-					'pdo_type'         => PDO::PARAM_INT,      // PARAM_STR in MySQL.
+					'native_type'      => 'DOUBLE',
+					'pdo_type'         => PDO::PARAM_STR,
 					'flags'            => array( 'not_null' ),
 					'table'            => '',
 					'name'             => 'col_expr_19',
-					'len'              => 21,                  // 23 in MySQL.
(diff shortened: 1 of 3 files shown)


JanJakes:
Thanks for drafting this! I went deeper into a review and realized that it needs multiple changes, so I created https://github.com/WordPress/sqlite-database-integration/pull/363. Most notably, it needed extensive test coverage, and there were unhandled cases that I addressed in that PR.

PR #342, opened 2026-03-30
Add tests for MySQL row-level locking clauses

wp-fuse:
Add tests for MySQL row-level locking clauses (e.g., `FOR UPDATE`, `LOCK IN SHARE MODE`, `SKIP LOCKED`, `NOWAIT`) by ensuring they are safely ignored during translation to SQLite.

### DIFF
--- a/packages/mysql-on-sqlite/tests/WP_SQLite_Driver_Tests.php
+++ b/packages/mysql-on-sqlite/tests/WP_SQLite_Driver_Tests.php
@@ -6805,6 +6805,31 @@ END;
 		$this->assertSame( array(), (array) array_column( $result, 'id' ) );
 	}
 
+	public function testRowLeveLockingClauses() {
+		$this->assertQuery( 'CREATE TABLE t (name VARCHAR(255), value VARCHAR(255))' );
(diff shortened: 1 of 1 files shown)


tests/WP_SQLite_Driver_Tests.php
```diff
@@ -7380,6 +7380,22 @@ public function testUpdateWithJoinComplexQuery(): void {
 		);
 	}
 
+	public function testLockingClauses() {
+		$this->engine->query( "INSERT INTO _options (option_name, option_value) VALUES ('test_lock', '1')" );
```
JanJakes: As in other tests, we should use `assertQuery` to assert that no errors were triggered. It's also better to create a new table for the test (I consider these pre-created tables to be legacy, and would like to get rid of them).

JanJakes (commented):
Thanks for your contribution! I only have some small notes, and it needs a rebase to resolve conflict.

Additionally, I will update the title and description to accurately reflect that this adds test coverage.

Thanks!

JanJakes:
I went ahead and rebased + updated the tests myself. Also added tests for other mentioned statements like `NOWAIT`.

PR #343, opened 2026-03-30
[AST] Support INSERT/REPLACE without INTO and MySQL priority modifiers

wp-fuse:
Objective: Enhance the AST driver to support MySQL INSERT and REPLACE statements that omit the INTO keyword, and safely handle MySQL-specific priority modifiers that are unsupported in SQLite.

Changes:

Implemented robust INTO keyword injection in execute_insert_or_replace_statement.
Added logic to automatically skip unsupported MySQL priority modifiers (LOW_PRIORITY, HIGH_PRIORITY, DELAYED) which trigger syntax errors in SQLite.
Ensured that INTO is injected at the correct position relative to survive modifiers like OR IGNORE (translated from MySQL IGNORE).
Added semantic check to only apply OR IGNORE transformation on INSERT statements, not REPLACE.
Fixed a potential syntax issue in array_splice for better cross-version PHP compatibility.

Verification:

Validated via unit tests including simple INSERT without INTO, INSERT IGNORE without INTO, and INSERT LOW_PRIORITY ohne INTO.
All tests passing with WP_SQLITE_AST_DRIVER enabled.
Coding standards (CS) validated and fixed.

### DIFF
--- a/wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
+++ b/wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php
@@ -1830,15 +1830,27 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 	private function execute_insert_or_replace_statement( WP_Parser_Node $node ): void {
 		$parts                   = array();
 		$on_conflict_update_list = null;
+		$seen_into               = false;
+		$last_modifier_index     = -1;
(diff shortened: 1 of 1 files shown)


JanJakes (commented):
Thanks for catching the `INSERT` without `INTO` case!

This requires test coverage, and I believe it can be implemented in a much simpler way. Let me try...

PR #350, opened 2026-04-02
Rework release workflow

JanJakes:
## Summary

Replace the draft-release-triggered CI workflow with a local `prepare-release.sh` script and a simplified publish workflow.

GitHub does not fire `release` events for draft releases, so the `release-prepare` workflow could never trigger. The new approach uses a non-interactive script (`bin/prepare-release.sh <version>`) that bumps versions, generates a changelog from merged PRs, creates a release branch, and opens a PR. Merging the PR triggers the build and GitHub release automatically.

Also adds pre-release support: versions with a hyphen (e.g., `2.3.0-beta.1`) are marked as pre-release and skip WordPress.org deployment.


JanJakes:
All green. I will test this one on trunk.

PR #356, opened 2026-04-06
Translate MySQL CONVERT() expressions to SQLite

### LINKED ISSUE #344: CONVERT() queries fail with error
The Jetpack CRM plugin uses CONVERT() queries, which are valid MySQL queries, but they fail in the Playground.

The example query that works on MySQL, but not on SQLite, looks as follows:

```
SELECT CONVERT('Customer' USING utf8mb4) COLLATE utf8mb4_bin;
```

JanJakes:
## Summary

Fixes #344.

MySQL's `CONVERT()` function was passed through to SQLite unchanged, causing queries like `SELECT CONVERT('Customer' USING utf8mb4) COLLATE utf8mb4_bin` to fail with a syntax error.

This PR has two parts:

1. **Extract `simpleExprBody` as a named grammar rule.** The `%simpleExpr_factored` fragment is promoted to a real `simpleExprBody` rule so it creates its own AST node. This separates the core expression (CONVERT, CAST, literals, etc.) from trailing modifiers (COLLATE, CONCAT_PIPES) that remain in the parent `simpleExpr` node, making individual expression handlers simpler.

2. **Translate `CONVERT()` expressions.** Adds explicit handling for both forms of `CONVERT()` in the AST-based driver:
   - **`CONVERT(expr, type)`** is translated to `CAST(expr AS type)`, reusing the existing `castType` translation.
   - **`CONVERT(expr USING charset)`** is reduced to just the expression, as SQLite stores all text as UTF-8 and charset conversions are not needed.

## Test plan

- [x] Added `testConvert` translation test — verifies SQL-to-SQL translation for both CONVERT forms and COLLATE.
- [x] Added `testConvertExpression` — verifies CONVERT with type casting (BINARY, CHAR, SIGNED, UNSIGNED, DECIMAL, DATE).
- [x] Added `testConvertUsingExpression` — verifies CONVERT with charset conversion (utf8mb4, utf8, latin1).
- [x] Added `testConvertUsingWithCollate` — verifies the exact query from #344.
- [x] Added `testConvertWithColumnReferences` — verifies CONVERT with column references in SELECT, WHERE, and ORDER BY.
- [x] Full test suite passes (640 tests, 0 failures).
- [x] PHPCS passes.


JanJakes:
The Query Monitor tests failure is unrelated to this PR and I will address it separately. It seems that QM 4.0 is incompatbible with our integration.

bgrgicak (dismissed):

packages/mysql-on-sqlite/tests/WP_SQLite_Driver_Tests.php
```diff
@@ -9986,6 +9986,89 @@ public function testCastExpression(): void {
...
 	}
 
+	public function testConvertExpression(): void {
+		// CONVERT(expr, type) should behave like CAST(expr AS type).
+		$result = $this->assertQuery(
+			"SELECT
+				CONVERT('abc', BINARY) AS expr_1,
+				CONVERT('abc', CHAR) AS expr_2,
+				CONVERT('-10', SIGNED) AS expr_3,
+				CONVERT('-10', UNSIGNED) AS expr_4,
```
bgrgicak, line 9997: I have one question, in MySQL `SELECT CONVERT('-10', UNSIGNED);` will return `18446744073709551606`. 
Why is '-10' expected here?

JanJakes: @bgrgicak I missed this, thanks for catching that! It seems that the main problem is that SQLite doesn't support UNSIGNED integers, so it cannot store values as large as `18446744073709551606`.

That looks like a bigger problem, and I'm not sure whether it's reasonably solvable 😬 I guess for now, we should add a `TODO` comment explaining the issue.

bgrgicak: A `TODO` sounds good if we can't resolve it in this PR.

In this case, is it better to throw an error or silently return the incorrect value?

JanJakes: @bgrgicak There are likely more issues like this where we translate a MySQL construct to a similar one in SQLite, but they don't behave 100% the same. I'm not sure if `UNSIGNED` is the only incompatibility here, but it will affect not only `CONVERT` but especially `CAST` and potentially some arithmetic, etc.

Therefore, I think the TODO is fine (and I can also create a ticket), and we don't need to try to detect and fail the invalid cases at the moment. Fortunately, casting a signed negative integer to an unsigned one seems to be a very strange use case that is hopefully not encountered in reality.

JanJakes: I added the TODOs.

bgrgicak (approved):

packages/mysql-on-sqlite/tests/WP_SQLite_Driver_Tests.php
JanJakes: Issue: https://github.com/WordPress/sqlite-database-integration/issues/359

PR #357, opened 2026-04-08
Add Query Monitor 4.0 support

JanJakes:
## Summary

Query Monitor 4.0 switched from server-side PHP rendering to client-side Preact rendering inside a shadow DOM. This broke the existing SQLite query display integration, which overrides `output_query_row()` — a method that QM 4.0 no longer calls.

This PR adds QM 4.0 support while maintaining backward compatibility with QM 3.x:

- **Version detection** in `boot.php` using `QM_VERSION` to choose between QM 3.x and 4.0+ integration paths.
- **Custom QM collector and outputter** (`qm4.php`) that extracts SQLite queries from `$wpdb->queries`. With `$client_side_rendered = true`, QM auto-serializes the collector data into `window.QueryMonitorData.data.sqlite`, so the outputter's only job is to emit an inline JS module.
- **Shadow DOM injection JS** (`query-monitor-sqlite.js`) that reads from `QueryMonitorData` and injects `<details>` elements into the DB queries panel. It runs on `DOMContentLoaded` (after QM attaches the shadow root) and uses a debounced `MutationObserver` to re-inject after Preact re-renders (panel switches, filters, etc.). Each injected element tracks its SQL key to handle Preact's DOM recycling on filter/sort changes.
- **E2E tests** for both QM 3.x and 4.0+, each auto-skipping when the other version is detected.

Fixes the CI failure introduced by QM 4.0 release.

## Test Plan

- [x] E2E test passes with QM 4.0+ (latest): `<details>` elements appear in shadow DOM, SQLite queries display correctly
- [x] E2E test passes with QM 3.x (3.16.4): existing server-side integration works unchanged
- [x] PHPCS passes clean
- [x] CI end-to-end tests pass


packages/plugin-sqlite-database-integration/integrations/query-monitor/qm4.php
```diff
@@ -0,0 +1,91 @@
...
+			// Query Monitor skips queries with 'wp_admin_bar' in the stack.
+			if ( false !== strpos( $query[2] ?? '', 'wp_admin_bar' ) ) {
+				continue;
+			}
+			if ( ! empty( $query['sqlite_queries'] ) ) {
+				$sql            = trim( preg_replace( '/\s+/', ' ', $query[0] ) );
+				$mapped[ $sql ] = array_column( $query['sqlite_queries'], 'sql' );
+			}
+		}
+		$this->data->queries = $mapped;
```
ashfame, line 52: Nit: Setting `->queries` dynamically on `$this->data` could trigger a PHP 8.2+ "Creation of dynamic property" deprecation if QM 4.0's base data object isn't a plain `stdClass`. It may be worth defining a small `SQLite_QM_Data extends QM_Data` class with a typed `$queries` property to be safe.

packages/plugin-sqlite-database-integration/integrations/query-monitor/query-monitor-sqlite.js
```diff
@@ -0,0 +1,95 @@
...
+		}
+		inject( shadowRoot, sqliteData );
+
+		// Re-inject after Preact re-renders (panel switches, filters, etc.).
+		// Debounced to avoid excessive work during rapid DOM updates.
+		let timer;
+		new MutationObserver( () => {
+			clearTimeout( timer );
+			timer = setTimeout( () => inject( shadowRoot, sqliteData ), 100 );
+		} ).observe( shadowRoot, { childList: true, subtree: true } );
```
ashfame, line 34: Minor: The observer watches the entire shadow root subtree, so any QM panel interaction (not just DB queries) triggers a re-injection cycle. Since `inject()` early-exits when `#qm-db_queries` isn't found the overhead is minimal, but if the panel element is stable in the DOM, scoping the observer to it would cut down on unnecessary callbacks.

tests/e2e/specs/query-monitor-plugin.test.js
```diff
@@ -12,22 +12,169 @@ test.describe( 'Query Monitor plugin', () => {
...
+		await expect( async () => {
+			const counts = await container.evaluate( ( el ) => {
+				const panel = el.shadowRoot.getElementById( 'qm-db_queries' );
+				return {
+					rows: panel.querySelectorAll( 'tbody tr' ).length,
+					details: panel.querySelectorAll( 'details.qm-sqlite' ).length,
+				};
+			} );
+			expect( counts.rows ).toBeGreaterThan( 0 );
+			expect( counts.details ).toBe( counts.rows );
```
ashfame, line 134: Nit: This asserts that every visible row after the `get_option` filter has an SQLite `<details>` element. It holds today because all DB queries on SQLite carry `sqlite_queries` metadata, but it could become fragile if QM ever injects synthetic/internal rows into the table. Low risk — just flagging for awareness.

ashfame (approved):
Clean and well-architected PR. The QM 3.x/4.0 split is sound, the SQL-text-based matching is a smart approach for surviving Preact re-renders, and the E2E tests are thorough. Left a few minor notes inline — nothing blocking.

packages/plugin-sqlite-database-integration/integrations/query-monitor/query-monitor-sqlite.js
JanJakes: The observer needs to watch the full shadow root because `#qm-db_queries` doesn't exist until the user navigates to the Database Queries tab — scoping to it would miss its creation. The early-exit in `inject()` keeps the overhead negligible.

packages/plugin-sqlite-database-integration/integrations/query-monitor/qm4.php
JanJakes: Fixed in https://github.com/WordPress/sqlite-database-integration/pull/357/changes/991af33d16909ba4ccf77fa8af15cf5792345117.

tests/e2e/specs/query-monitor-plugin.test.js
JanJakes: I'll leave this one as-is at the moment. The test will fail if the markup changes.

johnbillion:
Oh I forgot to tell you that I've been thinking of adding native support for SQLite Database Integration to Query Monitor: https://github.com/johnbillion/query-monitor/pull/1090

JanJakes:
@johnbillion Oh, that is great! This PR is just a hotfix, and I wanted to look into a better solution eventually. Same for the `db.php` drop-in handling, where I'd like to propose some nicer integration at some point.

PR #358, opened 2026-04-08
Remove legacy SQLite driver

JanJakes:
## Summary

Remove the legacy, token-based MySQL-to-SQLite translator and the \`WP_SQLITE_AST_DRIVER\` feature flag. The AST-based driver is now the only driver.

### Changes

- Delete 6 legacy driver files (~8,700 lines) and 7 legacy test files (~5,000 lines), along with the root \`phpunit.xml.dist\` and \`composer.json\` test command.
- Remove \`WP_SQLITE_AST_DRIVER\` conditionals and \`WP_SQLite_Translator\` branches from \`class-wp-sqlite-db.php\`, \`admin-page.php\`, \`constants.php\`, \`wp-mysql-proxy.php\`, and \`phpcs.xml.dist\`.
- Remove legacy driver CI test step and unused \`phpunit-config\` workflow parameter.
- Remove DB_NAME fallback in \`db.php\` that was kept for legacy driver compatibility.
- Fix dead \`get_error_message()\` call in \`install-functions.php\` left over from the legacy driver API.
- Update docs to reflect the removal (\`AGENTS.md\`, \`README.md\`, \`readme.txt\`).

## Test plan

- [x] New driver tests pass (635 tests)
- [x] PHPCS clean
- [x] CI passes


joaopedrofrech:
This is amazing! are we finally moving to the new driver as default?

chubes4:
+1 to this! Would solve a problem for me when using https://github.com/Automattic/studio for local dev

Thanks for your efforts!

JanJakes:
> This is amazing! are we finally moving to the new driver as default?

@joaopedrofrech Yes, coming soon 🙂

packages/plugin-sqlite-database-integration/admin-page.php
```diff
@@ -155,8 +155,7 @@ function sqlite_plugin_adminbar_item( $admin_bar ) {
 	global $wpdb;
 
 	if ( defined( 'SQLITE_DB_DROPIN_VERSION' ) && defined( 'DB_ENGINE' ) && 'sqlite' === DB_ENGINE ) {
-		$suffix = defined( 'WP_SQLITE_AST_DRIVER' ) && WP_SQLITE_AST_DRIVER ? ' (AST)' : '';
-		$title  = '<span style="color:#46B450;">' . __( 'Database: SQLite', 'sqlite-database-integration' ) . $suffix . '</span>';
+		$title = '<span style="color:#46B450;">' . __( 'Database: SQLite', 'sqlite-database-integration' ) . '</span>';
```
ashfame, line 158: It would be nice for this to still say (AST) as suffix.

packages/plugin-sqlite-database-integration/readme.txt
```diff
@@ -15,8 +15,6 @@ SQLite integration plugin by the WordPress Team.
 
 The SQLite plugin is a community, feature plugin. The intent is to allow testing an SQLite integration with WordPress and gather feedback, with the goal of eventually landing it in WordPress core.
 
-This feature plugin includes code from the PHPMyAdmin project (specifically parts of the PHPMyAdmin/sql-parser library), licensed under the GPL v2 or later. More info on the PHPMyAdmin/sql-parser library can be found on [GitHub](https://github.com/phpmyadmin/sql-parser).
```
ashfame, line 18: This is a general correction, not related to this PR, right?

packages/plugin-sqlite-database-integration/wp-includes/sqlite/db.php
```diff
@@ -47,33 +47,11 @@
...
- * There are some existing projects in which the DB_NAME constant is missing in
- * wp-config.php. To enable easier early adoption and testing of the new SQLite
- * driver, let's allow using a default database name when DB_NAME is not set.
- *
- * TODO: For version 3.0, enforce the DB_NAME constant and remove the fallback.
- */
-if ( defined( 'DB_NAME' ) && '' !== DB_NAME ) {
-	$db_name = DB_NAME;
-} else {
-	$db_name = apply_filters( 'wp_sqlite_default_db_name', 'database_name_here' );
```
ashfame, line 75: I think we should consider using [`apply_filters_deprecated()`](https://developer.wordpress.org/reference/functions/apply_filters_deprecated/) with an empty replacement and a message explaining the "why" instead of silently failing.

> apply_filters_deprecated() internally calls _deprecated_hook(), which in turn calls trigger_error() with E_USER_DEPRECATED.

packages/plugin-sqlite-database-integration/wp-includes/sqlite/db.php
```diff
@@ -47,33 +47,11 @@
...
-	require_once __DIR__ . '/../database/load.php';
-} else {
-	require_once __DIR__ . '/php-polyfills.php';
-	require_once __DIR__ . '/class-wp-sqlite-lexer.php';
-	require_once __DIR__ . '/class-wp-sqlite-query-rewriter.php';
-	require_once __DIR__ . '/class-wp-sqlite-translator.php';
-	require_once __DIR__ . '/class-wp-sqlite-token.php';
-	require_once __DIR__ . '/class-wp-sqlite-pdo-user-defined-functions.php';
-}
+require_once __DIR__ . '/../database/load.php';
```
ashfame, line 50: Nit: In `/../` cases, the parent directory probably has significance and is usually better to define in a constant and then used in the path.

ashfame (commented):
Looks great, but left a few comments and made a tiny PR for consideration - #362

packages/plugin-sqlite-database-integration/admin-page.php
JanJakes: Hmm, the problem is I think most users won't know what that means. It made some sense when it was a feature flag, but otherwise, I'd say it's an implementation detail of the parser.

packages/plugin-sqlite-database-integration/readme.txt
JanJakes: It's related, actually. The legacy driver used a simple lexer/parser extracted from PHPMyAdmin. The new driver doesn't use any of this.

packages/plugin-sqlite-database-integration/wp-includes/sqlite/db.php
JanJakes: Oh, this was actually a forgotten temporary fix for Studio. It's not an old filter or anything that would be used originally; it was added only for the new driver as a quick fix until it was superseded by a better solution and forgotten. It had no effect on the old driver.

packages/plugin-sqlite-database-integration/wp-includes/sqlite/db.php
JanJakes: Good point, it actually hints at what the directory structure looks like, and I think it will need a follow-up PR. It looks like this:
```bash
plugin-sqlite-database-integration/
  - wp-includes/   # This mirros WP structure for possible WP Core PR
    - database/    # This is a symlink to the new driver package sources
    - sqlite/      # This has plugin-specific files that are needed even with the new driver.
                   # I think a follow-up PR should move out of wp-includes, or place them
                   # wherever it will make sense. There is also chance for further cleanup.
```

packages/plugin-sqlite-database-integration/wp-includes/sqlite/db.php
ashfame: I think as a general rule, if something gets added to public api, we should treat the depreciation gracefully. But if you are sure, this couldn't have been used by anyone other than Studio, totally fine to make an exception to that rule.

packages/plugin-sqlite-database-integration/readme.txt
ashfame: I see.

packages/plugin-sqlite-database-integration/admin-page.php
ashfame: I agree, it's an implementation detail. Users who were not using AST would now be using it in future with the newer version of plugin as soon as they update, right? So with that context in mind and personally, as a user I would like to see it.
But feel free to choose what you think is best.

ashfame (approved):
Great work! 🚀

packages/plugin-sqlite-database-integration/wp-includes/sqlite/db.php
wojtekn: Actually, it breaks creating new sites in Studio - for a new site, I get “Error establishing a database connection” until I manually restart the site.

PR #362, opened 2026-04-15
Fix/legacy driver removal followup

ashfame:
Follow up items for consideration from reviewing #358


JanJakes (approved):
Thanks for the fixes!

PR #363, opened 2026-04-17
Fix `RAND()` function behavior

JanJakes:
## Summary

Replaces the old `mt_rand(0, 1)` stub — which returned an integer 0 or 1, not a float in `[0, 1)` as MySQL requires — with MySQL-compatible `RAND()` behavior.

Replaces #341.

## Unseeded `RAND()`

Compiles to a native SQLite expression:

```sql
((RANDOM() & 0x001FFFFFFFFFFFFF) / 9007199254740992.0)
```

The 53-bit mask matches the IEEE 754 double mantissa, so division by 2^53 is exact and strictly less than 1.0. Matches MySQL, where unseeded `RAND()` uses a thread-level state independent of `RAND(N)`.

## Seeded `RAND(N)`

Routes through a PHP UDF implementing MySQL's exact LCG from `my_rnd_init()` / `my_rnd()` (`sql/item_func.cc`, `mysys/my_rnd.cc`), bit-exact against MySQL 9.6. Requires 64-bit PHP.

Seed handling matches `val_int()`: `NULL` becomes 0, floats round to nearest (`RAND(3.9) == RAND(4)`), numeric strings follow the same path.

## Known divergences

- MySQL distinguishes constant vs non-constant seeds at parse time (constant = init once per statement, non-constant = reinit per row). A SQLite UDF can't see the expression, so we approximate by reinitializing only when the seed value changes. This diverges when a non-constant expression yields a stable value.
- The UDF keeps one LCG state per connection, so multiple `RAND(N)` call sites in one query share a stream: `SELECT RAND(1), RAND(1)` returns `(v1, v2)` here vs `(v, v)` in MySQL.

Both are documented in the `rand()` docblock.

## Metadata

`RAND()` column metadata is now reported as `DOUBLE` / `PARAM_STR`, removing a long-standing TODO.

## Test plan
- [ ] CI green.
- [ ] Bit-exact reference values against MySQL 9.6 (seeds 0, 1, 3, 5, and multi-row sequences).
- [ ] NULL, float, numeric/non-numeric string, and negative seed handling.
- [ ] `RAND()` in `WHERE`, `UPDATE`, `INSERT`, `ORDER BY` (`LIMIT` + seeded deterministic permutation).
- [ ] Per-statement flush contract inside a transaction.

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
@@ -4418,7 +4418,7 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 				 * The UDF also handles RAND(NULL) as RAND(0), matching MySQL.
 				 */
 				if ( 0 === count( $args ) ) {
-					return '((RANDOM() & 0x001FFFFFFFFFFFFF) / 9007199254740992.0)';
+					return '((RANDOM() & ((1 << 53) - 1)) / ((1 << 53) * 1.0))';
 				}
 				return $this->translate_sequence( $node->get_children() );
 			case 'DATE_FORMAT':
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
@@ -437,6 +437,13 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 	private $connection;
 
 	/**
+	 * User-defined functions registered on the SQLite connection.
+	 *
+	 * @var WP_SQLite_PDO_User_Defined_Functions
+	 */
+	private $user_defined_functions;
+
+	/**
 	 * A service for managing MySQL INFORMATION_SCHEMA tables in SQLite.
 	 *
 	 * @var WP_SQLite_Information_Schema_Builder
@@ -724,7 +731,7 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 		$this->connection->query( 'PRAGMA foreign_keys = ON' );
 
 		// Register SQLite functions.
-		WP_SQLite_PDO_User_Defined_Functions::register_for( $this->connection->get_pdo() );
+		$this->user_defined_functions = WP_SQLite_PDO_User_Defined_Functions::register_for( $this->connection->get_pdo() );
 
 		// Load MySQL grammar.
 		if ( null === self::$mysql_grammar ) {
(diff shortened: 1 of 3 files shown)


packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
```diff
@@ -167,19 +200,74 @@ public function md5( $field ) {
...
+	 *
+	 * @param int|float|string|null $seed Seed value.
+	 *
+	 * @return float A value in [0, 1).
 	 */
-	public function rand() {
-		return mt_rand( 0, 1 );
+	public function rand( $seed ) {
+		// Requires 64-bit PHP. Seed * 0x10000001 can exceed PHP_INT_MAX on 32-bit.
+		$max_value = 0x3FFFFFFF;
```
zaerl, line 234: I really appreciate this defensive polish we have here. Despite the rarity of 32-bit systems, it is always better to handle them.

packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -4395,6 +4402,25 @@ private function translate_function_call( WP_Parser_Node $node ): string {
...
+				 * 53 bits (IEEE 754 double mantissa width) and dividing by 2^53.
+				 * This avoids the edge case where masking to 63 bits and dividing
+				 * by 2^63 could round to 1.0 due to loss of precision in double.
+				 *
+				 * Seeded RAND(N) falls through to the UDF, which implements
+				 * MySQL's deterministic LCG (Linear Congruential Generator).
+				 * The UDF also handles RAND(NULL) as RAND(0), matching MySQL.
+				 */
+				if ( 0 === count( $args ) ) {
+					return '((RANDOM() & 0x001FFFFFFFFFFFFF) / 9007199254740992.0)';
```
zaerl: Having some of these magic numbers declared at the class level can be useful to clean up the code a little. But feel free to ignore this; the comment above already explains it well.

Unfortunately, PHP has no constant for `2^53 - 1`, which instead is `Number.MAX_SAFE_INTEGER` in Javascript.

zaerl (approved):
> In MySQL, unseeded RAND() uses the thread-level random state

I didn't know about this, pretty strange to me.

Thanks for the great work here. I've added some comments, but feel free to ignore everything and ship this.

packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
JanJakes: @zaerl Good point. Since we're in an SQL string here, I didn't make them PHP constants, but I did make them more readable in https://github.com/WordPress/sqlite-database-integration/pull/363/changes/409a43c5e9fed66d63f318f3adce4010cb064bbc.

PR #365, opened 2026-04-17
Add support for `DELETE` with `LIMIT` and `ORDER BY`

### LINKED ISSUE #100: Support DELETE LIMIT
LIMIT has been fixed for UPDATE with this PR https://github.com/WordPress/sqlite-database-integration/pull/93 but DELETE LIMIT is still unsupported.

JanJakes:
## Summary

Adds support for single-table `DELETE` statements with `ORDER BY` and/or `LIMIT` clauses, which SQLite rejects as a syntax error unless it was compiled with `SQLITE_ENABLE_UPDATE_DELETE_LIMIT` (not the case for standard builds, including the one bundled with WordPress Playground).

The fix mirrors the existing `UPDATE` LIMIT rewrite: when `ORDER BY` or `LIMIT` is present, the statement is translated to drive deletion off a `rowid` subquery.

```sql
-- MySQL
DELETE FROM t WHERE c = 1 ORDER BY c LIMIT 10

-- Translated SQLite
DELETE FROM `t` WHERE rowid IN (
    SELECT rowid FROM `t` WHERE `c` = 1 ORDER BY `c` ASC LIMIT 10
)
```

The subquery forwards `tableRef`, `tableAlias`, `whereClause`, `orderClause`, and `simpleLimitClause`, so queries that reference an aliased table keep the alias in scope inside the subquery (e.g. `DELETE FROM t AS a WHERE a.c = 1 LIMIT 1` works).

## Scope

- **In scope:** single-table `DELETE` with `ORDER BY`, `LIMIT`, or both — with or without `WHERE` and alias.
- **Out of scope:** multi-table `DELETE` — MySQL's grammar disallows `ORDER BY`/`LIMIT` there, so there's nothing to add.

## Tests

- Translation tests covering plain `LIMIT`, `WHERE` + `LIMIT`, `ORDER BY` + `LIMIT`, `WHERE` + `ORDER BY` + `LIMIT`, and alias + `WHERE` + `LIMIT`.
- Runtime tests that insert rows, run `DELETE ... LIMIT` through the driver, and assert on the surviving rows — including a case that matches multiple rows to verify `LIMIT` is actually enforced, and an alias case.

Fixes #100

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
@@ -2308,6 +2308,38 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 			throw $this->new_access_denied_to_information_schema_exception();
 		}
 
+		/*
+		 * SQLite doesn't support DELETE with ORDER BY/LIMIT.
+		 * We need to use a subquery to emulate this behavior.
+		 *
+		 * For instance, the following query:
+		 *   DELETE FROM t WHERE c = 2 LIMIT 1;
+		 * Will be rewritten to:
+		 *   DELETE FROM t WHERE rowid IN ( SELECT rowid FROM t WHERE c = 2 LIMIT 1 );
+		 */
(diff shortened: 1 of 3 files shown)


ashfame (approved):
Looks good!

PR #366, opened 2026-04-20
Add support for `NO_AUTO_VALUE_ON_ZERO` SQL mode

JanJakes:
## Summary

Emulates MySQL's `NO_AUTO_VALUE_ON_ZERO` SQL mode behavior on SQLite.

MySQL treats a literal `0` in an `AUTO_INCREMENT` column the same as `NULL` and generates the next sequence value. This can be suppressed by the `NO_AUTO_VALUE_ON_ZERO` SQL mode, which is **not** part of MySQL's default modes.

Some legacy WordPress versions rely on the default behavior — e.g. WP 1.0's `post.php` emits:

```sql
INSERT INTO wp_posts (ID, post_author, post_title, post_status)
VALUES ('0', '1', 'Hello', 'publish');
```

On MySQL this stores `ID = 1`. The current SQLite driver stored `ID = 0`, which fails with legacy WP.

## Scope

- **Affects:** `INSERT` / `REPLACE` value lists (including `VALUES`, `SET`, and `SELECT` forms).
- **Does not affect:** `UPDATE` (and the UPDATE half of `INSERT ... ON DUPLICATE KEY UPDATE`). MySQL never auto-generates `AUTO_INCREMENT` values on UPDATE — zeros are stored literally.

The default `sql_mode` was already correct — `NO_AUTO_VALUE_ON_ZERO` is absent, matching MySQL 8.0 defaults.

## Tests

- `testDefaultSqlModeDoesNotIncludeNoAutoValueOnZero` — asserts the default modes.
- `testAutoIncrementZeroAdvancesSequenceByDefault` — `0`, `'0'`, and `NULL` all generate new ids.
- `testAutoIncrementZeroAdvancesSequenceForAllInsertShapes` — covers `INSERT ... SET`, `INSERT ... SELECT`, and `REPLACE`.
- `testNoAutoValueOnZeroSqlMode` — with the mode on, literal `0` is stored as-is; only `NULL` advances the sequence.

See: https://dev.mysql.com/doc/refman/8.4/en/sql-mode.html#sqlmode_no_auto_value_on_zero

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
@@ -5351,9 +5351,24 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 				$fragment .= null === $default ? 'NULL' : $this->quote_sqlite_value( $default );
 			} else {
 				// When a column value is included, we need to apply type casting.
-				$position   = array_search( $column['COLUMN_NAME'], $insert_list, true );
-				$identifier = $this->quote_sqlite_identifier( $select_list[ $position ] );
-				$value      = $this->cast_value_for_saving( $column['DATA_TYPE'], $identifier );
+				$position          = array_search( $column['COLUMN_NAME'], $insert_list, true );
+				$identifier        = $this->quote_sqlite_identifier( $select_list[ $position ] );
+				$value             = $this->cast_value_for_saving( $column['DATA_TYPE'], $identifier );
+				$is_auto_increment = 'auto_increment' === $column['EXTRA'];
+
+				/*
+				 * In MySQL, inserting 0 into an AUTO_INCREMENT column increments
+				 * the sequence, unless the NO_AUTO_VALUE_ON_ZERO SQL mode is set.
+				 *
+				 * In SQLite, we need to rewrite 0 to NULL to advance the sequence.
+				 * The value is cast to INTEGER before the comparison, because
+				 * SQLite treats values of different types as unequal (0 != '0').
+				 *
+				 * See: https://dev.mysql.com/doc/refman/8.4/en/sql-mode.html#sqlmode_no_auto_value_on_zero
+				 */
+				if ( $is_auto_increment && ! $this->is_sql_mode_active( 'NO_AUTO_VALUE_ON_ZERO' ) ) {
+					$value = sprintf( 'NULLIF(CAST(%s AS INTEGER), 0)', $value );
+				}
 
 				/*
 				 * In MySQL non-STRICT mode, when inserting from a SELECT query:
@@ -5361,9 +5376,17 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 				 * When a column is declared as NOT NULL, inserting a NULL value
 				 * saves an IMPLICIT DEFAULT value instead. This behavior only
 				 * applies to the INSERT ... SELECT syntax (not VALUES or SET).
+				 *
+				 * AUTO_INCREMENT columns are excluded. A NULL value advances
+				 * the sequence regardless of the column's nullability.
 				 */
 				$is_insert_from_select = 'insertQueryExpression' === $node->rule_name;
-				if ( ! $is_strict_mode && $is_insert_from_select && 'NO' === $column['IS_NULLABLE'] ) {
+				if (
+					! $is_strict_mode
+					&& ! $is_auto_increment
+					&& $is_insert_from_select
+					&& 'NO' === $column['IS_NULLABLE']
+				) {
 					$implicit_default = self::DATA_TYPE_IMPLICIT_DEFAULT_MAP[ $column['DATA_TYPE'] ] ?? null;
 					if ( null !== $implicit_default ) {
 						$value = sprintf( 'COALESCE(%s, %s)', $value, $this->quote_sqlite_value( $implicit_default ) );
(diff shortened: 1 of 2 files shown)


packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -5132,19 +5132,42 @@ function ( $column ) use ( $is_strict_mode, $insert_map ) {
 				$fragment .= null === $default ? 'NULL' : $this->quote_sqlite_value( $default );
 			} else {
 				// When a column value is included, we need to apply type casting.
-				$position   = array_search( $column['COLUMN_NAME'], $insert_list, true );
-				$identifier = $this->quote_sqlite_identifier( $select_list[ $position ] );
-				$value      = $this->cast_value_for_saving( $column['DATA_TYPE'], $identifier );
+				$position          = array_search( $column['COLUMN_NAME'], $insert_list, true );
+				$identifier        = $this->quote_sqlite_identifier( $select_list[ $position ] );
+				$value             = $this->cast_value_for_saving( $column['DATA_TYPE'], $identifier );
+				$is_auto_increment = 'auto_increment' === $column['EXTRA'];
```
brandonpayton: @JanJakes why is this an equality check when in other places we are doing a str_contains() check like:
https://github.com/WordPress/sqlite-database-integration/blob/de8f9b0b5b36a2016fab34e5f069aa1b5ba28efd/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php#L5274

packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -5132,19 +5132,42 @@ function ( $column ) use ( $is_strict_mode, $insert_map ) {
...
+				 * the sequence, unless the NO_AUTO_VALUE_ON_ZERO SQL mode is set.
+				 *
+				 * In SQLite, we need to rewrite 0 to NULL to advance the sequence.
+				 * The value is cast to INTEGER before the comparison, because
+				 * SQLite treats values of different types as unequal (0 != '0').
+				 *
+				 * See: https://dev.mysql.com/doc/refman/8.4/en/sql-mode.html#sqlmode_no_auto_value_on_zero
+				 */
+				if ( $is_auto_increment && ! $this->is_sql_mode_active( 'NO_AUTO_VALUE_ON_ZERO' ) ) {
+					$value = sprintf( 'NULLIF(CAST(%s AS INTEGER), 0)', $value );
```
brandonpayton, line 5370: Should the `$value` be escaped any differently. Naively, it seems like this could be used for SQL injection. Is this safe because it is a specific value from parser output?

brandonpayton (approved):
@JanJakes these changes look good in general and make sense to me. I left a couple of concerns at least partially due to my inexperience with this project. Please merge as-is if you do not think they are issues.

packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
JanJakes: Good catch — switched to `str_contains` in 8b82485a95a2d89796a1fd7fe0ff11b96db3334f for consistency with the filter at line 5055 right above (which uses the same predicate to drop omitted auto_increment columns). Functionally equivalent today since `get_column_extra()` returns exactly `'auto_increment'` via early-return, but uniform with the adjacent code and future-proof if the EXTRA composition ever changes.

packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
JanJakes: Yes, safe by construction. `$value` here is built from `$identifier`, which is `quote_sqlite_identifier( $select_list[ $position ] )` — a backtick-quoted SQLite identifier from `$select_list`. That list is either positional `column1`, `column2`, … names for the VALUES path (line 5102) or column names returned by SQLite itself for the `INSERT … SELECT` path (line 5096, via `getColumnMeta`). Neither comes from user input.

The user-supplied literals from the original VALUES clause live inside the inner subquery (`FROM (VALUES (…))` at line 5210), where they go through the regular translation pipeline that quotes/escapes them. Our outer `SELECT` only references the wrapped identifiers, never raw user values, so wrapping `$value` in `NULLIF(CAST(…), 0)` cannot introduce injection.

packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
JanJakes: Ehh, Claude went ahead here without my confirmation. In any case, both `str_contains` and `===` are safe today (`auto_increment` is never combined with other "extra" fields), but for general consistency and future-proofness, I unified all these checks to `str_contains` in 8b82485a95a2d89796a1fd7fe0ff11b96db3334f.

packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
JanJakes: Claude went ahead with this cryptic comment without my confirmation. Here's how to say it in human words: `$value` comes from `cast_value_for_saving()` which returns an already translated and fully escaped value.

Issue #140, opened 2024-07-29
Add `AUTO_INCREMENT` sequence to `show create` statements


jeroenpf:
When running the `show create table` query, the returned create statement does not contain the sequence number for the auto-increment column. It would be good to add this. It can be obtained from the `sqlite_sequence` table.

Here is an example of a create statement that has the current value for the auto increment column:

```sql
CREATE TABLE `wp_options` (
    `option_id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
     `option_name` varchar(191) NOT NULL DEFAULT '',
     `option_value` longtext NOT NULL,
     `autoload` varchar(20) NOT NULL DEFAULT 'yes',
      PRIMARY KEY (`option_id`),
      UNIQUE KEY `option_name` (`option_name`),
      KEY `autoload` (`autoload`)
) ENGINE=InnoDB AUTO_INCREMENT=313 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
```

And here is one that the plugin returns:

```sql
DROP TABLE IF EXISTS `wp_options`;
CREATE TABLE `wp_options` (
	`option_id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
	`option_name` varchar(191) NOT NULL DEFAULT '',
	`option_value` longtext NOT NULL DEFAULT '',
	`autoload` varchar(20) NOT NULL DEFAULT 'yes',
	PRIMARY KEY (`option_id`),
	KEY `autoload` (`autoload`),
	UNIQUE KEY `option_name` (`option_name`)
);
```

JanJakes:
Implemented in https://github.com/WordPress/sqlite-database-integration/pull/367.

PR #367, opened 2026-04-20
Add support for `AUTO_INCREMENT` value management

### LINKED ISSUE #114: Support `ALTER TABLE x AUTO_INCREMENT = 100`
Support for adjusting the autoincrement counter would be useful in Playground for resetting the default WordPress content.

Right now you may get 3 or 4 or 5 pages by default, depending on a WordPress release and you can't easily create a page and redirect to, say, `/wp-admin/post.php?post=4&action=edit` because the ID won't be reliably 4. Maybe it will be 3, maybe 5, or maybe 6.

JanJakes:
## Summary

Adds support for **reporting** and **adjusting** `AUTO_INCREMENT` counters, which the driver previously ignored.

### Reporting `AUTO_INCREMENT` values (commit 1)

`SHOW TABLE STATUS`, `SHOW CREATE TABLE`, and `INFORMATION_SCHEMA.TABLES` now compute the `Auto_increment` value live from `sqlite_sequence` rather than returning `NULL`. The counter is preserved past `DELETE` and reset by `TRUNCATE`, matching InnoDB semantics.

### Adjusting `AUTO_INCREMENT` values (commit 2)

Three MySQL forms are now honored:

1. `CREATE TABLE t (...) AUTO_INCREMENT = N` — seeds the counter.
2. `ALTER TABLE t AUTO_INCREMENT = N` — instant; no table rebuild.
3. `ALTER TABLE t <actions>, AUTO_INCREMENT = N` — applied alongside other alter items.

The requested value is written as `N - 1` into `sqlite_sequence` (MySQL stores the next value, SQLite the last used), clamped to `MAX(col)` so the effective counter never drops below an existing row — matching MySQL. When the table has no `AUTO_INCREMENT` column, the option is silently ignored, matching MySQL.

**Not covered:** `SET @@auto_increment_increment` / `SET @@auto_increment_offset`. These change the *step* and *offset* of generated values, not the counter, and SQLite's `AUTOINCREMENT` always steps by 1 — so emulating non-1 increments would require generating IDs in PHP for every insert. Out of scope here.

## Notes

- `sqlite_sequence` has no `UNIQUE` constraint on `name`, so native upsert is unavailable; the row is maintained with `UPDATE`-then-`INSERT` (second query only runs when the first matches nothing).
- The clamp expression is inlined as an integer literal rather than a bound parameter. PDO binds PHP ints as `TEXT`, and SQLite's type affinity ranks `TEXT` above `INTEGER` in `MAX()`, which would silently produce wrong results.
- `ALTER TABLE` with *only* table options (no `alterListItem`) now skips the unnecessary full-table copy.

Fixes #114 and #140.

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
@@ -2958,41 +2958,64 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 		// LIKE and WHERE clauses.
 		$like_or_where = $node->get_first_child_node( 'likeOrWhere' );
 		if ( null !== $like_or_where ) {
-			$condition = $this->translate_show_like_or_where_condition( $like_or_where, 'table_name' );
+			$condition = $this->translate_show_like_or_where_condition( $like_or_where, 'Name' );
 		}
 
-		// Fetch table information.
-		$tables_tables = $this->information_schema_builder->get_table_name(
-			false, // SHOW TABLE STATUS lists only non-temporary tables.
-			'tables'
+		// SHOW TABLE STATUS lists only non-temporary tables.
+		$tables_table  = $this->information_schema_builder->get_table_name( false, 'tables' );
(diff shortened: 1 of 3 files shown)


mho22 (approved):
Well done!

Issue #204, opened 2025-07-03
Run tests against Turso DB


JanJakes:
When [Turso DB](https://turso.tech/) is more feature-complete, let's run the SQLite driver test suite also against Turso DB to evaluate the compatibility.

See:
- https://turso.tech/blog/turso-the-next-evolution-of-sqlite
- https://github.com/tursodatabase/turso
- https://github.com/tursodatabase/turso-client-php

PR #370, opened 2026-04-23
Add PHPUnit tests workflow for Turso DB

JanJakes:
## Summary

Adds a CI job that runs the SQLite driver unit tests (`packages/mysql-on-sqlite`) against [Turso DB](https://github.com/tursodatabase/turso), a Rust reimplementation of SQLite.

The workflow:

- Pins Turso `main` at commit `83d0f02f567f54ed9529f0b3955f294d3836a134` for reproducible builds.
- Applies the Turso source patches listed below.
- Builds the `turso_sqlite3` crate as a shared library and verifies its exported SQLite C API symbols.
- Builds and preloads a compatibility shim for SQLite C API symbols and formatting functions required by PHP's `pdo_sqlite`.
- Applies localized, workflow-only driver and test rewrites for known Turso compatibility differences.
- Runs focused smoke/probe coverage before PHPUnit.
- Runs each PHPUnit class in a fresh PHP process, aggregates the JUnit results, and fails on missing output, parse errors, test errors, or failures.

## Turso source patches

The workflow applies these patches in order from `.github/turso-patches/`:

1. `01-stub-macro.py` — return neutral values from unimplemented C API stubs instead of aborting PHP.
2. `02-column-functions-null-row.py` — return SQLite-compatible null values when column accessors are called without a current row.
3. `03-create-function-v2-skip-old-destroy.py` — avoid invoking stale UDF destroy callbacks when global function slots are reused.
4. `04-finalize-try-lock-on-gc-reentry.py` — avoid finalize deadlocks during PHP GC re-entry from UDF callbacks.
5. `05-max-custom-funcs-32-to-64.py` — expand the C binding's function bridge capacity for the driver's 44 UDFs.
6. `06-collation-mysql-aliases.py` — map common MySQL collation names to compatible Turso built-ins.
7. `07-collate-direct-column-refs-only.py` — limit implicit column collation inheritance to direct column references.
8. `08-create-trigger-preserve-original-sql.py` — preserve the original CREATE TRIGGER SQL text in `sqlite_master`.
9. `09-hash-join-correlated-subquery.py` — prevent incorrect hash joins in correlated query blocks.

The previous `sqlite_sequence` database-selection patch is no longer included because Turso fixed it upstream in [8f7d31f](https://github.com/tursodatabase/turso/commit/8f7d31f33059b96d42a48582936d9388785a4af2).

Refs: https://github.com/WordPress/sqlite-database-integration/issues/204

## Test plan

- [x] All PR checks pass.
- [x] All nine Turso source patches apply successfully.
- [x] `turso_sqlite3` builds successfully from the pinned source.
- [x] The shared library exposes the expected `sqlite3_*` symbols.
- [x] The `pdo_sqlite` smoke test and Turso compatibility probes pass.
- [x] PHPUnit completes with 652/669 passing, 17 skipped, and zero errors or failures.


JanJakes:
Closing this for now. We can reopen when needed. Turso DB is more behind SQLite than I thought. It's improving, but many patches are still needed. We can re-evaluate in the future.

PR #371, opened 2026-04-24
Add PHPUnit tests workflow for Doltlite

JanJakes:
## Summary

Runs the mysql-on-sqlite PHPUnit suite against [Doltlite](https://github.com/dolthub/doltlite), pinned to [`0745bbab`](https://github.com/dolthub/doltlite/commit/0745bbab29778393780b220d2a91fc5fa753853f). The full suite passes with **740 tests, 1,428,287 assertions, 15 skipped, 2 incomplete**.

The workflow builds Doltlite from source, installs it as the system `libsqlite3` so PHP's `pdo_sqlite` resolves against it, verifies that Doltlite is active, and runs PHPUnit without filters. Installing the library replaces the earlier `LD_PRELOAD` approach, which loses to `pdo_sqlite`'s `DT_NEEDED` binding on `libsqlite3.so.0`.

Three narrow, CI-only compatibility patches remain:

1. **Keep mysql-on-sqlite tables as rowid tables.** Doltlite auto-converts ordinary composite or text primary-key tables to `WITHOUT ROWID`, while the driver still relies on rowid access for table recreation, limited UPDATE/DELETE rewrites, on-update triggers, zero-column result statements, and information-schema ordering.
2. **Preserve parseable schema SQL across rollback.** Doltlite's schema canonicalization now strips line and block comments outside quoted strings, preventing collapsed `--` comments from consuming the rest of a restored `CREATE TABLE` statement.
3. **Expose configurable logical journal modes.** Doltlite's storage engine does not use SQLite journal or WAL sidecars, but the pager shim tracks the requested logical mode and preserves stock SQLite failure behavior when WAL cannot be opened.

File-based tests also remove Doltlite's persistent per-database lock file during cleanup, alongside standard SQLite sidecars.

Upstreaming the remaining Doltlite compatibility changes is a separate task.

## Test plan

- [x] `PHPUnit Tests (Doltlite)` passes: `740 tests, 1,428,287 assertions, 15 skipped, 2 incomplete`.
- [x] The full PR check suite passes.
- [x] Local unprivileged Linux verification against patched Doltlite passes with the same counts.
- [x] Local stock-SQLite PHPUnit suite and coding-standards checks pass.

### DIFF
--- a/.github/doltlite-patches/canonicalize-schema-sql-comments.patch
+++ b/.github/doltlite-patches/canonicalize-schema-sql-comments.patch
@@ -1,0 +1,47 @@
+Strip SQL comments while canonicalizing sqlite_schema SQL
+
+Doltlite canonicalizes schema SQL by collapsing whitespace before storing
+sqlite_schema rows after rollback and restore paths. When the SQL contains
+line comments, collapsing newlines while preserving `--` makes the comment
+consume the rest of the CREATE statement. SQLite then reports a malformed
+schema with incomplete input.
+
+Skip line and block comments outside quoted strings during schema SQL
+canonicalization. This keeps canonicalized schema text parseable while
+preserving the existing whitespace normalization behavior.
+
+diff --git a/src/doltlite_hashof.c b/src/doltlite_hashof.c
+index 1563b18..7f8e073 100644
+--- a/src/doltlite_hashof.c
++++ b/src/doltlite_hashof.c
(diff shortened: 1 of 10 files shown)


JanJakes:
Closing this for now. We can reopen when needed. Doltlite compatibility with SQLite is pretty good and improving. If needed, it is viable to be used with this project.

PR #373, opened 2026-04-25
Parser performance: 2-3x speedup via grammar preprocessing and interpreter optimisations

JanJakes:
## Summary

This is a draft PR to measure CI runtime improvement from a series of parser-performance optimisations on the recursive descent parser. Aimed primarily at lowering the time tests spend in PHPUnit, which is dominated by the 70-batch MySQL server parser test suite (~70k queries parsed per run).

End-to-end (lex+parse, 69,577 queries, local benchmark):
- Trunk, no opcache:        ~10,500 QPS  (6.6 s)
- Trunk, tracing JIT:       ~23,700 QPS  (2.9 s)
- This branch, no opcache:  **~25,100 QPS  (2.8 s)**  — 2.4× faster
- This branch, tracing JIT: **~43,500 QPS  (1.6 s)**  — 1.8× faster end-to-end / 2.7× parse-only

## CI runtime comparison

`Run PHPUnit tests` step duration on this branch vs the average of the last 3 trunk-merged feature PRs (#365 / #367 / #369):

| PHP   | Trunk avg | This PR  | Speedup |
|-------|-----------|----------|---------|
| 7.2   | 165 s     | **97 s**  | 1.70×   |
| 7.3   | 126 s     | **73 s**  | 1.73×   |
| 7.4   | 120 s     | **57 s**  | 2.11×   |
| 8.0   | 132 s     | **68 s**  | 1.94×   |
| 8.1   | 141 s     | **75 s**  | 1.88×   |
| 8.2   | 128 s     | **70 s**  | 1.83×   |
| 8.3   | 134 s     | **74 s**  | 1.81×   |
| 8.4   | 138 s     | **72 s**  | 1.92×   |
| 8.5   | 141 s     | **79 s**  | 1.78×   |

Total: **20.4 min → 11.1 min** of cumulative phpunit-step time across the matrix per run (~9 min saved per CI run).

The 70/70 server-suite parser test sub-suite is what dominates the step. Per-PHPUnit-step speedup is consistent with the local end-to-end benchmark.

## Sequence of commits

1. `f935d03` Inline terminal matching and defer node allocation
2. `0e34f25` Per-branch FIRST sets to skip unreachable branches
3. `6df347a` Short-circuit nullable fallback + inline single-branch fragments
4. `9eea849` Strip epsilon markers + cache grammar refs
5. `ddfe4b6` Return fragment results as children arrays
6. `77756bf` End-of-input sentinel token (drop range checks)
7. `daec1cb` Embed branch sequences in selector
8. `2609898` Compare selectStatement by rule id
9. `c4e7f8f` Add bench-parser-split.php tooling
10. `cd0b609` Deduplicate selector entries (memory: 40 MB → 10 MB)
11. `e8b006f` Cache branch sequences in deduped selector
12. `b5959e8` Grammar compilation experiment + bench tools (kept as documentation, not used at runtime)
13. `bf1b1fe` Pack switch case labels (compiled output cleanup)
14. `fa2fd65` Single-candidate rule fast path
15. `1e7e3cf` Direct-return refactor for single-candidate path
16. `9fcfb27` Mark WP_Parser_Node as final
17. `b9b64a6` bench-final.php helper

Tests: 667/667 phpunit tests pass with the same ~1.43M assertions as trunk.
Grammar memory: ~10 MB (from ~1.5 MB on trunk; cost of the new per-token branch selector + FIRST/NULLABLE precomputation).
PHPCS: clean.

## Test plan
- [x] `composer run test` (mysql-on-sqlite) - green locally
- [x] `composer run check-cs` - green
- [x] CI matrix on PHP 7.2 → 8.5 - all green, ~1.85× faster
- [ ] Confirm WordPress PHPUnit / E2E Tests pass

## Notes

This PR also includes a grammar→PHP compiler experiment under `tests/tools/compile-grammar.php`. It was kept as documentation: it's faster than the interpreter without JIT (+14-19%) but slightly slower under tracing JIT, so the interpreter remains the production path.

Draft because the goal here is to compare CI runtimes against recent merges before deciding what to ship.

### DIFF
--- a/packages/mysql-on-sqlite/src/mysql/class-wp-mysql-parser.php
+++ b/packages/mysql-on-sqlite/src/mysql/class-wp-mysql-parser.php
@@ -29,7 +29,7 @@ class WP_MySQL_Parser extends WP_Parser {
 	 * @return bool Whether a query was successfully parsed.
 	 */
 	public function next_query(): bool {
-		if ( $this->position >= count( $this->tokens ) ) {
+		if ( $this->position >= $this->token_count ) {
 			return false;
 		}
 		$this->current_ast = $this->parse();
--- a/packages/mysql-on-sqlite/src/parser/class-wp-parser-grammar.php
+++ b/packages/mysql-on-sqlite/src/parser/class-wp-parser-grammar.php
@@ -111,7 +111,37 @@ class WP_Parser_Grammar {
 		}
 
 		$this->inline_single_branch_fragments();
+		$this->strip_epsilon_markers();
 		$this->build_branch_selectors();
+	}
+
+	/**
+	 * Remove explicit `EMPTY_RULE_ID` markers from branches.
+	 *
+	 * The epsilon marker is a zero-width, always-matching symbol used in the
+	 * grammar to express optional productions. At parse time it would still
+	 * be walked and "continued" over for no effect, so stripping it ahead of
+	 * time removes a per-symbol branch in the hot loop.
+	 *
+	 * A pure-epsilon branch (`[EMPTY_RULE_ID]`) becomes an empty branch (`[]`)
+	 * which the parser already handles: the inner symbol loop does nothing and
+	 * the rule returns a successful empty match.
+	 */
+	private function strip_epsilon_markers() {
+		foreach ( $this->rules as $rule_id => $branches ) {
+			foreach ( $branches as $i => $branch ) {
+				if ( in_array( self::EMPTY_RULE_ID, $branch, true ) ) {
+					$this->rules[ $rule_id ][ $i ] = array_values(
+						array_filter(
+							$branch,
+							static function ( $s ) {
+								return self::EMPTY_RULE_ID !== $s;
+							}
+						)
+					);
+				}
+			}
+		}
 	}
 
 	/**
--- a/packages/mysql-on-sqlite/src/parser/class-wp-parser-grammar.php
+++ b/packages/mysql-on-sqlite/src/parser/class-wp-parser-grammar.php
@@ -337,15 +337,20 @@ class WP_Parser_Grammar {
 				$this->nullable_branches[ $rule_id ] = true;
 			}
 			if ( $selector ) {
-				// Store the candidate branch sequences directly so the parser
-				// can foreach over them without an extra $branches[$idx]
-				// indirection on every branch attempt.
+				// Many tokens in the same rule end up mapping to the same
+				// branch-id list (often because they all belong to a single
+				// branch's FIRST set). Deduplicate by signature so tokens
+				// share a single inner array via copy-on-write, turning the
+				// nested selector table from ~40 MB into ~1-2 MB without
+				// changing runtime behavior.
+				$by_signature = array();
 				foreach ( $selector as $tid => $idx_list ) {
-					$seqs = array();
-					foreach ( $idx_list as $idx ) {
-						$seqs[] = $branches[ $idx ];
+					$sig = implode( ',', $idx_list );
+					if ( isset( $by_signature[ $sig ] ) ) {
+						$selector[ $tid ] = $by_signature[ $sig ];
+					} else {
+						$by_signature[ $sig ] = $idx_list;
 					}
-					$selector[ $tid ] = $seqs;
 				}
 				$this->branches_for_token[ $rule_id ] = $selector;
 			}
--- a/packages/mysql-on-sqlite/src/parser/class-wp-parser-grammar.php
+++ b/packages/mysql-on-sqlite/src/parser/class-wp-parser-grammar.php
@@ -188,8 +188,8 @@ class WP_Parser_Grammar {
 		}
 
 		// Depth-first expansion memoized per rule, with cycle detection.
-		$expanded = array();
-		$visiting = array();
+		$expanded      = array();
+		$visiting      = array();
 		$expand_branch = function ( array $branch ) use ( &$expand_branch, &$expanded, &$visiting, $rules, $low_nt, $inlinable ) {
 			$out = array();
 			foreach ( $branch as $sym ) {
@@ -207,8 +207,8 @@ class WP_Parser_Grammar {
 					continue;
 				}
 				if ( ! isset( $expanded[ $sym ] ) ) {
-					$visiting[ $sym ]    = true;
-					$expanded[ $sym ]    = $expand_branch( $rules[ $sym ][0] );
+					$visiting[ $sym ] = true;
+					$expanded[ $sym ] = $expand_branch( $rules[ $sym ][0] );
 					unset( $visiting[ $sym ] );
 				}
 				foreach ( $expanded[ $sym ] as $s ) {
@@ -239,12 +239,12 @@ class WP_Parser_Grammar {
 	 * of the branch attempts that the parser used to try and fail.
 	 */
 	private function build_branch_selectors() {
-		$rules        = $this->rules;
-		$low_nt       = $this->lowest_non_terminal_id;
-		$empty_rule   = self::EMPTY_RULE_ID;
-		$rule_ids     = array_keys( $rules );
-		$nullable     = array();
-		$first_sets   = array();
+		$rules      = $this->rules;
+		$low_nt     = $this->lowest_non_terminal_id;
+		$empty_rule = self::EMPTY_RULE_ID;
+		$rule_ids   = array_keys( $rules );
+		$nullable   = array();
+		$first_sets = array();
 
 		foreach ( $rule_ids as $rule_id ) {
 			$nullable[ $rule_id ]   = false;
--- a/packages/mysql-on-sqlite/src/parser/class-wp-parser-grammar.php
+++ b/packages/mysql-on-sqlite/src/parser/class-wp-parser-grammar.php
@@ -29,7 +29,32 @@ class WP_Parser_Grammar {
 	public $rules;
 	public $rule_names;
 	public $fragment_ids;
-	public $lookahead_is_match_possible = array();
+
+	/**
+	 * Per-rule branch selector keyed by the next token id.
+	 *
+	 * When set, `$branches_for_token[$rule_id][$token_id]` is the ordered list
+	 * of branch indexes in `$rules[$rule_id]` that can possibly match when the
+	 * current token has the given id. Nullable branches appear in every entry.
+	 *
+	 * If an entry does not exist for the current token, `$nullable_branches`
+	 * is consulted. If both are empty, the rule cannot match and the parser
+	 * returns immediately.
+	 *
+	 * Rules whose FIRST set could not be computed do not appear in the map;
+	 * for those the parser falls back to trying every branch.
+	 *
+	 * @var array<int,array<int,int[]>>
+	 */
+	public $branches_for_token = array();
+
+	/**
+	 * Per-rule list of nullable branch indexes.
+	 *
+	 * @var array<int,int[]>
+	 */
+	public $nullable_branches = array();
+
 	public $lowest_non_terminal_id;
 	public $highest_terminal_id;
 
@@ -56,8 +81,8 @@ class WP_Parser_Grammar {
 		$this->highest_terminal_id    = $this->lowest_non_terminal_id - 1;
 
 		foreach ( $grammar['rules_names'] as $rule_index => $rule_name ) {
-			$this->rule_names[ $rule_index + $grammar['rules_offset'] ] = $rule_name;
-			$this->rules[ $rule_index + $grammar['rules_offset'] ]      = array();
+			$rule_id                      = $rule_index + $grammar['rules_offset'];
+			$this->rule_names[ $rule_id ] = $rule_name;
 
 			/**
 			 * Treat all intermediate rules as fragments to inline before returning
@@ -75,7 +100,7 @@ class WP_Parser_Grammar {
 			 * They are prefixed with a "%" to be distinguished from the original rules.
 			 */
 			if ( '%' === $rule_name[0] ) {
-				$this->fragment_ids[ $rule_index + $grammar['rules_offset'] ] = true;
+				$this->fragment_ids[ $rule_id ] = true;
 			}
 		}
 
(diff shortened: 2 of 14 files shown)


JanJakes:
## CI runtime comparison — full matrix

Compared this PR (`parser-performance`) against the average of three recent merged feature PRs (#365, #367, #369). All values are full-job duration (includes SQLite build + composer + tests).

| Workflow                        | Parser-perf | Trunk avg | Δ            |
|--------------------------------|-------------|-----------|--------------|
| PHP 7.2 / SQLite 3.27.0        | **132 s**   | 201 s     | **−34%**     |
| PHP 7.3 / SQLite 3.31.1        | **116 s**   | 172 s     | **−32%**     |
| PHP 7.4 / SQLite 3.34.1        | **94 s**    | 158 s     | **−40%**     |
| PHP 8.0 / SQLite 3.37.0        | **111 s**   | 169 s     | **−34%**     |
| PHP 8.1 / SQLite 3.40.1        | **116 s**   | 183 s     | **−36%**     |
| PHP 8.2 / SQLite 3.45.1        | **115 s**   | 176 s     | **−35%**     |
| PHP 8.3 / SQLite 3.46.1        | **114 s**   | 175 s     | **−35%**     |
| PHP 8.4 / SQLite 3.51.2        | **103 s**   | 175 s     | **−41%**     |
| PHP 8.5 / SQLite latest        | **119 s**   | 183 s     | **−35%**     |
| **PHPUnit Tests (sum)**        | **1020 s**  | 1592 s    | **−36%, ~9.5 min saved** |
| WordPress PHPUnit Tests        | **587 s**   | 713 s     | **−18%**     |
| MySQL Proxy Tests              | 18 s        | 20 s      | -10%         |
| End-to-end Tests               | 340 s       | 315 s     | +8% (noise)  |
| WordPress End-to-end Tests     | 441 s       | 360 s     | +22% (noise) |

The PHPUnit numbers above include ~30-40 s of SQLite build and ~10 s of composer install which are unchanged. Looking at just the `Run PHPUnit tests` step:

| PHP   | Trunk avg | This PR  | Speedup |
|-------|-----------|----------|---------|
| 7.2   | 165 s     | **97 s**  | 1.70×   |
| 7.3   | 126 s     | **73 s**  | 1.73×   |
| 7.4   | 120 s     | **57 s**  | 2.11×   |
| 8.0   | 132 s     | **68 s**  | 1.94×   |
| 8.1   | 141 s     | **75 s**  | 1.88×   |
| 8.2   | 128 s     | **70 s**  | 1.83×   |
| 8.3   | 134 s     | **74 s**  | 1.81×   |
| 8.4   | 138 s     | **72 s**  | 1.92×   |
| 8.5   | 141 s     | **79 s**  | 1.78×   |

**~1.85× faster phpunit across all PHP versions on real CI**, matching the local end-to-end benchmark.

E2E variance is browser/wp-env startup dominated and within normal run-to-run noise; not parser-related.

All tests pass. Total saved per CI run on PHPUnit + WordPress PHPUnit alone: **~11 minutes**.

JanJakes:
## Bonus experiment: compile the grammar to a single PCRE2 regex

Tried the `( COND1 (*THEN) FOO | COND2 (*THEN) BAR | ... )` pattern from PCRE2's "Backtracking control verbs" section. Findings:

### Approach

- Each token id encoded as a Unicode codepoint (`token_id + 0x4000`), so the lex output becomes a UTF-8 string the regex matches against.
- Each grammar rule emitted as `(?<rN>...)` named subroutine.
- Each branch's first symbol followed by `(*THEN)` so PCRE commits to the alternative once that token matches.
- Aggressive transitive inlining of single-use, non-recursive rules to fit under PCRE2's compiled-pattern size limit (LINK_SIZE=2 default).
- `(*THEN)` only emitted on rules whose sibling-branch FIRST sets are disjoint (the same `single_candidate_rules` set the interpreter uses), so committing is correctness-safe.
- Mirrors the parser's `SELECT ... INTO ...` negative lookahead via `(?!INTO_TOKEN)` after `selectStatement`.

### Results on the 69,576-query MySQL test corpus

- Pattern: 76 KB source / 1,127 named subroutines (789 of the original 1,916 rules inlined transitively).
- PCRE2 compile time: ~22 ms.
- **Match throughput with PCRE2 JIT: ~101,400 QPS**
- vs the optimised interpreter on this branch: ~62,500 QPS (best with tracing JIT)
- vs the original trunk parser: ~23,700 QPS
- **~62% faster than the optimised interpreter, ~4.3× the trunk parser**

### Caveats

1. **Match-only**: the regex returns true/false for "is this a valid query?". It does not produce an AST, so it's *not* a drop-in replacement for the parser — the SQLite driver consumes the AST.
2. **~0.15% spurious failures** (102 out of 69,576): the grammar has ambiguities the regex doesn't model perfectly, mostly around `SELECT ... INTO ... FOR UPDATE`, weird identifiers, and nested versioned comments.
3. **Without `(*THEN)` it is catastrophically slow** — a few CREATE TRIGGER queries with nested `BEGIN ... IF ... END` blocks individually take 3–4 seconds before that flag was added. With it, the worst-case queries drop to ~8 ms.
4. **PCRE2's bytecode size limit** is the binding constraint. The full grammar without inlining is ~1.2 MB of regex source and PCRE2 refuses to compile it. Aggressive single-use rule inlining was needed to bring it down to a compilable form.

### Conclusion

PCRE2 + `(*THEN)` *can* recognise this MySQL grammar at ~100 K QPS, ~60% faster than our pure-PHP interpreter. But it's a recogniser, not a parser — useful as a fast "does this parse?" gate, not as a replacement for the AST-producing parser the driver needs.

Committed in `tests/tools/exp-regex-v3.php` for documentation.

JanJakes:
## Follow-up: can the regex return / reconstruct an AST?

**Short answer: no, and even using it as a pre-validator costs more than it saves.**

### What PCRE2 in PHP exposes

- `preg_match` returns binary match / no-match, plus offsets and the LAST match per named capture group. A recursive `(?<name>...)` group, like our rules, only retains the outermost or final instance — useless for an AST.
- `(?J)` (duplicate names) doesn't change this in PHP's preg API.
- `(?C)` callouts compile cleanly but **PHP has no `pcre_callout` callback exposed**, so we can't intercept matching to record structural events.
- `(*MARK:NAME)` only retains the last MARK along the successful path.

So we can't structurally extract a parse tree from a single PCRE2 match in PHP without writing an extension or using FFI.

### Hybrid: regex pre-validate, then parser builds AST

Tested on the full 69,576-query MySQL test corpus, tracing JIT enabled:

| Approach              | Time     | QPS      | Notes                              |
|-----------------------|----------|----------|-------------------------------------|
| regex only (no AST)   | 0.752 s  | 92,519   | 102 spurious failures              |
| parser only (AST)     | 1.136 s  | 61,240   | 9 baseline failures                |
| regex + parser        | 1.480 s  | 47,008   | regex pre-validates; parser builds AST on success |

The hybrid is **slower than the parser alone** — the regex pre-check is pure overhead because the parser still has to run on every valid query (99.99% of the corpus), and valid is the common case. Hybrid only pays off when many inputs are **invalid**, where the regex can reject them faster than the parser; that's not our workload.

### What it's actually good for

- **A fast "does this query parse?" gate**: ~92K QPS with no AST, useful in defensive contexts where you want to reject malformed input cheaply before doing serious work.
- **An upper bound on PHP-side parsing speed for this grammar** (~100K QPS) when AST shape is not required.
- **As-is, not a replacement for the production parser**, because the SQLite driver consumes the AST.

### Bottom line for the parser-performance PR

The optimised interpreter on this branch (~62K QPS with JIT, ~24K without) is the production path because it's the only approach that returns an AST. The regex experiment is committed in `tests/tools/exp-regex-v3.php` and the hybrid in `tests/tools/exp-regex-hybrid.php` purely as documentation of where PCRE2 can and can't help.

JanJakes:
## Deep dive: PCRE2 / PHP options for AST extraction

Surveyed every PCRE2 feature exposed in PHP and a few that aren't:

### What PHP's preg_* exposes

| Feature | Behavior with recursive named groups | Useful? |
|---|---|---|
| `PREG_OFFSET_CAPTURE` | One offset pair per numbered group, last-match-wins | partial |
| `PREG_UNMATCHED_AS_NULL` | NULL vs empty string for unmatched | partial |
| `(?<name>...)` | Last match per name | no |
| `(?J)` duplicate named groups | `[name]` returns first non-empty; numbered slots [N] each store last match of *that occurrence* in the static pattern, **not** all match instances during execution | no |
| `(*MARK:NAME)` | Only the last MARK along the successful path is in `$matches['MARK']` | no |
| `\K` | Resets match start; cannot record multiple positions | no |
| `(?C)` callouts | **Compile cleanly but PHP exposes no callback registration** — they are silently no-ops | blocked |
| `preg_match_all` | Multiple top-level matches; for a recursive whole-input pattern, only one outer match | no |
| `preg_replace_callback` | Callback fires per outer match, not on sub-pattern entries | no |

Verified all of the above with concrete tests; results pasted in exp-regex-* tools.

### What's NOT exposed

PCRE2's C API has **callouts** (`pcre2_set_callout`) where a callback fires at every `(?C)` site with full `pcre2_callout_block` info: `current_position`, `offset_vector`, `capture_top`, `capture_last`, `mark`, `callout_flags` (incl. `PCRE2_CALLOUT_BACKTRACK`). That's exactly what's needed to record entry/exit events for every rule and reconstruct an AST.

PHP's `php_pcre.c` does not expose this API. There is no `pcre_callout()` user function, no class, no INI-driven hook, and no other documented mechanism in PHP 8.5.

### Tried FFI as an escape hatch

PHP 7.4+ has FFI. In principle we can call `pcre2_compile_8`, `pcre2_match_8`, `pcre2_set_callout_8` directly. **But** PHP FFI deliberately does **not** support binding PHP closures to C function pointers — libffi closure support is not enabled. So we cannot supply a callout function from PHP. Source pasted in `tests/tools/exp-pcre-ffi.php`.

That leaves a custom PHP extension wrapping the callout API as the only path to PCRE2-driven AST extraction. Out of scope here.

### Multi-pass extraction is just a parser

The remaining alternative — match the full pattern to validate, then `preg_match_all` repeatedly with simpler flat patterns to extract structure — is structurally equivalent to writing a parser in PHP that uses preg_match per rule. Per-call overhead (~55 ns for cached patterns) is fine, but the algorithmic structure is the same as the existing recursive-descent interpreter, with no speedup expected.

### Summary

In stock PHP the compiled regex is **a fast yes/no validator only**:

- ~92K QPS on the full corpus, no AST.
- ~100K QPS upper bound on parse-recognition speed for this grammar.
- Regex + parser hybrid is slower than the parser alone (~47K QPS vs ~62K), because the regex is pure overhead on valid input and ~99.99% of input is valid.

The existing recursive-descent interpreter (~62K QPS with tracing JIT, ~34K without; both delivering an AST) remains the production path. Three exp- scripts in `tests/tools/` document the regex experiments for future reference:

- `exp-regex-v3.php` — match-only, the fast version
- `exp-regex-hybrid.php` — regex pre-validate + parser
- `exp-pcre-ffi.php` — explains why FFI can't supply callouts

JanJakes:
## Re-checked the SO technique — partial correction

The SO answer at https://stackoverflow.com/a/24279157 uses **`preg_match_all` with a flat pattern + a lookbehind anchor** to get every instance of a repeated sub-pattern. That technique does work and I should have tried it more thoroughly before declaring "no AST".

Here's what I verified:

### What `preg_match_all` actually does for our case

| Pattern | Behavior |
|---|---|
| `(?<=\d),\d+` (flat + lookbehind) | Returns each match as a separate element. Works fine for flat repeating data. |
| `(?P<paren>\((?:[^()]+|(?P>paren))*\))` (recursive) | **Returns each TOP-LEVEL non-overlapping match.** Inner nested instances live inside an outer match — to get them, recurse with another `preg_match_all` on the outer's content. |

So the recursive named-group + `preg_match_all` combination *can* walk a tree, level by level. Each call returns the outermost matches at the current scope; you recurse into each match's body to find inner ones.

Speed for the tree-walking approach (10k iterations on `SELECT a, (SELECT b, (SELECT c FROM t3) FROM t2), d FROM t1`):

- 1 outer `preg_match_all` (recursive paren pattern): ~12.5M QPS
- 1 anchored extraction (`/FROM\s+\K\w+/` for tables): bundled together ~3.8M QPS

That's two orders of magnitude faster than the parser per call. The economics make sense for **focused extraction** (e.g., "find all table refs", "find all subquery boundaries").

### Where this stops being a path to a full AST

The MySQL grammar has ~1,900 rules. To produce a full AST via this technique you'd need a regex per rule applied at every level of the tree. For our 17-token-avg corpus with depth ~10 levels:

- **Partial AST** (top-level statement structure + a handful of rules like tableRef / columnRef / whereClause): **a few preg_match_all calls per query** — easily faster than the parser, useful for queries that only need certain shapes.
- **Full AST** (every rule, every nesting): would require ~50+ regex calls per query for typical queries, ~3–5M total over the corpus. At ~55 ns/cached call that's ~200 ms minimum, but each call's actual regex work is the dominant cost — and many of those patterns must include the full grammar definitions to disambiguate context. Estimated worse than the existing parser.

### Why the answer to "AST from a single match?" is still no

`preg_match_all` works because it **runs the pattern multiple times** under the hood, advancing the cursor between matches. Each match is a *fresh* pattern execution. So you do not get "all captures from one match" — you get "many top-level matches, each with last-match-wins captures inside".

For the recursive grammar where the same rule is matched in nested positions during a single execution, those nested instances are *not* exposed because PCRE2 only stores last-match-wins per group at the C level too. PHP would need PCRE2 callouts — and as documented in `exp-pcre-ffi.php`, PHP doesn't expose them and FFI can't bridge the gap because PHP closures aren't bindable to C function pointers.

### Practical takeaway

The SO technique gives a real win for **partial / flat / repeated** extraction:

```php
// Anchor + capture: every table that follows FROM
preg_match_all('/FROM\s+\K\w+/', $sql, $m);

// Recursive: every top-level subquery span (non-overlapping)
preg_match_all('/(?P<paren>\((?:[^()]+|(?P>paren))*\))/', $sql, $m, PREG_OFFSET_CAPTURE);
```

Both run at multi-million QPS. Useful as a query-shape introspection layer alongside the parser — e.g. for fast "list referenced tables" or "is there a subquery here" without touching the AST.

For producing the AST the SQLite driver actually consumes, the recursive-descent interpreter on this branch (~62K QPS with JIT) remains the right tool.

Updated mental model: PCRE in PHP gives you **(1) a fast yes/no recogniser, (2) cheap `preg_match_all`-driven flat/partial extraction, but not (3) a one-shot full nested parse tree**.

JanJakes:
Closing in favor of https://github.com/WordPress/sqlite-database-integration/pull/378.

PR #374, opened 2026-04-27
Lexer: Fix possible OOB read in quoted strings

sirreal:
## Summary

Fixes an out-of-bounds string access in `WP_MySQL_Lexer::read_quoted_text()` that produces PHP warnings when lexing unclosed quoted strings with trailing backslashes:

```
Uninitialized string offset N in class-wp-mysql-lexer.php on line 2855
```

This appears to happen in streaming SQL processing (e.g., WordPress Playground's `runSql` blueprint step) when a buffer boundary splits a quoted string literal and a backslash falls at the end of the buffer. Standard MySQL dumps with escaped string literals contain thousands of backslashes, making this likely to hit in practice.

## The bug

The backslash-counting loop in `read_quoted_text()` ran **before** the EOF check. When `strcspn()` reached the end of the string without finding a closing quote:

1. `$at` pointed to `strlen($sql)` (one past the last byte).
2. The backslash loop accessed `$this->sql[$at - 1]` — valid, but if it was `\`:
3. The loop treated the absent quote as escaped and did `$at += 1` (now past end).
4. Next iteration: `strcspn` returned 0, `$at` stayed past end.
5. The backslash loop accessed `$this->sql[strlen($sql)]` — **out of bounds**.

## Fix

Two changes to the `while (true)` loop body:

1. **Move the EOF check before the backslash-counting loop.** When `strcspn` reaches end-of-string without finding the quote, `$this->sql[$at] ?? null` won't match `$quote`, so we return `null` immediately — the backslash loop is never reached.

2. **Add a lower-bound guard to the backslash loop.** The `for` condition now includes `($at - $i - 1) >= 0` to prevent underflow when a quote appears near the start of the string. Belt-and-suspenders — the EOF reorder already prevents the primary bug.

## Tests

- Unclosed strings with odd/even trailing backslashes (single and double quotes).
- Regression tests for valid escaped strings, doubled quotes, and backtick identifiers.
- Chunk boundary test simulating a streaming SQL processor splitting input at a backslash inside a quoted string.

## Use of AI

This was diagnosed and implemented with the help of Claude Code.

### DIFF
--- a/packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
+++ b/packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
@@ -2842,6 +2842,11 @@ class WP_MySQL_Lexer {
 		while ( true ) {
 			$at += strcspn( $this->sql, $quote, $at );
 
+			// Unclosed string - unexpected EOF.
+			if ( ( $this->sql[ $at ] ?? null ) !== $quote ) {
+				return null; // Invalid input.
+			}
+
 			/*
 			 * By default, quotes can be escaped with a "\".
 			 * When NO_BACKSLASH_ESCAPES SQL mode is active, the "\" treated as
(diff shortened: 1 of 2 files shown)


JanJakes (approved):
@sirreal Thanks for the fix! LGTM and works great!

PR #376, opened 2026-04-27
[codex] Split parser single-branch path

adamziel:
## What changed

This stacked draft PR builds on #375 and gives the parser a dedicated fast path for the common case where FIRST-set dispatch resolves a rule to exactly one candidate branch.

- Splits single-branch parsing from the multi-branch fallback in `WP_Parser::parse_recursive()`.
- Keeps the existing multi-branch backtracking path for ambiguous dispatch.
- Changes `WP_Parser_Node::has_child()` from `count() > 0` to `! empty()` for the hot AST completion check.

## Why

After #375, branch dispatch commonly resolves directly to a single branch. The parser still paid loop/index bookkeeping designed for multiple candidates. This change duplicates a small amount of branch parsing code so the dominant single-candidate path can avoid that work while keeping the generic parser architecture intact.

This is deliberately still compact: it does not generate a large parser or expand the grammar on disk.

## Performance

Benchmarks are noisy on this machine, so I compared this branch against the stacked base branch immediately before opening this PR.

- Base #375 branch: 69,577 queries in `11.55021s` @ `6,023 QPS`
- This branch: 69,577 queries in `10.90099s` @ `6,382 QPS`

That is roughly a `6%` incremental parser improvement on top of #375 in this run.

## Parser size constraint

`src/parser/*.php` plus `src/mysql/mysql-grammar.php` remains under the requested 200 KB on-disk cap:

- `95,298` bytes total

## Validation

- `git diff --check`
- `php -l` on modified parser files
- `composer run test -- --filter 'WP_MySQL_(Lexer|Server_Suite_(Lexer|Parser))|WP_Parser_Node'`
  - `143` tests
  - `1,421,037` assertions
- `composer run test`
  - `667` tests
  - `1,427,673` assertions
  - `2` skipped, `2` incomplete
- `php packages/mysql-on-sqlite/tests/tools/run-parser-benchmark.php`

## Notes

I also tested broader parser reshapes before settling on this smaller fast path: inline terminal matching, in-memory single-fragment expansion, lightweight fragment arrays, direct branch-array dispatch, and by-reference parser cursor state. Those were slower in this codebase, so they are not included here.

### DIFF
--- a/packages/mysql-on-sqlite/src/parser/class-wp-parser-node.php
+++ b/packages/mysql-on-sqlite/src/parser/class-wp-parser-node.php
@@ -108,7 +108,7 @@ class WP_Parser_Node {
 	 * @return bool True if this node has any child nodes or tokens, false otherwise.
 	 */
 	public function has_child(): bool {
-		return count( $this->children ) > 0;
+		return ! empty( $this->children );
 	}
 
 	/**
--- a/packages/mysql-on-sqlite/src/parser/class-wp-parser.php
+++ b/packages/mysql-on-sqlite/src/parser/class-wp-parser.php
@@ -84,12 +84,8 @@ class WP_Parser {
 
 		$branch_matches     = false;
 		$node               = null;
-		$branch_index_count = null === $single_branch_index ? count( $branch_indexes ) : 1;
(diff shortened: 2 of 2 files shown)


adamziel:
Closing per request; scrapping this parser rearchitecture experiment.

PR #378, opened 2026-04-28
Parser + lexer performance: consolidated 2–3× end-to-end speedup

JanJakes:
## Summary

Consolidates the parser optimisations from #373, the lexer + token-construction wins from #375, and the `has_child()` micro-opt from #376 into a clean, linear history (one shippable change per commit). On top of those it adds a grammar-construction speedup and reworks the native (Rust) parser to materialise its AST eagerly so the parse-node class can be `final`.

End-to-end (lex+parse) on the 69,577-query MySQL server corpus, pure PHP, best across 3 ABAB-alternated rounds × 5 timed iterations (2 warmup iters per round to heat the tracing JIT):

| Config | Trunk | This branch | **Speedup** |
|---|---:|---:|---:|
| Lex-only, JIT | 188,323 QPS | 369,687 QPS | **1.96×** |
| Parse-only, JIT | 27,091 QPS | 56,691 QPS | **2.09×** |
| End-to-end, JIT | 23,855 QPS | 50,061 QPS | **2.10×** |
| End-to-end, no opcache | 10,722 QPS | 28,362 QPS | **2.65×** |

PHP 8.5; PHP 8.1 verified within ~5%. Tracing JIT is per-worker and shared across requests, so steady-state traffic hits warm JIT (the 2 warmup iters model that); cold first-request numbers are roughly 1.2× for the JIT configs.

## Per-process startup cost and memory

The speedup comes from precomputing per-token FIRST sets and branch selectors. The grammar is built once per worker process (PHP's shared-nothing model resets statics between requests) and cached in `self::$mysql_grammar`, so the cost is amortised across every query in a request — but short requests pay it up front:

| | Trunk | This branch | Δ |
|---|---:|---:|---:|
| Grammar construction | ~0.8 ms | ~3.7 ms | **+~2.9 ms** |
| Grammar memory, resident at startup | ~1.4 MiB | ~9.0 MiB | **+~7.6 MiB** |
| Peak after parsing the full corpus | ~47 MiB | ~66 MiB | +~19 MiB |

A naive build of these structures cost ~40 ms. A worklist fixpoint for FIRST/NULLABLE (recompute a rule only when a referenced rule grows) plus lazy per-rule selector denormalisation (a typical request touches ~7% of rules; the rest is deferred and amortised into parsing) cut construction to the ~3.7 ms above. Numbers are plain CLI — opcache trims construction further. The extra resident memory is a real tradeoff on memory-constrained shared hosts (a supported no-opcache target).

## Native parser: eager AST materialisation

Marking `WP_Parser_Node` `final` (a measured +7% win) is incompatible with `WP_MySQL_Native_Parser_Node` subclassing it. Rather than drop `final`, the native parser now materialises its arena AST into plain `WP_Parser_Node` instances at parse time. This removes the entire lazy layer — the per-AST identity cache, the wrapper registry, ~18 `wp_sqlite_mysql_native_ast_*` bridge functions, and the `WP_MySQL_Native_Parser_Node` class (~600 LOC Rust + 179 LOC PHP), plus the two wrapper-lifetime test files they backed.

Because the translator walks essentially the whole AST for every query, eager materialisation removes per-node FFI round trips and is **neutral-to-faster while far simpler**: building the full tree, the native parser runs at **~60,100 QPS vs ~28,600 QPS pure PHP (2.1×)** on the corpus. It regresses only a hypothetical consumer that parses but inspects a tiny fraction of a large tree.

The grammar is exchanged with the extension as a runtime ABI, so `load.php` now pins the supported extension minor line and falls back cleanly to pure PHP on a mismatch (e.g. a plugin update outpacing the installed binary) instead of failing at parse time.

## What was kept from #373 / #375 / #376

- **#373 (parser):** all 13 parser optimisations preserved — they carry ~95% of the wall-clock win.
- **#375 (lexer):** only the lexer-side wins (cached `$sql_length`, byte comparisons replacing `strspn`/`strcspn` mask checks, `strpos`-based comment-end and quote scans, inlined `remaining_tokens`) plus the `WP_MySQL_Token` parent-constructor bypass. The parser overlap was deliberately *not* applied — #373's equivalents are stronger and mixing the two measured no incremental win.
- **#376:** only the `! empty( $children )` micro-opt; the `parse_recursive` split didn't beat the existing single-candidate fast path.

## Cost vs benefit (src LOC, end-to-end JIT)

- **Eight parser optimisations carry ~95% of the ~2× win for ~278 LOC** — biggest: per-token FIRST sets to skip unreachable branches (+40%), inline terminal matching + deferred node allocation (+30%), inline single-branch fragments + short-circuit nullable fallback (+16%), `final` parse-node class (+7%), parent-constructor bypass (+5%).
- **Three lexer dispatch fast paths: +13% end-to-end / +55% lex-only for +42 LOC** (inline whitespace skip, identifier/keyword arm at the top of the chain, single-byte operator dispatch table). Lex is only ~13% of parse time under JIT, so end-to-end dilutes the lex gain.
- **Four small opts (~127 LOC, ~+6% combined):** end-of-input sentinel token, embedded branch-symbol sequences, rule-id (vs name) comparison, single-candidate direct-return.
- **Zero under JIT but kept:** lexer byte-comparison primitives (free for non-JIT shared hosts; the dispatch fast paths build on them), the `has_child()` one-liner, CS whitespace re-alignment.

## Test plan

- [x] `composer run test` (mysql-on-sqlite) — **721/721, 1,533,741 assertions** (2 skipped, 2 incomplete), incl. `WP_Parser_Grammar_Tests` for the build-time transforms (epsilon stripping, fragment inlining + cycle termination, FIRST/NULLABLE selectors, single-candidate classification, `merge_sorted`)
- [x] `composer run check-cs` — clean
- [x] CI matrix PHP 7.2–8.5 (incl. `PHP 7.2 / SQLite 3.27.0`), the native-extension jobs, WordPress PHPUnit/E2E, WASM, and MySQL Proxy — all green
- [x] PHP 8.5 + PHP 8.1 verified within ~5%
- [x] Native-extension benchmark confirms the eager-materialisation change (~2.1× pure PHP)


packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
```diff
@@ -2284,10 +2295,49 @@ public function get_token(): ?WP_MySQL_Token {
...
-		while ( true === $this->next_token() ) {
-			$token    = $this->get_token();
-			$tokens[] = $token;
+		$tokens                            = array();
+		$no_backslash_escapes_sql_mode_set = $this->is_sql_mode_active(
+			self::SQL_MODE_NO_BACKSLASH_ESCAPES
+		);
+
+		while ( true ) {
+			if (
```
adamziel, line 2311:
```suggestion
			// Break on file end
			if (
```

packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
```diff
@@ -2284,10 +2295,49 @@ public function get_token(): ?WP_MySQL_Token {
...
+				self::EOF === $this->token_type
+				|| ( null === $this->token_type && $this->bytes_already_read > 0 )
```
adamziel, line 2313: Shouldn't EOF cover that?

packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
```diff
@@ -2354,12 +2404,52 @@ private function read_next_token(): ?int {
 		$byte      = $this->sql[ $this->bytes_already_read ] ?? null;
 		$next_byte = $this->sql[ $this->bytes_already_read + 1 ] ?? null;
 
-		if ( "'" === $byte || '"' === $byte || '`' === $byte ) {
+		// A map for a single-byte symbol fast path.
```
adamziel, line 2414: nice

packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
```diff
@@ -2354,12 +2404,52 @@ private function read_next_token(): ?int {
...
+			'=' => self::EQUAL_OPERATOR,
+		);
+
+		// Fast path for keywords and identifiers.
+		// These are the most common token types in MySQL payloads.
+		if (
+			(
+				( $byte >= 'a' && $byte <= 'z' )
+				|| ( $byte >= 'A' && $byte <= 'Z' )
+				|| $byte > "\x7F"
```
adamziel, line 2440: I'd leave a comment on why `\x7F` is special here

packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
```diff
@@ -2354,12 +2404,52 @@ private function read_next_token(): ?int {
...
+			)
+			&& "'" !== $next_byte
```
adamziel, line 2442: Why just `'` and not `"`? Would any quotes-related sql mode/session options have impact here?

packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
```diff
@@ -2547,7 +2637,13 @@ private function read_next_token(): ?int {
 			}
 		} elseif ( '#' === $byte ) {
 			$type = $this->read_line_comment();
-		} elseif ( null !== $byte && strspn( $byte, self::WHITESPACE_MASK ) > 0 ) {
+		} elseif (
+			' ' === $byte
```
adamziel, line 2622: Would array + isset() be faster?

packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
```diff
@@ -2675,7 +2767,7 @@ private function read_number(): ?int {
 				'0' === $byte
 				&& 'x' === $next_byte
 				&& null !== $third_byte
-				&& strspn( $third_byte, self::HEX_DIGIT_MASK ) > 0
+				&& false !== strpos( self::HEX_DIGIT_MASK, $third_byte )
```
adamziel, line 2755: clever

packages/mysql-on-sqlite/src/parser/class-wp-parser-node.php
```diff
@@ -9,17 +9,18 @@
  * In this way, a parser node constitutes a recursive structure that represents
  * a parse (sub)tree at each level of the full grammar tree.
  */
-class WP_Parser_Node {
+final class WP_Parser_Node {
```
adamziel, line 12: does `final` make it faster somehow?

packages/mysql-on-sqlite/src/parser/class-wp-parser.php
```diff
@@ -11,85 +11,177 @@
...
+	private $single_candidate_rules;
+
 	public function __construct( WP_Parser_Grammar $grammar, array $tokens ) {
-		$this->grammar  = $grammar;
-		$this->tokens   = $tokens;
-		$this->position = 0;
+		$this->grammar     = $grammar;
+		$this->token_count = count( $tokens );
+		// Append an end-of-input sentinel token whose id is EMPTY_RULE_ID
+		// (0). The hot path can then read $tokens[$pos]->id unconditionally
```
adamziel: pretty cool

packages/mysql-on-sqlite/src/parser/class-wp-parser.php
```diff
@@ -11,85 +11,177 @@
...
+		$this->fragment_ids           = $grammar->fragment_ids;
+		$this->branches_for_token     = $grammar->branches_for_token;
+		$this->nullable_branches      = $grammar->nullable_branches;
+		$this->highest_terminal_id    = $grammar->highest_terminal_id;
+		$this->single_candidate_rules = $grammar->single_candidate_rules;
+
+		// The INTO negative-lookahead only fires for selectStatement. Cache
+		// the rule id so the per-call check is an int compare instead of a
+		// string compare.
+		$this->select_statement_rule_id = $grammar->get_or_cache_rule_id( 'selectStatement' );
```
adamziel, line 43: Any memory impact of caching all the rules?

adamziel:
I've left some notes. Haven't read deeply into the diff but the idea makes sense – inline some stuff, reorder, cache, add a trailing token. Nothing revolutionary, but it's still a pretty clever way to get more juice out of it.

packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
JanJakes: Addressed in f9172e1.

packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
JanJakes: Addressed in f9172e1. EOF and the second arm catch different cases: `self::EOF` is set when `read_next_token()` sees a null byte at the *start* of a token (clean end-of-input). The `null === $this->token_type && $this->bytes_already_read > 0` arm catches the case where `read_next_token()` returned null mid-stream because of an invalid byte. The `> 0` guard keeps the very first iteration alive — at that point `$this->token_type` is still null because nothing has been read yet, not because we've failed.

packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
JanJakes: Addressed in f9172e1.

packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
JanJakes: Addressed in f9172e1.

packages/mysql-on-sqlite/src/mysql/class-wp-mysql-lexer.php
JanJakes: Marginally faster, but this branch rarely fires. `next_token()` and `remaining_tokens()` inline-skip whitespace before calling `read_next_token()` (commit f5b8932), so this arm only handles whitespace that appears between comments. Keeping the `===` chain for consistency with the rest of the dispatch.

packages/mysql-on-sqlite/src/parser/class-wp-parser-node.php
JanJakes: Yes — `final` lets opcache/JIT skip the vtable check on method calls. Measured at +7% end-to-end, see commit daa4185 and the "Big, robust wins" table in the PR description.

packages/mysql-on-sqlite/src/parser/class-wp-parser.php
JanJakes: Negligible. Those are array assignments, not copies — PHP arrays are copy-on-write, so the parser instance just holds references to the grammar's arrays. No actual duplication unless something writes to them, which the parser doesn't.

adamziel:
> If the native parser is reworked to not extend WP_Parser_Node in the future, this can be restored.

Let's file an issue and explore that, 7% is huge

JanJakes:
Closing in favor of https://github.com/WordPress/sqlite-database-integration/pull/429.

PR #384, opened 2026-04-28
Add a hook for native MySQL lexer and parser implementations

adamziel:
## Summary

Add a small entry point in `load.php` so an optional native (e.g. Rust) MySQL lexer/parser extension can plug itself in as `WP_MySQL_Lexer` / `WP_MySQL_Parser`. The extension pre-declares `WP_MySQL_Native_Lexer` / `WP_MySQL_Native_Parser` before this file loads; otherwise the existing pure-PHP classes ship as the implementation, unchanged.

The existing PHP files keep their original names, locations, and class names — no rename, no polyfill scaffolding. Native shims live in a separate `mysql/native/` directory and are only required when the native classes are present.

## Testing

- `php -l` on changed files
- Smoke: `WP_MySQL_Lexer` and `WP_MySQL_Parser` instantiate and parse `SELECT 1` end-to-end with no native extension loaded
- `phpcs` clean

### DIFF
--- a/packages/mysql-on-sqlite/src/load.php
+++ b/packages/mysql-on-sqlite/src/load.php
@@ -12,8 +12,24 @@ require_once __DIR__ . '/parser/class-wp-parser.php';
 require_once __DIR__ . '/parser/class-wp-parser-node.php';
 require_once __DIR__ . '/parser/class-wp-parser-token.php';
 require_once __DIR__ . '/mysql/class-wp-mysql-token.php';
-require_once __DIR__ . '/mysql/class-wp-mysql-lexer.php';
-require_once __DIR__ . '/mysql/class-wp-mysql-parser.php';
+
+/*
+ * The MySQL lexer and parser have an optional native (e.g. Rust) implementation.
+ * When the native extension is loaded, it pre-declares WP_MySQL_Native_Lexer /
+ * WP_MySQL_Native_Parser; otherwise we fall back to the pure-PHP classes shipped
+ * here. WP_MySQL_Lexer / WP_MySQL_Parser is the public entrypoint either way.
+ */
+if ( class_exists( 'WP_MySQL_Native_Lexer', false ) ) {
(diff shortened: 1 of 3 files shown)


JanJakes (commented):
@adamziel The changes here are clear. I just have two API/naming observations:

1. The "polyfill" naming feels a bit misleading. In many use cases, and especially in the plugin context, the PHP classes are the source of truth and the only possible implementation.

2. Related to 1, is the `WP_MySQL_Parser extends WP_MySQL_Polyfill_Parser` part necessary? Would it be possible to just require the existing classes conditionally in https://github.com/WordPress/sqlite-database-integration/blob/trunk/packages/mysql-on-sqlite/src/load.php without even moving and renaming them:
```php
if ( class_exists( 'WP_MySQL_Native_Parser', false ) ) {
    class WP_MySQL_Parser extends WP_MySQL_Native_Parser {}
} else {
    require_once __DIR__ . '/mysql/class-wp-mysql-parser.php';
}
```

Or, if both have the same class name, just:
```php
if ( ! class_exists( 'WP_MySQL_Parser', false ) ) {
    require_once __DIR__ . '/mysql/class-wp-mysql-parser.php';
}
```

That way, the existing code would remain largely as-is, and we'd just add some native extension entrypoint.

Btw. If we need to stick with renames, I don't know why, but GitHub diff doesn't see them as renames. Perhaps that's fixable for better history and blames?

adamziel:
@JanJakes I like this option:

```php
if ( class_exists( 'WP_MySQL_Native_Parser', false ) ) {
    class WP_MySQL_Parser extends WP_MySQL_Native_Parser {}
} else {
    require_once __DIR__ . '/mysql/class-wp-mysql-parser.php';
}
```

It's implemented now :) I'll fix PHPCS and we may be good to go here

JanJakes (approved):
@adamziel Amazing, thanks!

PR #385, opened 2026-04-28
Add optional native parser routing

adamziel:
## Summary
- add fallback `WP_MySQL_Native_*` classes that extend the PHP polyfills
- route public MySQL lexer/parser classes through native classes when the extension preloads them
- add the minimal PHP bridge that exports grammar data for native parser setup

## Testing
- `php -l` on changed PHP files
- parser smoke: lex and parse `SELECT 1` through `WP_MySQL_Lexer` and `WP_MySQL_Parser`

This PR is intentionally stacked after the polyfill extraction PR so the native routing diff is small and reviewable.

### DIFF
--- a/packages/mysql-on-sqlite/src/load.php
+++ b/packages/mysql-on-sqlite/src/load.php
@@ -26,6 +26,7 @@ if ( class_exists( 'WP_MySQL_Native_Lexer', false ) ) {
 }
 
 if ( class_exists( 'WP_MySQL_Native_Parser', false ) ) {
+	require_once __DIR__ . '/mysql/native/mysql-rust-bridge.php';
 	require_once __DIR__ . '/mysql/native/class-wp-mysql-parser.php';
 } else {
 	require_once __DIR__ . '/mysql/class-wp-mysql-parser.php';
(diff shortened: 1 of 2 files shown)


packages/mysql-on-sqlite/src/load.php
```diff
@@ -26,6 +26,7 @@
 }
 
 if ( class_exists( 'WP_MySQL_Native_Parser', false ) ) {
+	require_once __DIR__ . '/mysql/mysql-rust-bridge.php';
```
JanJakes: @adamziel Should we move this file under the `native` directory?

adamziel: Good call — moved it to `mysql/native/mysql-rust-bridge.php`. Same logic applies to #386: `class-wp-mysql-native-parser-node.php` is only loaded when native is in use, so it now lives at `mysql/native/class-wp-mysql-native-parser-node.php` too.

adamziel: Done!

adamziel:
oops merged into the wrong branch, will merge to trunk

PR #386, opened 2026-04-28
Add lazy native parser node facade

adamziel:
## Summary

Adds `WP_MySQL_Native_Parser_Node`, a `WP_Parser_Node` subclass the native MySQL parser extension constructs when it produces an AST. Read methods delegate into the Rust-owned AST so children aren't copied into PHP unless something walks the tree. On the first `append_child()` / `merge_fragment()`, the node copies its native children into the inherited `$children` array and behaves like a plain `WP_Parser_Node` from then on — that's what `was_mutated()` tracks.

Mutation matters because `WP_PDO_MySQL_On_SQLite` rewrites parsed queries (e.g. synthetic `count(*)` expressions) by appending children to existing nodes.

## Other change

`WP_Parser_Node::$children` private → protected so the subclass can populate it during materialization.


JanJakes (approved):
Looks good! Thanks!

PR #391, opened 2026-04-30
Preserve child wrapper identity in the native AST facade

adamziel:
Stacked on #381.

#381's review surfaced a real semantics regression in the lazy native AST facade: every accessor on `WP_MySQL_Native_Parser_Node` calls into Rust and returns a freshly constructed PHP wrapper, so `get_first_child_node()` returns a different object every time. `WP_Parser_Node` has always given callers stable child identity — attach state to a child once, walk past it, walk back, the state is still there — and the lazy native facade is meant to keep that contract intact. This restores it.

A per-AST `WP_MySQL_Native_AST_Cache` is created lazily on the root and shared by reference with every wrapper that gets interned through it. Each accessor looks the returned wrapper's `native_node_index` up in the cache and either returns the canonical instance or registers the new one. `materialize_native_children()` pulls children through the same cache so any mutation a caller made through `get_first_child_node()` before the parent went through `append_child()` survives into `$this->children` — same instance, same mutations.

Tokens are unchanged. The public token API has no mutators and no callers in this repo rely on `WP_MySQL_Token` identity; if that becomes a need we can extend the cache.

## What's in here

- `class-wp-mysql-native-ast-cache.php` — small holder, one per AST.
- Native node accessors run results through `intern()` / `intern_all()` / `intern_nodes()`.
- `materialize_native_children()` reuses the interned wrappers so prior mutations don't get lost.
- Regression test covering same-instance reads, descendant/child identity sharing, and the mutate-then-materialize edge case. Skips when the native extension isn't loaded.
- `run-native-ast-walk-benchmark.php` parses the MySQL server suite, walks each AST, and reports `parsed`, `walked_nodes`, `duration`, `peak_mem`, and an identity-stability flag.
- `Native AST Walk Perf` workflow runs the benchmark on this PR and on the PR base (`codex/native-lazy-ast-facade`) on the same runner, so the identity-cache cost is measured apples-to-apples on every push.

## Performance

Benchmarked on CI against the no-cache baseline, same runner, same corpus (69,567 queries, 4.8M walked nodes). Full numbers in PR comments — links below.

| scenario | baseline (no cache) | this PR | delta |
|---|---|---|---|
| native parse only | 1.28s | 1.28s | 0% |
| native walk duration | 3.52s | 3.33s | **+5% (+0.19s)** |
| native walk qps | 19,766 | 20,884 | **−5%** |
| native walk peak memory | 50.0MB | 60.5MB | **+10.5MB (+21%)** |
| native walk identity stable | **FALSE** | true | regression fixed |

Hot-path optimization in `intern_all()` (cache reference hoisted out of the loop, per-item logic inlined) plus a typed `intern_nodes()` fast path for accessors that return only nodes brought the time penalty down from an initial +17% to +5%. Memory delta is structural — one retained PHP wrapper per visited node — and is the price of stable identity.

Native walk is still ~5× faster than the pure-PHP path while now preserving `WP_Parser_Node` semantics.

- First measurement: https://github.com/WordPress/sqlite-database-integration/pull/391#issuecomment-4354700002
- After hot-path optimization: https://github.com/WordPress/sqlite-database-integration/pull/391#issuecomment-4354952752

## Test plan

- [x] PHPUnit suite passes on the pure-PHP path.
- [x] PHPUnit suite passes with the native extension loaded; the new identity tests run instead of skipping.
- [x] `Native AST Walk Perf` workflow reports the cache cost vs. the PR base on every push.

### DIFF
--- a/.github/workflows/mysql-parser-extension-tests.yml
+++ b/.github/workflows/mysql-parser-extension-tests.yml
@@ -81,7 +81,10 @@ jobs:
               exit( 1 );
           }
           '
-          ./vendor/bin/phpunit -c ./phpunit.xml.dist tests/mysql/WP_MySQL_Lexer_Tests.php tests/parser/WP_Parser_Node_Tests.php
+        working-directory: packages/mysql-on-sqlite
+
+      - name: Run PHPUnit tests with parser extension
+        run: php -d extension="$GITHUB_WORKSPACE/packages/php-ext-wp-mysql-parser/target/debug/libwp_mysql_parser.so" ./vendor/bin/phpunit -c ./phpunit.xml.dist
         working-directory: packages/mysql-on-sqlite
 
   sqlite-driver-extension-tests:
@@ -149,3 +152,7 @@ jobs:
               exit( 1 );
           }
           '
+
+      - name: Run PHPUnit tests with SQLite driver using parser extension
+        run: php -d extension="$GITHUB_WORKSPACE/packages/php-ext-wp-mysql-parser/target/debug/libwp_mysql_parser.so" ./vendor/bin/phpunit -c ./phpunit.xml.dist
+        working-directory: packages/mysql-on-sqlite
--- a/.github/workflows/native-ast-perf.yml
+++ b/.github/workflows/native-ast-perf.yml
@@ -120,14 +120,14 @@ jobs:
       - name: Benchmark — baseline native path (walk, no identity cache)
         working-directory: ../baseline/packages/mysql-on-sqlite
         run: |
-          BASE_EXT="$(realpath ../packages/php-ext-wp-mysql-parser/target/release/libwp_mysql_parser.so)"
+          BASE_EXT="$(realpath ../../packages/php-ext-wp-mysql-parser/target/release/libwp_mysql_parser.so)"
           php -d extension="$BASE_EXT" tests/tools/run-native-ast-walk-benchmark.php \
             | tee "$GITHUB_WORKSPACE/packages/mysql-on-sqlite/baseline-native-walk.txt"
 
       - name: Benchmark — baseline native path (parse only)
         working-directory: ../baseline/packages/mysql-on-sqlite
         run: |
-          BASE_EXT="$(realpath ../packages/php-ext-wp-mysql-parser/target/release/libwp_mysql_parser.so)"
+          BASE_EXT="$(realpath ../../packages/php-ext-wp-mysql-parser/target/release/libwp_mysql_parser.so)"
           php -d extension="$BASE_EXT" tests/tools/run-native-ast-walk-benchmark.php --no-walk \
             | tee "$GITHUB_WORKSPACE/packages/mysql-on-sqlite/baseline-native-parse-only.txt"
 
--- a/.github/workflows/native-ast-perf.yml
+++ b/.github/workflows/native-ast-perf.yml
@@ -131,6 +131,48 @@ jobs:
           php -d extension="$BASE_EXT" tests/tools/run-native-ast-walk-benchmark.php --no-walk \
             | tee "$GITHUB_WORKSPACE/packages/mysql-on-sqlite/baseline-native-parse-only.txt"
 
+      # Hit-heavy scenarios — these are where the per-AST identity cache is
+      # supposed to win. The baseline reallocates wrappers on every accessor
+      # call, while the PR reuses them. Run on both to make the gap visible.
+      - name: Benchmark — native rewalk x10 (this PR)
+        working-directory: packages/mysql-on-sqlite
+        run: |
+          php -d extension="$NATIVE_EXT" tests/tools/run-native-ast-walk-benchmark.php --mode=rewalk --repeat=10 \
+            | tee native-rewalk.txt
+
+      - name: Benchmark — baseline rewalk x10
+        working-directory: ../baseline/packages/mysql-on-sqlite
+        run: |
+          BASE_EXT="$(realpath ../../packages/php-ext-wp-mysql-parser/target/release/libwp_mysql_parser.so)"
+          php -d extension="$BASE_EXT" tests/tools/run-native-ast-walk-benchmark.php --mode=rewalk --repeat=10 \
+            | tee "$GITHUB_WORKSPACE/packages/mysql-on-sqlite/baseline-native-rewalk.txt"
+
+      - name: Benchmark — native reread x20 (this PR)
+        working-directory: packages/mysql-on-sqlite
+        run: |
+          php -d extension="$NATIVE_EXT" tests/tools/run-native-ast-walk-benchmark.php --mode=reread --repeat=20 \
+            | tee native-reread.txt
+
+      - name: Benchmark — baseline reread x20
+        working-directory: ../baseline/packages/mysql-on-sqlite
+        run: |
+          BASE_EXT="$(realpath ../../packages/php-ext-wp-mysql-parser/target/release/libwp_mysql_parser.so)"
+          php -d extension="$BASE_EXT" tests/tools/run-native-ast-walk-benchmark.php --mode=reread --repeat=20 \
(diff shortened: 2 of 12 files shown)


adamziel:
## Perf measurement (CI run [25179776990](https://github.com/WordPress/sqlite-database-integration/actions/runs/25179776990))

Benchmarked on the same runner: parse the full MySQL server-suite corpus (69,567 queries) and walk every AST exhaustively via `get_descendants()` + a `get_first_child_node()` re-read, with and without the identity cache. The baseline is `codex/native-lazy-ast-facade` (PR base, no cache) built and run on the same hardware.

| scenario | duration | qps | peak mem | walked nodes | identity stable |
|---|---|---|---|---|---|
| pure-PHP, parse only | 10.71s | 6,495 | 68.0MB | — | true |
| pure-PHP, parse + walk | 12.84s | 5,419 | 70.0MB | 4,786,316 | true |
| native, parse only | 0.98s | 70,631 | 30.0MB | — | true |
| native, parse + walk **(baseline, no cache)** | 2.72s | 25,620 | 50.0MB | 4,786,316 | **FALSE** |
| native, parse + walk **(this PR, with cache)** | 3.18s | 21,867 | 62.5MB | 4,786,316 | true |

### Identity-cache cost on the walk path

| metric | baseline | this PR | delta |
|---|---|---|---|
| duration | 2.72s | 3.18s | **+0.47s (+17%)** |
| qps | 25,620 | 21,867 | **−15%** |
| peak memory | 50.0MB | 62.5MB | **+12.5MB (+25%)** |

The baseline identity check fails (`identity_ok=FALSE`) — that's the regression #381's review flagged, reproduced. With the cache it passes. The cost is one PHP wrapper retained per walked node, which on this corpus is ~4.8M nodes resolved into ~12.5MB of additional peak memory, plus the per-call hashmap lookup that adds 17% to the walk time. Parse-only timings are within run-to-run noise (0.98 vs 0.99s), confirming the cost only shows up when callers actually traverse — which is exactly the contract being preserved.

Net: the native walk is still **4×** faster than pure-PHP (3.18s vs 12.84s) and uses less memory, so the cache fits comfortably under the perf budget #381 set out to deliver while restoring `WP_Parser_Node` identity semantics.

adamziel:
## Perf after the hot-path optimization (CI run [25181172304](https://github.com/WordPress/sqlite-database-integration/actions/runs/25181172304), commit 6c94572)

Same corpus (69,567 queries, 4.8M walked nodes), same workflow, baseline rebuilt from `codex/native-lazy-ast-facade` on the same runner. Note this runner was a bit slower than the previous run overall (PHP walk drifted from 12.84s → 16.38s), so the right comparison is **within-run** baseline vs PR.

| scenario | duration | qps | peak mem | identity stable |
|---|---|---|---|---|
| native, parse only (baseline) | 1.28s | 54,142 | 30.0MB | true |
| native, parse only (this PR) | 1.28s | 54,352 | 30.0MB | true |
| native, parse + walk **(baseline, no cache)** | 3.52s | 19,766 | 50.0MB | **FALSE** |
| native, parse + walk **(this PR, optimized cache)** | 3.33s | 20,884 | 60.5MB | true |

### Identity-cache cost — before vs after the hot-path fix

| metric | before optimization | after optimization |
|---|---|---|
| walk duration delta | +17% (+0.47s) | **+5% (+0.19s)*** |
| walk qps delta | −15% | **−5%** |
| peak memory delta | +12.5MB (+25%) | +10.5MB (+21%) |

<sub>*on a different (slower) runner; both numbers are vs the same-run baseline.</sub>

### What changed

`intern_all()` previously called `intern()` per item, which re-resolved the cache and re-checked types on every one of the ~4.8M nodes walked. Now the cache reference and `&$cache->nodes` are hoisted out of the loop, the per-item logic is inlined, and a typed `intern_nodes()` fast path is used by `get_child_nodes()` / `get_descendant_nodes()` to skip the `instanceof` check entirely (the Rust bridge guarantees node-only results for those accessors).

Net: most of the cache cost on time is gone — the walk is now within ~5% of the no-identity baseline while still preserving `WP_Parser_Node` semantics. Native walk is now **~5×** faster than pure-PHP (3.33s vs 16.38s) on this runner. Memory delta is unchanged in shape (one PHP wrapper retained per visited node), as expected — the cache *is* the memory.

adamziel:
## Hit-heavy benchmarks (CI run [25189505049](https://github.com/WordPress/sqlite-database-integration/actions/runs/25189505049), commit 2c07dac)

Same corpus (69,567 queries), same runner, baseline = `codex/native-lazy-ast-facade` (no cache). All four scenarios run through the same script.

| scenario | baseline | this PR | time delta | mem delta | identity (baseline → PR) |
|---|---|---|---|---|---|
| **walk** (single full pass, 4.8M nodes) | 3.62s / 50.0MB | 3.52s / 60.5MB | **−3%** | +10.5MB | FALSE → true |
| **rewalk x10** (10 full passes, 47.9M visits) | 22.16s / 70.0MB | 18.37s / 88.5MB | **−17%** | +18.5MB | true → true |
| **reread x20** (root child × 20, 1.4M reads) | 2.04s / 30.0MB | 2.59s / 38.0MB | **+27%** | +8.0MB | FALSE → true |
| **subtree x5** (4.8M descendants × 5 child reads, 15.4M calls) | 11.68s / 50.0MB | 15.32s / 64.5MB | **+31%** | +14.5MB | true → true |

### Reading the numbers

The cache **wins where it's supposed to** — `rewalk` saves 17% wall time across 10 full re-traversals because the second-and-onward passes return cached wrappers instead of reallocating ~4.8M of them.

The cache **loses on `subtree`** — and this is the load-bearing finding. Subtree mode populates the cache to 4.8M entries during the first walk, then does 24M `get_first_child_node()` calls. Every one of those is a `WP_MySQL_Native_AST_Cache->nodes` lookup against a 4.8M-entry PHP hashmap. PHP's `zend_array` probe cost on a table that size, multiplied by 24M, dominates — and the per-call work the baseline avoids (just allocate, no bookkeeping) ends up cheaper than probing a fat hashmap. `reread` shows the same effect at smaller scale: per-call overhead beats raw allocation when the workload is tiny.

This is exactly the workload shape a translator/rewriter pass produces — walk once to find candidates, then re-enter subtrees by index. So while the correctness fix is right, **the PHP-side cache adds 27–31% to the realistic re-entry workload**.

### Would a Rust-side cache fix this?

Yes, materially. The two cost centers a Rust cache eliminates that PHP can't:

1. **`zend_array` probe replaced by `std::HashMap` probe.** Rust's hashmap probe on a 4.8M-entry table is roughly an order of magnitude cheaper per call than PHP's `zend_hash_index_find` — no zval boxing, no internal value type checks, better cache locality. On 24M probes (subtree mode), this is the difference between a multi-second tax and a sub-second one.
2. **Cache hit skips `ClassEntry::new()` + 4 `zend_update_property` calls.** Today even with the cache, Rust still allocates a wrapper before PHP gets to discard it. Moving the cache check into Rust means a hit returns a refcount-bumped pointer to the existing wrapper — no allocation, no property writes.

Estimated impact, holding everything else constant:
- `subtree` x5: +31% → likely **break even or single-digit %** vs baseline.
- `rewalk` x10: −17% → likely **−25 to −30%** (the win extends).
- `reread` x20: +27% → likely **break even** (per-call overhead drops below alloc cost).
- `walk` (miss-heavy): +5% → ~similar (misses still pay alloc + insert).

Memory delta is structural and unchanged — one retained wrapper per visited node, regardless of where the cache lives.

### Recommendation

The PHP-side cache is correctness-first and ships today, but it has a real performance regression on translator-style re-entry workloads (the same workloads this codebase actually runs). The Rust-side cache is the right architecture; the numbers above quantify the headroom (10–30% recoverable on hit-heavy scenarios).

Two paths:
- **Land #391, stack the Rust cache as a follow-up.** Correctness fix in users' hands now, perf win comes next.
- **Replace #391's implementation with the Rust cache.** Single PR, but blocks the correctness fix on a non-trivial Rust refactor that needs CI iteration to land safely.

PR #393, opened 2026-05-01
Restore `instanceof WP_Parser` for the native parser

adamziel:
Addresses Jan's note on #381: in native mode, `new WP_MySQL_Parser(...) instanceof WP_Parser` returns false because the native-mode class extends the Rust-registered `WP_MySQL_Native_Parser`, which has no `WP_Parser` in its chain. Existing downstream code doing `if ($parser instanceof WP_Parser)` silently skipped the parser whenever the extension was loaded.

This restores the contract by always extending the pure-PHP `WP_Parser` and pulling the native-mode behaviour in via a trait:

```php
class WP_MySQL_Parser extends WP_Parser {
    use WP_MySQL_Native_Parser_Impl;
}
```

`WP_MySQL_Native_Parser_Impl` owns the composed `WP_MySQL_Native_Parser` instance and the four-method delegation surface (`parse`, `next_query`, `get_query_ast`, `reset_tokens`). `WP_Parser`'s protected state (`$grammar`, `$tokens`, `$position`) is initialised by `parent::__construct` and stays inert — the trait's overrides never read it.

Adding a public method later means adding it to the trait — the class file itself is two lines and doesn't need touching.

## Why a trait, not a private property?

A bare property would also work, but the trait keeps the class file expressing only the routing decision (`extends WP_Parser` + `use Rust_Implementation;`). The implementation lives in one place, symmetric to where a future PHP-mode trait could live if we ever want to mirror the structure. Behaviour-wise the two are equivalent.

## Performance

The trait adds **one extra method-call frame per public-API call**. The public API is `parse()`, `next_query()`, `get_query_ast()`, `reset_tokens()` — called once per query. The actual parsing work happens inside the native call, so the delegation overhead is a small constant per query, not a multiplier on the parsing work.

The `Parser Delegation Perf` workflow runs `tests/tools/run-parser-benchmark.php` (parses the full MySQL server-suite corpus, ~70k queries) three times on this PR and three times on the PR base, on the same runner, with the extension loaded both times. The comparison goes into the job summary on every push.

## Test plan

- [x] PHP-only PHPUnit suite passes.
- [ ] PHPUnit suite passes with the native extension loaded; the new `WP_MySQL_Parser_Instanceof_Tests` confirm `instanceof WP_Parser` and `instanceof WP_MySQL_Parser` both hold.
- [ ] `Parser Delegation Perf` workflow shows the delegation cost is within noise.

### DIFF
--- a/.github/workflows/parser-delegation-perf.yml
+++ b/.github/workflows/parser-delegation-perf.yml
@@ -1,0 +1,146 @@
+name: Parser Delegation Perf
+
+# Measures the wall-time cost of routing the public parser API through a
+# composed `WP_MySQL_Native_Parser` instance (this PR) instead of letting
+# `WP_MySQL_Parser` directly extend the Rust class (PR base). The parse
+# benchmark walks the full MySQL server-suite corpus on both branches —
+# the delta is the per-call delegation cost.
+
+on:
+  push:
+    paths:
+      - '.github/workflows/parser-delegation-perf.yml'
+      - 'packages/mysql-on-sqlite/src/mysql/native/**'
+      - 'packages/mysql-on-sqlite/tests/tools/run-parser-benchmark.php'
+      - 'packages/php-ext-wp-mysql-parser/**'
+  pull_request:
+    paths:
+      - '.github/workflows/parser-delegation-perf.yml'
+      - 'packages/mysql-on-sqlite/src/mysql/native/**'
+      - 'packages/mysql-on-sqlite/tests/tools/run-parser-benchmark.php'
+      - 'packages/php-ext-wp-mysql-parser/**'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  perf:
+    name: Parser delegation benchmark
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.2'
+          coverage: none
+
+      - name: Set up Rust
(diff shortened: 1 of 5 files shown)


adamziel:
## Parser delegation perf

CI run [25216548485](https://github.com/WordPress/sqlite-database-integration/actions/runs/25216548485). Three back-to-back runs of `tests/tools/run-parser-benchmark.php` on each branch, same runner, native extension loaded both times, 69,577 queries from the MySQL server suite.

| run | baseline (`extends WP_MySQL_Native_Parser`) | this PR (`extends WP_Parser` + trait) |
|---|---|---|
| 1 | 14.740s / 4720 qps | 14.356s / 4846 qps |
| 2 | 14.659s / 4746 qps | 14.452s / 4814 qps |
| 3 | 14.556s / 4780 qps | 14.350s / 4848 qps |
| **median** | **14.659s / 4746 qps** | **14.404s / 4836 qps** |

Median delta: **−0.26s (−1.7%) duration, +1.9% qps** — i.e. the trait/composition variant measured *faster* than baseline. That's not a real win; the trait adds an extra `WP_MySQL_Native_Parser` instantiation per query plus one delegating PHP method-call frame on `parse()`, which can only be neutral or slower in principle. The 2% gap is **run-to-run noise** on a shared GHA runner — the dominant cost per query is the parsing work itself (~210µs/query at 4800 qps), and the delegation overhead is single-digit microseconds, well below what this benchmark resolves.

The honest summary: **the delegation cost is invisible at this granularity**, which is what we wanted to confirm before fixing the `instanceof WP_Parser` regression. The trait/composition pattern restores the contract Jan flagged at no measurable perf cost.

### Other CI

`PHP 8.2 / Rust extension` and `SQLite driver / Rust extension` both pass on the latest commit, which means the new `WP_MySQL_Parser_Instanceof_Tests` regression test runs and verifies `instanceof WP_Parser` and `instanceof WP_MySQL_Parser` both hold under the extension.

PR #405, opened 2026-05-26
Default SQLite connections to WAL

JanJakes:
## What changed

**[Journal mode:](https://sqlite.org/pragma.html#pragma_journal_mode)**
  - Default new SQLite connections to **`journal_mode = WAL`**.
  - Skip defaulting to WAL when unavailable (e.g., on network filesystems).

**[Synchronous flag:](https://sqlite.org/pragma.html#pragma_synchronous)**
  - Apply **`synchronous = NORMAL`** when WAL is used.
  - Otherwise, use the SQLite default.

**Other changes:**
- Wire the `SQLITE_JOURNAL_MODE` override through the WordPress plugin connection paths.
- Accept `journal_mode` and `synchronous` in the `WP_PDO_MySQL_On_SQLite` driver options.
- Add connection setup tests for the defaults, overrides, invalid values, and the WAL fallback.

## Why

`WAL` improves read/write concurrency for SQLite-backed WordPress installs, and `synchronous = NORMAL` avoids frequent sync to the main database. In WAL mode, `NORMAL` is safe. [From the SQLite docs](https://sqlite.org/pragma.html#pragma_synchronous):

> [WAL mode](https://sqlite.org/wal.html) is safe from corruption with synchronous=NORMAL, and probably DELETE mode is safe too on modern filesystems. WAL mode is always consistent with synchronous=NORMAL, but WAL mode does lose durability. A transaction committed in WAL mode with synchronous=NORMAL might roll back following a power loss or system crash. Transactions are durable across application crashes regardless of the synchronous setting or journal mode.

> The synchronous=NORMAL setting provides the best balance between performance and safety for most applications running in [WAL mode](https://sqlite.org/wal.html). You lose durability across power lose with synchronous NORMAL in WAL mode, but that is not important for most applications. Transactions are still atomic, consistent, and isolated, which are the most important characteristics in most use cases.

## Benchmark

A local benchmark ran a WordPress-like 90% read / 10% write workload against the same database file with 4, 8, and 16 concurrent workers, comparing the trunk defaults with WAL + NORMAL (median of 3 runs, PHP 8.5.5, SQLite 3.53.0).

| Workers | Config | Ops/s | p50 | p95 | p99 |
|---|---|---:|---:|---:|---:|
| 4 | default | 4,985 | 0.21 ms | 3.94 ms | 10.2 ms |
| 4 | WAL + NORMAL | 15,982 (3.2×) | 0.19 ms | 0.73 ms | 1.2 ms |
| 8 | default | 5,249 | 0.24 ms | 4.22 ms | 22.6 ms |
| 8 | WAL + NORMAL | 25,371 (4.8×) | 0.20 ms | 1.02 ms | 1.5 ms |
| 16 | default | 5,226 | 0.25 ms | 10.25 ms | 59.0 ms |
| 16 | WAL + NORMAL | 30,300 (5.8×) | 0.22 ms | 1.43 ms | 4.0 ms |

This was run on an M4 MacBook Pro Max.

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-connection.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-connection.php
@@ -113,18 +113,21 @@ class WP_SQLite_Connection {
 		if ( is_string( $journal_mode ) ) {
 			$journal_mode = strtoupper( $journal_mode );
 		}
-		if ( $journal_mode && in_array( $journal_mode, self::SQLITE_JOURNAL_MODES, true ) ) {
-			try {
-				$effective_journal_mode = strtoupper(
-					(string) $this->query( 'PRAGMA journal_mode = ' . $journal_mode )->fetchColumn()
-				);
-			} catch ( PDOException $e ) {
-				// WAL may be unavailable in some environments, such as on network
-				// filesystems. When it is explicitly configured, surface the error.
-				// Otherwise, fall back to the default SQLite behavior.
-				if ( isset( $options['journal_mode'] ) ) {
-					throw $e;
-				}
+		if ( ! in_array( $journal_mode, self::SQLITE_JOURNAL_MODES, true ) ) {
+			throw new InvalidArgumentException(
+				sprintf( 'Invalid SQLite journal mode: %s.', $options['journal_mode'] )
+			);
+		}
+		try {
+			$effective_journal_mode = strtoupper(
+				(string) $this->query( 'PRAGMA journal_mode = ' . $journal_mode )->fetchColumn()
+			);
+		} catch ( PDOException $e ) {
+			// WAL may be unavailable in some environments, such as on network
+			// filesystems. When it is explicitly configured, surface the error.
+			// Otherwise, fall back to the default SQLite behavior.
+			if ( isset( $options['journal_mode'] ) ) {
+				throw $e;
 			}
 		}
 
@@ -154,6 +157,11 @@ class WP_SQLite_Connection {
 				$synchronous = self::SQLITE_SYNCHRONOUS_SETTINGS[ $synchronous ];
 			} elseif ( is_string( $synchronous ) ) {
 				$synchronous = strtoupper( $synchronous );
+			}
+			if ( ! in_array( $synchronous, self::SQLITE_SYNCHRONOUS_SETTINGS, true ) ) {
+				throw new InvalidArgumentException(
+					sprintf( 'Invalid SQLite synchronous setting: %s.', $options['synchronous'] )
+				);
 			}
 		} elseif ( 'WAL' === $effective_journal_mode ) {
 			// Default to NORMAL for WAL mode.
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-connection.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-connection.php
@@ -46,6 +46,8 @@ class WP_SQLite_Connection {
 	/**
 	 * The supported SQLite synchronous settings.
 	 *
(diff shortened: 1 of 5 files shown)


packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-connection.php
```diff
@@ -92,9 +118,38 @@ public function __construct( array $options ) {
...
+					throw $e;
+				}
+			}
+		}
+
+		// Configure SQLite synchronous setting. In WAL mode, default to NORMAL.
+		// Otherwise, use SQLite's default value.
+		$synchronous = $options['synchronous'] ?? null;
+		if ( null === $synchronous && 'WAL' === $effective_journal_mode ) {
+			$synchronous = self::DEFAULT_SQLITE_WAL_SYNCHRONOUS;
```
adamziel: This is super confusing. It's SYNCHRONOUS but it's NORMAL. Let's just stick to SQLite terminology and maybe even let go of constants:

```suggestion
			$synchronous = 'NORMAL';
```

Or even go super simple:

```suggestion
			$this->query( 'PRAGMA synchronous = NORMAL' );
```

The comment above explains what we're doing, I assume it's because the logic is so opaque here. I'd love to, instead, understand why we're doing this. Document the NORMAL mode inline and explain why we're defaulting to it when WAL is enabled and not otherwise and what is SQLite default value?

adamziel (changes requested):

packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-connection.php
JanJakes: @adamziel Good point, thanks! Improved in https://github.com/WordPress/sqlite-database-integration/pull/405/changes/a15ecdbdc5be13a1d7769c09be1dfbb5dc68b9b3.

packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-connection.php
```diff
@@ -91,10 +107,60 @@ public function __construct( array $options ) {
...
+			} elseif ( is_string( $synchronous ) ) {
+				$synchronous = strtoupper( $synchronous );
+			}
+		} elseif ( 'WAL' === $effective_journal_mode ) {
+			// Default to NORMAL for WAL mode.
+			$synchronous = 'NORMAL';
+		}
+		if ( in_array( $synchronous, self::SQLITE_SYNCHRONOUS_SETTINGS, true ) ) {
+			$this->query( 'PRAGMA synchronous = ' . $synchronous );
 		}
```
adamziel, line 172:
```suggestion
		if ( in_array( $synchronous, self::SQLITE_SYNCHRONOUS_SETTINGS, true ) ) {
			$this->query( 'PRAGMA synchronous = ' . $synchronous );
		} else if ( isset( $synchronous ) ) {
			throw new PDOException( sprintf( '%s is not a valid `synchronous` PRAGMA value. ' ) );
		}
```

adamziel: Let's not accept invalid option values.

adamziel: Ditto for the journal mode, let's not skip that silently.

JanJakes: @adamziel Fixed in https://github.com/WordPress/sqlite-database-integration/pull/405/changes/8de3d1a787d658ba565ffe5a4fa41707aba463ed. I used `InvalidArgumentException` as in the case of missing path, but not sure which is better—vs `PDOException`.

PR #406, opened 2026-05-26
CI baseline for WAL comparison

JanJakes:
Temporary draft PR used to collect trunk CI timing samples for WAL comparison.

This branch contains only an empty commit on top of trunk, so the tree is identical to trunk. I will close it after collecting the requested timing data.


JanJakes:
Closed after collecting trunk baseline CI timing samples for #405.

PR #434, opened 2026-06-15
Add REVERSE() to user defined functions

teizz:
I ran into an issue where a plugin is using the REVERSE() function to add the reverse of a string into the database. Since this isn't supported by SQLite it breaks the plugin. This fixes it for me.

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-pdo-user-defined-functions.php
@@ -90,6 +90,7 @@ class WP_SQLite_PDO_User_Defined_Functions {
 		'utc_time'                     => 'utc_time',
 		'utc_timestamp'                => 'utc_timestamp',
 		'version'                      => 'version',
+		'reverse'                      => 'reverse',
 
 		// Internal helper functions.
 		'_helper_like_to_glob_pattern' => '_helper_like_to_glob_pattern',
@@ -927,6 +928,22 @@ class WP_SQLite_PDO_User_Defined_Functions {
 	 */
 	public function version() {
 		return '5.5';
+	}
+
+	/**
+	 * Method to emulate MySQL REVERSE() function.
+	 *
+	 * Takes a string and returns the reverse of it.
+	 *
+	* @param string|null $str The string to reverse.
+	*
+	* @return string|null reversed string, or NULL.
+	*/
+	public function reverse( $str ) {
+		if ( null === $str ) {
+			return null;
+		}
+		return strrev( $str );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


adamziel:
Thank you @teizz !

JanJakes:
Hi @teizz, thanks for implementing this!

One question: This implementation doesn't handle multibyte strings, so calling `REVERSE()` on a multi-byte Unicode string will corrupt the string. Could we address that in a follow-up, without strictly requiring any additional PHP extensions to be installed? I haven't looked into how hard this is to solve universally.

teizz:
Hi @JanJakes!

I had actually given that some thought. I couldn't tell if the the plugin that broke for me really needed anything more than byte reversal. I think it's only using it as some index speed optimization. I also wasn't sure to what extend MySQL would actually handle unicode (especially combining characters and other grapheme clusters) and if you're ideally want some sort of bug-for-bug compatibility.

In any case, I did have a couple of "workarounds" to get full unicode reversal without needing additional PHP extensions and I guess there's two choices:
  * Reversal of just the code points, which works for both newer and older versions of PHP regardless of the version of PCRE that was compiled with. This will reverse most UTF-8 characters including emojis, just not the ones that are multiple code points combined.
  * Full reversal of all unicode characters, but requires PRCE 10.44 (07-June-2024) because the older ones will glob multiple characters together if they all have multiple code points essentially treating a string of emojis as one big long emoji.
  
The first option is simply
```
function reverse($str) {
    if (null === $str) {
        return null;
    }
    $chars = preg_split('//u', $str, -1, PREG_SPLIT_NO_EMPTY);
    return implode('', array_reverse($chars));
}
```

The second which is very elegant in my opinion, but does require PHP 8.4 or newer (since that includes PRCE 10.44) to not see a sequence of multiple characters with multiple code points as a single very long character.
```
function reverse($str) {
    if ($str === null) {
        return null;
    }
    $chars = preg_split('/(\X)/u', $str, -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY);
    return implode('', array_reverse($chars));
}
```

Both these functions don't break the plugin I need. Feel free to swap the `strrev()` for either of these 😇

JanJakes:
Thanks, @teizz!

Ah, there is no `mb_strrev` so no fast path is possible when mbstring is available. I guess `preg_split('//u', $str, -1, PREG_SPLIT_NO_EMPTY)` is reasonable, but we should probably first check whether the string contains any multi-byte characters at all (sth like `preg_match('/[^\x00-\x7F]/u', $str)`) to avoid allocating arrays for simple strings. The fast path would use `strrev()` and multi-byte could use the `preg_split` approach.

teizz:
I was surprised there isn't a `mb_strrev` as well! Totally agree there should be a fastpath here if we're doing slightly more expensive operations. And while were optimizing, I did a quick check and doing a `preg_match_all` on any character and then just reversing and imploding the `match[0]` is actually twice as fast as doing a `preg_split` on every character, so lets choose the faster approach there. So the end result would be:
``` php
    public function reverse($str) {
        if (null === $str) {
            return null;
        }
        if(preg_match('/[^\x00-\x7F]/', $str)) {
            preg_match_all('/./u', $str, $matches);
            return implode('', array_reverse($matches[0]));
        }
        return strrev($str);
    }
```

Not sure on the PR etiquette here. Would you prefer me opening another PR, or are you wanting to patch this in yourself? Either is fine with me :)

PR #439, opened 2026-06-18
Normalize BIT column default literals

JanJakes:
## What

BIT column defaults were stored and rendered incorrectly. A column such as `BIT(1) DEFAULT '0'` (e.g., as produced by Ninja Forms) emitted an invalid `SHOW CREATE TABLE` / dump line:

```sql
`maintenance` bit(1) DEFAULT '0'
```

MySQL does not accept a quoted-string default on a BIT column—the value has to be a bit literal (`b'0'`) or a number (`0`). The invalid output broke dump/restore flows (e.g., Studio → Pressable).

## Fix

Normalize BIT column default literals end to end:

- **Store:** Record BIT defaults in the information schema in a normalized format (`b'...'`).
- **`SHOW CREATE TABLE`:** Emit the stored bit literal verbatim (`bit(1) DEFAULT b'0'`).
- **Internal SQLite DDL / `column = DEFAULT`:** Render the bit literal as the SQLite integer it maps to.
- **Legacy migration:** The information-schema reconstructor converts integer BIT defaults into a bit literal.

## Tests

- `SHOW CREATE TABLE` output **and** the inserted/read-back values for every default form (quoted string, integer, `b'…'`, `0b…`, `x'…'`, `TRUE`, `FALSE`).
- `UPDATE … SET col = DEFAULT` on a BIT column.
- Reconstruction of BIT defaults for legacy databases.
- Full unit suite green; coding standards (PHPCS) pass.

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-information-schema-builder.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-information-schema-builder.php
@@ -1503,13 +1503,14 @@ class WP_SQLite_Information_Schema_Builder {
 	 * @return array                       Column data for the information schema.
 	 */
 	private function extract_column_data( string $table_name, string $column_name, WP_Parser_Node $node, int $position ): array {
-		$default  = $this->get_column_default( $node );
+		list ( $data_type, $column_type ) = $this->get_column_data_types( $node );
+
+		$default  = $this->get_column_default( $node, $data_type );
 		$nullable = $this->get_column_nullable( $node );
 		$key      = $this->get_column_key( $node );
 		$extra    = $this->get_column_extra( $node );
 		$comment  = $this->get_column_comment( $node );
 
-		list ( $data_type, $column_type )    = $this->get_column_data_types( $node );
 		list ( $charset, $collation )        = $this->get_column_charset_and_collation( $node, $data_type );
 		list ( $char_length, $octet_length ) = $this->get_column_lengths( $node, $data_type, $charset );
 		list ( $precision, $scale )          = $this->get_column_numeric_attributes( $node, $data_type );
@@ -2040,10 +2041,11 @@ class WP_SQLite_Information_Schema_Builder {
 	/**
 	 * Extract column default value from the "columnDefinition" or "fieldDefinition" AST node.
 	 *
-	 * @param  WP_Parser_Node $node The "columnDefinition" or "fieldDefinition" AST node.
-	 * @return string               The column default as stored in information schema.
+	 * @param  WP_Parser_Node $node      The "columnDefinition" or "fieldDefinition" AST node.
+	 * @param  string         $data_type The column data type as stored in information schema.
+	 * @return string|null               The column default as stored in information schema.
 	 */
-	private function get_column_default( WP_Parser_Node $node ): ?string {
+	private function get_column_default( WP_Parser_Node $node, string $data_type ): ?string {
 		$default_attr = null;
 		foreach ( $node->get_descendant_nodes( 'columnAttribute' ) as $attr ) {
 			if ( $attr->has_child_token( WP_MySQL_Lexer::DEFAULT_SYMBOL ) ) {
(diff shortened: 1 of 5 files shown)


packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-information-schema-reconstructor.php
```diff
@@ -496,6 +496,11 @@ private function generate_column_default( string $mysql_type, ?string $default_v
 		}
 		$mysql_type = strtolower( $mysql_type );
 
+		if ( str_starts_with( $mysql_type, 'bit' ) ) {
```
akirk, line 499: Could this match something else starting with `bit`?

packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-information-schema-builder.php
```diff
@@ -2099,6 +2091,74 @@ private function get_column_default( WP_Parser_Node $node ): ?string {
...
+
+		// Bit literal, e.g. b'101' or 0b101.
+		if ( str_starts_with( $value, "b'" ) || str_starts_with( $value, '0b' ) ) {
+			$bits = ltrim( rtrim( substr( $value, 2 ), "'" ), '0' );
+			return "b'" . ( '' === $bits ? '0' : $bits ) . "'";
+		}
+
+		// Hex literal, e.g. x'05' or 0x05.
+		if ( str_starts_with( $value, "x'" ) || str_starts_with( $value, '0x' ) ) {
+			return "b'" . decbin( hexdec( rtrim( substr( $value, 2 ), "'" ) ) ) . "'";
```
zaerl, line 2151: Didn't run, but I think this can potentially accept `0x05'`, but not a problem.

packages/mysql-on-sqlite/tests/WP_SQLite_Driver_Tests.php
```diff
@@ -658,6 +658,66 @@ public function testShowCreateTableWithDefaultValues(): void {
...
 
+	public function testShowCreateTableWithBitDefaultValues(): void {
+		$this->assertQuery(
+			"CREATE TABLE _tmp_bit_defaults (
+				id INT DEFAULT 0,
+				quoted_zero BIT(1) DEFAULT '0',
+				integer_five BIT(4) DEFAULT 5,
+				bit_literal_five BIT(4) DEFAULT b'0101',
+				binary_number_five BIT(4) DEFAULT 0b0101,
+				hex_literal_five BIT(4) DEFAULT x'05',
```
zaerl, line 669: Should we also add a `hex_five BIT(4) DEFAULT 0x05` line?

packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-information-schema-builder.php
JanJakes: @zaerl Here we're operating on literals that the MySQL grammar recognizes, so I think that such a string cannot appear here. The lexer would attempt to read a HEX number due to the prefix and fail on the quote.

packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-information-schema-reconstructor.php
JanJakes: @akirk There is no other type starting with `bit` in MySQL, so this should be safe. It's matching on a prefix to catch the full `bit(4)` syntax that is stored in the information schema.

packages/mysql-on-sqlite/tests/WP_SQLite_Driver_Tests.php
JanJakes: @zaerl Good idea, added in https://github.com/WordPress/sqlite-database-integration/pull/439/changes/ab6385a085b0dda46e3277171241f0dc587464a0.

zaerl (approved):

PR #361, opened 2026-04-13
Improve concurrent database access

JanJakes:
## Summary

Reduces lock contention under concurrent access by avoiding unnecessary write locks for read-only queries.

1. **Skip the wrapper transaction for SELECT statements.** A standalone SELECT translates to a single SQLite query, so the implicit autocommit transaction provides sufficient consistency — no explicit `BEGIN`/`COMMIT` needed.
2. **Use deferred `BEGIN` for other read-only wrapper transactions.** SHOW and DESCRIBE queries still use a wrapper transaction (they may translate to multiple SQLite queries), but now use a deferred `BEGIN` (SHARED lock) instead of `BEGIN IMMEDIATE` (RESERVED lock).

### Background

The wrapper transaction previously used `BEGIN IMMEDIATE` for all queries, including read-only SELECTs. This acquired a RESERVED (write) lock on every query — but SQLite only allows one RESERVED lock at a time. Under concurrent load (e.g., multiple PHP-FPM workers handling requests simultaneously), all processes competed for the single write lock. When the 10-second busy timeout was exceeded, SQLite returned `SQLITE_BUSY` ("database is locked", error 5).

SHARED locks, acquired by a deferred `BEGIN`, are compatible with RESERVED locks — multiple readers can proceed concurrently alongside an active writer.

Fixes #318

PR #443, opened 2026-06-29
Don't take a write lock for SET statements (fixes "database is locked" on connect)

### LINKED ISSUE #318: Address `database is locked` error
Certain combination of plugins can trigger a `database is locked` error. Here's an example log:

```
[04-Feb-2026 15:59:07 UTC] WordPress database error <div style="clear:both">&nbsp;</div>
<div class="queries" style="clear:both;margin-bottom:2px;border:red dotted thin;">
<p>MySQL query:</p>
<p>SELECT option_value FROM wp_options WHERE option_name = 'elementor_experiment-home_screen' LIMIT 1</p>
<p>Queries made or created this session were:</p>
<ol>
<li>Executing: BEGIN IMMEDIATE | (no parameters)</li>
</ol>
</div>
<div style="clear:both;margin-bottom:2px;border:red dotted thin;" class="error_message" style="border-bottom:dotted blue thin;">
SQLSTATE[HY000]: General error: 5 database is locked
</div>
<p>Backtrace:</p>
<pre>#0 /wordpress/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite-ast/class-wp-pdo-mysql-on-sqlite.php(921): WP_PDO_MySQL_On_SQLite->new_driver_exception('SQLSTATE[HY000]...', 'HY000', Object(PDOException))
#1 /wordpress/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite-ast/class-wp-sqlite-driver.php(160): WP_PDO_MySQL_On_SQLite->query('SELECT option_v...', 5)
#2 /wordpress/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php(524): WP_SQLite_Driver->query('SELECT option_v...')
#3 /wordpress/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php(436): WP_SQLite_DB->_do_query('SELECT option_v...')
#4 /wordpress/wp-includes/class-wpdb.php(3075): WP_SQLite_DB->query('SELECT option_v...')
#5 /wordpress/wp-includes/option.php(206): wpdb->get_row('SELECT option_v...')
#6 /wordpress/wp-content/plugins/elementor/core/experiments/manager.php(706): get_option('elementor_exper...')
#7 /wordpress/wp-content/plugins/elementor/core/experiments/manager.php(80): Elementor\Core\Experiments\Manager->get_saved_feature_state('home_screen')
#8 /wordpress/wp-content/plugins/elementor/core/modules-manager.php(56): Elementor\Core\Experiments\Manager->add_feature(Array)
#9 /wordpress/wp-content/plugins/elementor/includes/plugin.php(703): Elementor\Core\Modules_Manager->__construct()
#10 /wordpress/wp-content/plugins/elementor/includes/plugin.php(621): Elementor\Plugin->init_components()
#11 /wordpress/wp-includes/class-wp-hook.php(341): Elementor\Plugin->init('')
#12 /wordpress/wp-includes/class-wp-hook.php(365): WP_Hook->apply_filters(NULL, Array)
#13 /wordpress/wp-includes/plugin.php(522): WP_Hook->do_action(Array)
#14 /wordpress/wp-settings.php(742): do_action('init')
#15 /wordpress/wp-config.php(86): require_once('/wordpress/wp-s...')
#16 /wordpress/wp-load.php(50): require_once('/wordpress/wp-c...')
#17 /wordpress/wp-blog-header.php(13): require_once('/wordpress/wp-l...')
#18 /wordpress/index.php(17): require('/wordpress/wp-b...')
#19 {main}</pre>
 for query SELECT option_value FROM wp_options WHERE option_name = 'elementor_experiment-home_screen' LIMIT 1 made by require('wp-blog-header.php'), require_once('wp-load.php'), require_once('wp-config.php'), require_once('wp-settings.php'), do_action('init'), WP_Hook->do_action, WP_Hook->apply_filters, Elementor\Plugin->init, Elementor\Plugin->init_components, Elementor\Core\Modules_Manager->__construct, Elementor\Core\Experiments\Manager->add_feature, Elementor\Core\Experiments\Manager->get_saved_feature_state, get_option, WP_SQLite_DB->query, WP_SQLite_DB->print_error
```

Please see the details in STU-1285.

gcsecsey:
Related to STU-1821

## What changed

Every WordPress request runs `SET SESSION sql_mode = …` at bootstrap (via `wpdb::set_sql_mode()`). This statement changes only session state and writes nothing to the database, but the driver acquires an SQLite write lock for it. Under concurrent access, this makes an otherwise read-only request fatal with `database is locked` (`SQLITE_BUSY`) the moment any other connection holds the write lock.

This PR fixes the two places the `SET` path takes a write lock (`WP_PDO_MySQL_On_SQLite`):

1. **The wrapper transaction.** `query()` opens `BEGIN IMMEDIATE` (a RESERVED/write lock) for every non-`SELECT` statement, and `SET` was treated as a write. → `SET` is now detected as read-only, so it opens a deferred `BEGIN` (SHARED) instead — the same treatment `SHOW`/`DESCRIBE` already receive (#361).

2. **The empty-result builder.** `create_result_statement_from_data()` fabricates a 0-column result with a no-op `INSERT … WHERE FALSE` against a regular table. SQLite acquires a write lock for *any* DML, even one that inserts zero rows, so this re-took a write lock right after the wrapper committed. → The 0-column statement is now built with a no-op `INSERT` into a connection-private **`TEMP` table**, whose temp database has no shared lock.

Both changes are required: with only (1), the fatal simply moves from the `BEGIN IMMEDIATE` to the no-op `INSERT`.

## Why

This is a direct follow-up to #361 (which stopped `SELECT`/`SHOW`/`DESCRIBE` from taking write locks and closed #318). `SET` was left on the write path. Because `set_sql_mode()` runs on every connection, the impact is broad: a single slow request holding a write transaction (a background WooCommerce/Jetpack write, an import, a long render on a heavy site) makes every concurrent request fail at connect, not only concurrent writers. SQLite is single-writer by design, but statements that perform no writes should never contend for the write lock.

Journal mode does not address this on its own: it reproduces identically under both `DELETE` and `WAL` (WAL still serialises writers).

## Reproduction

Deterministic, single process, two driver connections to one file:

```php
$path = tempnam( sys_get_temp_dir(), 'sqlite' );
@unlink( $path );

$make = fn() => new WP_SQLite_Driver( new WP_SQLite_Connection( array( 'path' => $path ) ), 'db' );

// Build both connections first, so B's schema setup doesn't run under the lock.
$a = $make();
$a->query( 'CREATE TABLE t ( id INTEGER PRIMARY KEY )' );
$b = $make();
$b->execute_sqlite_query( 'PRAGMA busy_timeout = 1500' ); // fail fast on a lock

// A holds a write lock.
$a->query( 'START TRANSACTION' );
$a->query( 'INSERT INTO t ( id ) VALUES ( 1 )' );

// B runs the connect-time statement. It writes nothing, so it must not block.
$b->query( "SET SESSION sql_mode = 'NO_ENGINE_SUBSTITUTION'" );
// Before this PR: throws "database is locked". After: returns immediately.
```

**Before:** `WP_SQLite_Driver_Exception: … database is locked`, first in `begin_wrapper_transaction()`, or after fixing only (1), in `create_result_statement_from_data()`.

**After:** the `SET` returns immediately while A still holds its write lock.

This also reproduces end-to-end in a multi-worker WordPress install (eg. PHP-FPM with more than one workers) under concurrent requests, where one slow request holding a write transaction makes concurrent page loads fatal at `set_sql_mode`.

## Tests

Two regression tests added to `WP_SQLite_Driver_Concurrency_Tests`:

- `testSetQueryOpensReadOnlyTransaction`: `SET` opens `BEGIN`, not `BEGIN IMMEDIATE`.
- `testSetQuerySucceedsWhileAnotherConnectionHoldsWriteLock`: with another connection holding `BEGIN IMMEDIATE` and `timeout = 0`, `SET` completes instead of throwing `SQLITE_BUSY`.

Both fail on `trunk` and pass with this change.

## Notes / risk

- **The 0-column contract is preserved.** `WP_SQLite_Driver::query()` returns `fetchAll()` when `columnCount() > 0`, otherwise `rowCount()`. The replacement empty-result statement keeps `columnCount() === 0`, so writes and `SET` still return their affected-row count (not an empty array). Verified: running a representative `CREATE`/`INSERT`/`SELECT`/`COUNT`/`UPDATE`/`DELETE`/`SET` sequence produces byte-identical results before and after.
- **Bonus:** real writes (`INSERT`/`UPDATE`/`DELETE`) also fabricate their empty result through this path, which previously meant a second, redundant write-lock acquisition after the real write. That redundant lock is now gone, slightly reducing lock pressure for all writes.
- **Temp-table lifetime.** The `TEMP` table is created once per connection (guarded by a flag); `TEMP` tables live for the connection's lifetime. A reviewer preferring maximum robustness over the micro-optimisation could drop the flag and always issue the idempotent `CREATE TEMP TABLE IF NOT EXISTS`.

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
@@ -44,6 +44,12 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 	const GLOBAL_VARIABLES_TABLE_NAME = self::RESERVED_PREFIX . 'global_variables';
 
 	/**
+	 * Name of the connection-private TEMP table used to build empty result sets
+	 * without acquiring a write lock on the database. See create_result_statement_from_data().
+	 */
+	const EMPTY_RESULT_TABLE_NAME = self::RESERVED_PREFIX . 'empty_result';
+
+	/**
 	 * The name of the SQLite driver version variable.
 	 *
 	 * This internal variable is used to store the latest version of the SQLite
@@ -528,6 +534,14 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 	private $is_readonly;
 
 	/**
+	 * Whether the TEMP table backing empty result sets has been created on this
+	 * connection. See create_result_statement_from_data().
+	 *
+	 * @var bool
+	 */
+	private $empty_result_table_ready = false;
+
+	/**
 	 * Type of wrapper transaction that is active for the MySQL query emulation.
 	 *
 	 * Possible values:
@@ -920,6 +934,11 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 				if (
 					'selectStatement' === $statement_node->rule_name
 					|| 'showStatement' === $statement_node->rule_name
+					// A SET statement only changes session state (sql_mode, etc.)
+					// and writes nothing to the database, so it must not take a
+					// write lock. Treat it as read-only -> deferred BEGIN instead
+					// of BEGIN IMMEDIATE. (Mirrors the SHOW/DESCRIBE handling.)
+					|| 'setStatement' === $statement_node->rule_name
 				) {
 					$this->is_readonly = true;
 				} elseif ( 'utilityStatement' === $statement_node->rule_name ) {
@@ -7003,10 +7022,29 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 		 * This can be done using a noop INSERT statement that modifies no data.
 		 */
 		if ( 0 === count( $columns ) ) {
+			/*
+			 * Build a 0-column statement to represent an empty result set.
+			 *
+			 * A no-op "INSERT ... WHERE FALSE" into a regular table still acquires
+			 * SQLite's write lock, even though it writes nothing. That makes
+			 * no-write statements such as "SET sql_mode = ..." fatal with
+			 * "database is locked" whenever another connection holds the write
+			 * lock. Writing to a connection-private TEMP table uses the temp
+			 * database, which has no shared lock, so it never contends.
+			 */
+			if ( ! $this->empty_result_table_ready ) {
+				$pdo->exec(
+					sprintf(
+						'CREATE TEMP TABLE IF NOT EXISTS %s ( x )',
+						$this->quote_sqlite_identifier( self::EMPTY_RESULT_TABLE_NAME )
+					)
+				);
+				$this->empty_result_table_ready = true;
+			}
 			return $pdo->query(
 				sprintf(
-					'INSERT INTO %s (rowid) SELECT NULL WHERE FALSE',
-					$this->quote_sqlite_identifier( self::GLOBAL_VARIABLES_TABLE_NAME )
+					'INSERT INTO %s ( x ) SELECT NULL WHERE FALSE',
+					$this->quote_sqlite_identifier( self::EMPTY_RESULT_TABLE_NAME )
 				)
 			);
 		}
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
@@ -934,10 +934,11 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 				if (
 					'selectStatement' === $statement_node->rule_name
 					|| 'showStatement' === $statement_node->rule_name
-					// A SET statement only changes session state (sql_mode, etc.)
-					// and writes nothing to the database, so it must not take a
-					// write lock. Treat it as read-only -> deferred BEGIN instead
-					// of BEGIN IMMEDIATE. (Mirrors the SHOW/DESCRIBE handling.)
+					// Supported SET statements mutate only connection-local driver state
+					// (sql_mode, user variables, etc.) or fail before touching SQLite.
+					// They do not write to the SQLite database, so they must not take a
+					// write lock. Use a deferred BEGIN instead of BEGIN IMMEDIATE.
+					// This mirrors the SHOW/DESCRIBE handling.
 					|| 'setStatement' === $statement_node->rule_name
 				) {
 					$this->is_readonly = true;
@@ -7035,7 +7036,7 @@ class WP_PDO_MySQL_On_SQLite extends PDO {
 			if ( ! $this->empty_result_table_ready ) {
 				$pdo->exec(
 					sprintf(
-						'CREATE TEMP TABLE IF NOT EXISTS %s ( x )',
+						'CREATE TEMP TABLE IF NOT EXISTS %s ( placeholder_for_empty_insert )',
 						$this->quote_sqlite_identifier( self::EMPTY_RESULT_TABLE_NAME )
 					)
 				);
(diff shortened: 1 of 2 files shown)


packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -920,6 +934,11 @@ public function query( string $query, ?int $fetch_mode = null, ...$fetch_mode_ar
 				if (
 					'selectStatement' === $statement_node->rule_name
 					|| 'showStatement' === $statement_node->rule_name
+					// A SET statement only changes session state (sql_mode, etc.)
+					// and writes nothing to the database, so it must not take a
+					// write lock. Treat it as read-only -> deferred BEGIN instead
+					// of BEGIN IMMEDIATE. (Mirrors the SHOW/DESCRIBE handling.)
+					|| 'setStatement' === $statement_node->rule_name
```
adamziel: This is accurate for the driver’s currently supported `SET` behavior, but too broad as a statement about MySQL generally (`SET PERSIST` / `PERSIST_ONLY` write `mysqld-auto.cnf` which affect more than the current session's state). Could we narrow the comment to the implementation invariant?

```suggestion
					// Supported SET statements mutate only connection-local driver state
					// (sql_mode, user variables, etc.) or fail before touching SQLite.
					// They do not write to the SQLite database, so they must not take a
					// write lock. Use a deferred BEGIN instead of BEGIN IMMEDIATE.
					// This mirrors the SHOW/DESCRIBE handling.
					|| 'setStatement' === $statement_node->rule_name
```

packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -7003,10 +7022,29 @@ private function create_result_statement_from_data( array $columns, array $rows
...
+			 * SQLite's write lock, even though it writes nothing. That makes
+			 * no-write statements such as "SET sql_mode = ..." fatal with
+			 * "database is locked" whenever another connection holds the write
+			 * lock. Writing to a connection-private TEMP table uses the temp
+			 * database, which has no shared lock, so it never contends.
+			 */
+			if ( ! $this->empty_result_table_ready ) {
+				$pdo->exec(
+					sprintf(
+						'CREATE TEMP TABLE IF NOT EXISTS %s ( x )',
```
adamziel: Could we make the placeholder column self-documenting? `x` makes the statement harder to parse at a glance.

```suggestion
						'CREATE TEMP TABLE IF NOT EXISTS %s ( placeholder_for_empty_insert )',
```

packages/mysql-on-sqlite/src/sqlite/class-wp-pdo-mysql-on-sqlite.php
```diff
@@ -7003,10 +7022,29 @@ private function create_result_statement_from_data( array $columns, array $rows
...
+						$this->quote_sqlite_identifier( self::EMPTY_RESULT_TABLE_NAME )
+					)
+				);
+				$this->empty_result_table_ready = true;
+			}
 			return $pdo->query(
 				sprintf(
-					'INSERT INTO %s (rowid) SELECT NULL WHERE FALSE',
-					$this->quote_sqlite_identifier( self::GLOBAL_VARIABLES_TABLE_NAME )
+					'INSERT INTO %s ( x ) SELECT NULL WHERE FALSE',
```
adamziel: Same here, to match the temp table definition.

```suggestion
					'INSERT INTO %s ( placeholder_for_empty_insert ) SELECT NULL WHERE FALSE',
```

adamziel:
This looks pretty good, thank you! I'll take the liberty of applying my inline suggestions and merging this PR once the CI passes again.

PR #450, opened 2026-07-13
Add WordPress 7.0 compatibility

JanJakes:
## Summary
Mark the SQLite plugin as **tested up to WordPress 7.0**, and:
- Update the WordPress test environment from 6.7.2 to 7.0.1.
- Normalize non-padded temporal literals used by WordPress 6.8 and later.
- Remove WordPress test failures that now pass.
- Add exclusions for tests that now fail. These will be fixed in a follow-up PR.

## Why
Make sure the SQLite plugin is tested up to WordPress 7.0.

Normalize date literals generated by WordPress 6.8 and later. WordPress 6.8 changed `get_calendar()` to generate date bounds with integer month components, producing literals such as `2026-2-01`. MySQL interprets these values as dates, while SQLite compares temporal strings lexically. Normalizing the components preserves correct calendar queries.

## Verification

CI passing with tests updated to use WordPress 7.0.


mho22 (approved):
Looks good to me!

PR #452, opened 2026-07-15
Add ANSI_QUOTES and SQL mode validation

JanJakes:
## Summary

This PR improves SQL mode handling and adds **`ANSI_QUOTES`** and **`ANSII`** support. This includes:

- **SQL mode storage:** SQL modes use MySQL's native bitmap and canonical order.
- **Validation:** Named and numeric assignments are validated with MySQL-compatible errors.
- **ANSI_QUOTES:** `ANSI_QUOTES` is supported by the driver, standalone parser, and native lexer.
- **ANSI** Composite `ANSI` SQL mode is expanded into its component modes.
- **Quoted identifiers:** Fixed backslash behavior in quoted identifiers.
- **Native builds:** Lexer constants are regenerated, and the generator's monorepo path is fixed.

## Why
The `ANSI_QUOTES` SQL mode support keeps reappearing in our recent performance optimization and parser experiments. This PR extracts and completes that functionality.

Additionally, the driver stored arbitrary mode names, accepted invalid assignments, always treated double quotes as string delimiters, did not expand did not expand composite SQL modes, and applied string escaping to quoted identifiers. The standalone parser and native lexer lacked equivalent quoted-identifier handling.


packages/mysql-parser/tests/WP_MySQL_Token_Tests.php
```diff
@@ -38,6 +38,13 @@ public function test_get_value_unquotes_backtick_identifiers(): void {
 		$this->assertSame( 'col name', self::first_token( 'SELECT `col name` FROM t', 'BACK_TICK_QUOTED_ID' )->get_value() );
 	}
 
+	public function test_get_value_unquotes_ansi_identifiers(): void {
+		$this->assertSame(
+			'a"b',
+			self::first_token( 'SELECT "a""b"', 'BACK_TICK_QUOTED_ID', array( 'ANSI_QUOTES' ) )->get_value()
```
brandonpayton, line 44: How does this reduce `a""b` to `a"b`?  I think this is probably expected ANSI SQL behavior, but is there any standard we can link to somewhere that says this explicitly?

IIUC, the actual ANSI SQL standard is not publicly available, but is available for purchase. So I was trying to think what kind of doc we could refer to as an explanation.

Also, if this seems silly to you, please ignore it. I just had a question about the quote handling and thought it would be cool if we showed why it is what it is.

packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
```diff
@@ -3518,14 +3580,20 @@ private function execute_set_system_variable_statement(
...
 		}
 
 		if ( WP_MySQL_Lexer::SESSION_SYMBOL === $type ) {
 			if ( 'sql_mode' === $name ) {
-				$modes                  = explode( ',', strtoupper( $value ) );
-				$this->active_sql_modes = $modes;
+				if ( null !== $value_node->get_first_child_token( WP_MySQL_Lexer::DEFAULT_SYMBOL ) ) {
+					$sql_modes = $this->get_default_sql_modes();
+				} elseif ( is_string( $value ) ) {
+					$sql_modes = explode( ',', rtrim( $value, ' ' ) );
```
brandonpayton, line 3681: Naive question:
Why just rtrim() and not trim()?

packages/mysql-on-sqlite/tests/WP_MySQL_On_SQLite_Tests.php
```diff
@@ -2860,6 +2860,252 @@ public function testDefaultSqlModeDoesNotIncludeNoAutoValueOnZero() {
...
+
+		$this->assertQuery( 'SELECT @@sql_mode AS mode;' );
+		$this->assertSame(
+			'ONLY_FULL_GROUP_BY,STRICT_ALL_TABLES,NO_ENGINE_SUBSTITUTION',
+			$this->last_result[0]->mode
+		);
+	}
+
+	public function testSqlModesAcceptNumericBitmap() {
+		$this->assertQuery( 'SET sql_mode = 4294967299' );
```
brandonpayton: Nit: It might be nice to compose the number of specific bits so a human can better confirm how the expected mode corresponds to the assigned value. Same for other places mode numbers are used in these tests.

packages/mysql-on-sqlite/tests/WP_MySQL_On_SQLite_Tests.php
```diff
@@ -2860,6 +2860,252 @@ public function testDefaultSqlModeDoesNotIncludeNoAutoValueOnZero() {
...
+			array(
+				'pdo'           => $this->sqlite,
+				'mysql_version' => 50744,
+			)
+		);
+
+		$this->assertQuery( 'SET sql_mode = 16' );
+		$this->assertQuery( 'SELECT @@sql_mode AS mode;' );
+		$this->assertSame( ',', $this->last_result[0]->mode );
+		$this->assertFalse( $this->engine->is_sql_mode_active( 'NOT_USED' ) );
```
brandonpayton: What does this do?

I think this sets the mode to:
```
		'ONLY_FULL_GROUP_BY'         => 1 << 5,
```

Why would `$this->last_result[0]->mode` equal `','`?

packages/mysql-on-sqlite/tests/WP_MySQL_On_SQLite_Tests.php
```diff
@@ -2860,6 +2860,252 @@ public function testDefaultSqlModeDoesNotIncludeNoAutoValueOnZero() {
...
+		$this->assertSame( 'ANSI_QUOTES', $this->last_result[0]->mode );
+	}
+
+	public function testSqlModeDefaultRestoresDefaultBitmap() {
+		$this->assertQuery( "SET sql_mode = ''" );
+		$this->assertQuery( 'SET sql_mode = DEFAULT' );
+
+		$this->assertQuery( 'SELECT @@sql_mode AS mode;' );
+		$this->assertSame(
+			'ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION',
```
brandonpayton, line 3027: Is there a reasonable to way to assert against the actual default bitmap rather than a copy? Is there a reason it would not be a good idea?

packages/mysql-on-sqlite/tests/WP_MySQL_On_SQLite_Tests.php
```diff
@@ -2924,6 +3170,77 @@ public function testNoAutoValueOnZeroSqlMode() {
...
+		$this->assertEquals( 'alpha', $results[0]->name );
+	}
+
+	public function testAnsiQuotesAllowsDoubleQuotedIdentifiersInDdl() {
+		$this->assertQuery( "SET sql_mode = 'ANSI_QUOTES'" );
+
+		// Identifiers may contain spaces and escape the double quote by doubling it.
+		$this->assertQuery( 'CREATE TABLE "my ""tbl""" ("my col" INTEGER);' );
+		$this->assertQuery( 'INSERT INTO "my ""tbl""" ("my col") VALUES (42);' );
+		$this->assertQuery( 'SELECT "my col" FROM "my ""tbl""";' );
```
brandonpayton: What identifier does `""tbl"""` represent if two double quotes are combined to make one double quote?

brandonpayton (approved):
Hi @JanJakes, I read and reviewed this and left some notes and questions. I'm not confident enough in how everything works to say one way or the other today. But everything looks reasonable to me. Thanks!

packages/mysql-parser/tests/WP_MySQL_Token_Tests.php
JanJakes: @brandonpayton I'm not sure if this is part of the ANSI standard (AI says it actually is), but quoted identifiers in MySQL escape the quote character by doubling it (backslashes don't work in identifiers). So if you want to express a table name like `myˋtable` in a query, you'd need to write `SELECT * FROM ˋmyˋˋtableˋ`.

With the `ANSI_QUOTES` mode enabled, double quotes behave like backticks, so for a table name like `my"table`, you'd write `SELECT * FROM "my""table"` (or `SELECT * FROM ˋmy"tableˋ`).

I guess [the MySQL docs](https://dev.mysql.com/doc/refman/8.4/en/identifiers.html) are something we could link, as they say:

> If the character to be included within the identifier is the same as that used to quote the identifier itself, then you need to double the character.

I'll review how this is documented in the code.

packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
JanJakes: This just mirrors MySQL behavior. For some reason, MySQL trims only the trailing space, so `'ANSI_QUOTES '` is interpreted as `'ANSI_QUOTES'`, but `' ANSI_QUOTES'` is invalid. Good question, though; I need to document this better.

packages/mysql-on-sqlite/tests/WP_MySQL_On_SQLite_Tests.php
JanJakes: Ah, that's an edge case that the agent covered in tests. `16` is `1 << 4`, which is `'NOT_USED'`. In MySQL 5.7, it's literally named `,`, under MySQL 8, it's named `NOT_USED`. I'll document it better in the test.

packages/mysql-on-sqlite/tests/WP_MySQL_On_SQLite_Tests.php
JanJakes: It represents the table name `my "tbl"` (quotes within identifiers are escaped by doubling them; see also https://github.com/WordPress/sqlite-database-integration/pull/452#discussion_r3715911422). I'll see if I can use a better identifier and/or document it better.

packages/mysql-on-sqlite/tests/WP_MySQL_On_SQLite_Tests.php
JanJakes: I think the explicit value is preferable here. The test is verifying MySQL default, including the canonical order. This ensures we don't accidentally change the default to something else.

brandonpayton:
Thanks for all the answers, @JanJakes!

packages/mysql-parser/tests/WP_MySQL_Token_Tests.php
JanJakes: Addressed in 66bc1048.

packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
JanJakes: Addressed in a6580efc.

packages/mysql-on-sqlite/tests/WP_MySQL_On_SQLite_Tests.php
JanJakes: Addressed in e3d36a6f.

packages/mysql-on-sqlite/tests/WP_MySQL_On_SQLite_Tests.php
JanJakes: Addressed in e3d36a6f.

packages/mysql-on-sqlite/tests/WP_MySQL_On_SQLite_Tests.php
JanJakes: Addressed in 66bc1048.

JanJakes:
Rebased to resolve conflicts and addressed all comments. I guess it's ready to go.

PR #462, opened 2026-07-28
Preserve aliases in UPDATE JOIN translation

### LINKED ISSUE #461: UPDATE JOIN translation can update every target row
## Problem

The MySQL-on-SQLite translation of a single-target `UPDATE ... JOIN (derived table)` can lose correlation with the update target and update every row. This corrupts Action Scheduler claims by assigning completed, canceled, and unrelated pending actions to a claim that requested only a bounded pending subset.

The affected translation path is `execute_update_statement()`. It emits the target table without its MySQL alias, then includes that same table alias as an independent `FROM` item. The join predicate correlates the independent `FROM` rows with the derived table, not the actual update target.

## Minimal reproduction

Create four rows: two `pending`, one `complete`, and one `canceled`. Then execute Action Scheduler's claim-query shape with `LIMIT 1`:

```sql
UPDATE actions t1
JOIN (
  SELECT action_id
  FROM actions
  WHERE claim_id = 0
    AND status = 'pending'
  ORDER BY action_id ASC
  LIMIT 1
  FOR UPDATE
) t2 ON t1.action_id = t2.action_id
SET claim_id = 99;
```

Expected: one pending row is updated.

Observed with the 3.0.0-rc.3 driver: all four rows are updated and `rowCount()` returns 4.

The emitted SQLite statement has this effective shape:

```sql
UPDATE actions
SET claim_id = 99
FROM actions AS t1, (SELECT ... LIMIT 1) AS t2
WHERE t1.action_id = t2.action_id;
```

Because the update target is not aliased/correlated, the `WHERE` clause only proves that one `t1`/`t2` pair exists and therefore updates every target row.

## Production impact

On a WordPress Studio site using Action Scheduler, one claim was assigned to all 22,636 actions: 21,128 complete, 1,443 canceled, 5 failed, 2 in-progress, and only 58 pending. Queue runners revisited every terminal action and generated 13.2 million logs in nine days, 99.5% of them `action ignored via ...`.

## Expected fix

- Preserve the MySQL update-target alias when translating a single-target `UPDATE ... JOIN`.
- Do not repeat the target table as an independent `FROM` source.
- Add a regression test using this exact derived-table/`LIMIT` query shape.
- Assert both affected-row count and unchanged terminal rows.
- Preserve atomic bounded claim semantics.

## AI assistance

OpenAI `openai/gpt-5.6-sol` via OpenCode traced production Action Scheduler data, reduced the failure to an isolated in-memory driver reproduction, inspected the emitted SQLite query, and drafted this report. Chris Huber reviewed and owns the result.

chubes4:
## Summary
- resolve unqualified UPDATE targets back to their unique source alias
- preserve the target alias and exclude it from SQLite FROM clauses
- cover the Action Scheduler derived-table claim query from #461

## Testing
- composer run test -- --filter WP_MySQL_On_SQLite_Translation_Tests
- composer run test
- composer run check-cs

Fixes #461

## AI assistance
OpenAI model `openai/gpt-5.6-sol` via OpenCode analyzed the generic translator path, implemented the alias-resolution, and drafted the focused PDO regression coverage. Chris Huber reviews and owns the change.

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
@@ -2147,15 +2147,21 @@ class WP_MySQL_On_SQLite extends PDO {
 						)->fetchAll( PDO::FETCH_COLUMN );
 					}
 
-					$matched_tables          = array_merge( $matched_temporary_tables, $matched_persistent_tables );
-					$matched_aliases         = array_keys(
-						array_filter(
-							$table_alias_map,
-							function ( $data ) use ( $matched_tables ) {
-								return in_array( $data['table_name'], $matched_tables, true );
+					$matched_tables  = array_merge( $matched_temporary_tables, $matched_persistent_tables );
+					$matched_aliases = array();
+					foreach ( $table_alias_map as $alias => $data ) {
+						// Derived tables do not have a table name.
+						if ( null === $data['table_name'] ) {
+							continue;
+						}
+
+						foreach ( $matched_tables as $matched_table ) {
+							if ( 0 === strcasecmp( $data['table_name'], $matched_table ) ) {
+								$matched_aliases[] = $alias;
+								break;
 							}
-						)
-					);
+						}
+					}
 					$updates_multiple_tables = count( $matched_aliases ) > 1;
 					if ( 1 === count( $matched_aliases ) ) {
 						$table_or_alias = $matched_aliases[0];
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
@@ -2148,9 +2148,17 @@ class WP_MySQL_On_SQLite extends PDO {
 					}
 
 					$matched_tables          = array_merge( $matched_temporary_tables, $matched_persistent_tables );
-					$updates_multiple_tables = count( $matched_tables ) > 1;
-					if ( 1 === count( $matched_tables ) ) {
-						$table_or_alias = $matched_tables[0];
+					$matched_aliases         = array_keys(
+						array_filter(
+							$table_alias_map,
+							function ( $data ) use ( $matched_tables ) {
+								return in_array( $data['table_name'], $matched_tables, true );
+							}
+						)
+					);
+					$updates_multiple_tables = count( $matched_aliases ) > 1;
+					if ( 1 === count( $matched_aliases ) ) {
+						$table_or_alias = $matched_aliases[0];
 					} else {
 						break;
 					}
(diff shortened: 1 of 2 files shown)


JanJakes:
Hi @chubes4! Thanks for the PR. Before I dive deeper, I have one question. Is there a real-world use case for implementing `UPDATE ... FROM` for the legacy SQLite versions? That is, SQLite < 3.33.0?

The plugin only supports these versions when `WP_SQLITE_UNSAFE_ENABLE_UNSUPPORTED_VERSIONS` is set. I'd like to know if the 2nd and 3rd commits were perhaps added just because the tests failed after the first commit. In that case, the simpler fix of that test would be:

```php
$sqlite_version = $this->engine->get_sqlite_version();
if ( version_compare( $sqlite_version, '3.33.0', '<' ) ) {
	$this->markTestSkipped(
		sprintf( "SQLite version %s doesn't support UPDATE with FROM clause.", $sqlite_version )
	);
	return;
}
```
And then we could drop commits 2 and 3.

If we have a need for supporting `UPDATE ... FROM` on legacy SQLite versions in the `WP_SQLITE_UNSAFE_ENABLE_UNSUPPORTED_VERSIONS` mode, we can consider that, but otherwise I'd say the complexity is not worth it until we see a use case. Thanks!

chubes4:
There is no real world use case for that. It was definitely added by Codex after the failing tests. I should have noticed that was slop. I'll drop those commits. Thank you for calling that out!

packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
```diff
@@ -2217,9 +2148,17 @@ private function execute_update_statement( WP_Parser_Node $node ): void {
...
-					if ( 1 === count( $matched_tables ) ) {
-						$table_or_alias = $matched_tables[0];
+					$matched_aliases         = array_keys(
+						array_filter(
+							$table_alias_map,
+							function ( $data ) use ( $matched_tables ) {
+								return in_array( $data['table_name'], $matched_tables, true );
+							}
+						)
+					);
```
JanJakes: The alias lookup compares table names case-sensitively. To fix that, Codex suggests using something like this:

```php
// Resolve matched table names back to their query aliases.
$matched_aliases = array();
foreach ( $table_alias_map as $alias => $data ) {
	// Derived tables do not have a table name.
	if ( null === $data['table_name'] ) {
		continue;
	}

	foreach ( $matched_tables as $matched_table ) {
		if ( 0 === strcasecmp( $data['table_name'], $matched_table ) ) {
			$matched_aliases[] = $alias;
			break;
		}
	}
}
```

Let's also add a test for that. Thanks!

JanJakes (commented):
Thanks for the PR! I found one small issue that's probably worth addressing. Otherwise LGTM.

packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
chubes4: Agreed. The information-schema lookup can return the stored table spelling (`actions`) while the table-reference map preserves the query spelling (`Actions`), so the strict comparison could fail to map the physical table back to its alias. I updated the lookup to skip derived tables and compare physical table names with `strcasecmp()`.

I also strengthened the existing issue #461 regression by creating lowercase `actions` but issuing `UPDATE Actions t1`. That keeps the original alias-preservation coverage while proving case-insensitive resolution without duplicating the fixture. The focused test, all 747 driver tests, and PHPCS pass locally. Thanks for catching this!

JanJakes (approved):
Thanks!

PR #464, opened 2026-07-29
Restrict SQLite installation permissions

JanJakes:
## Summary

The SQLite integration is enabled by creating `wp-content/db.php`, and plugin deactivation removes that file. On **multisite**, the file is shared by the entire network, so the integration cannot meaningfully be configured per site. Site administrators must not be able to install or remove the network-shared file.

> [!NOTE]  
> The commits reflect the logical sequence of changes, so the PR is best reviewed by commit.

> [!IMPORTANT]
> Multisite installation still fails because enabling the SQLite drop-in switches WordPress to an empty database while multisite remains enabled, causing WordPress to query missing network tables before reaching the installer. This pre-existing issue can be addressed separately.

## Multisite changes
To address this, this PR implements the following **multisite** changes:

1. **Make the plugin network-only** on multisites.<br>This includes moving the plugin settings under network admin for multisite installs.
2. **Restrict installation and configuration** that manipulate `db.php` to network admins on multisites.<br>This means gating the plugin management by the `manage_network_options` capability.
3. **Restrict plugin deactivation** to network admins on multisites.<br>This is necessary, because deactivation removes the `db.php` file.
4. **Handle a special case for bulk deactivation.** <br>WordPress Core doesn't perform per-plugin capability checks in network admin. Therefore, we remove the plugin from the batch when the permissions are missing, and the plugin's deactivation hook refuses to delete `db.php`.

The permissions are designed as follows:

| Action | Capabilities |
|---|---|
| Activate,&nbsp;bulk&nbsp;activate | `manage_network_plugins` + `activate_plugins` |
| Enable/configure&nbsp;SQLite | **`manage_network_options`** |
| Deactivate | `manage_network_plugins` + `activate_plugins` + **`manage_network_options`** |
| Bulk&nbsp;deactivate | `manage_network_plugins` + `activate_plugins` + **`manage_network_options`**<br>Without these permissions, the SQLite plugin is removed from the batch, and the plugin's deactivation hook refuses to delete `db.php`. |

_*Added capabilities marked in **bold**._

## Single-site changes
On single-site installs, there is no risk of network-wide effects, but since the `db.php` drop-in significantly alters the site configuration, it makes sense to require `manage_options` capabilities, consistent with the multisite implementation.

This includes the following **single-site** changes:

| Action | Capabilities |
|---|---|
| Activate,&nbsp;bulk&nbsp;activate | `activate_plugins` |
| Enable/configure&nbsp;SQLite | **`manage_options`** |
| Deactivate,&nbsp;bulk&nbsp;deactivate | `activate_plugins` + **`manage_options`** |

_*Added capabilities marked in **bold**._

## Tests
This PR also adds a new single-site and multisite PHPUnit test suite and CI job for the authorization and installation behavior.

### DIFF
--- a/packages/plugin-sqlite-database-integration/activate.php
+++ b/packages/plugin-sqlite-database-integration/activate.php
@@ -14,10 +14,15 @@
  * @param string $plugin The plugin basename.
  */
 function sqlite_plugin_activation_redirect( $plugin ) {
-	if ( plugin_basename( SQLITE_MAIN_FILE ) === $plugin ) {
-		if ( wp_safe_redirect( sqlite_plugin_get_admin_page_url() ) ) {
-			exit;
-		}
+	if (
+		plugin_basename( SQLITE_MAIN_FILE ) !== $plugin
+		|| ! current_user_can( sqlite_plugin_get_manage_capability() )
+	) {
+		return;
+	}
+
+	if ( wp_safe_redirect( sqlite_plugin_get_admin_page_url() ) ) {
+		exit;
 	}
 }
 add_action( 'activated_plugin', 'sqlite_plugin_activation_redirect' );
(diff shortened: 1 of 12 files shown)


packages/plugin-sqlite-database-integration/activate.php
```diff
@@ -14,47 +14,63 @@
...
  * If the above conditions are met, run the sqlite_plugin_copy_db_file() function,
  * and redirect to the install screen.
  *
  * @since 1.0.0
  */
 function sqlite_activation() {
-	global $current_screen;
-	if ( isset( $current_screen->base ) && 'settings_page_sqlite-integration' === $current_screen->base ) {
+	if (
+		! isset( $_GET['page'], $_GET['confirm-install'] )
```
adamziel, line 41: This won't work in wp-cli, will it? Although, it probably already didn't before this change.

adamziel:
> Multisite installation still fails because enabling the SQLite drop-in switches WordPress to an empty database while multisite remains enabled, causing WordPress to query missing network tables before reaching the installer. This pre-existing issue can be addressed separately.

We're installing a better lock in a gate, but behind the gate there's still a cliff. How about we weld the gate until it's safe to go through it? In here, that would mean removing plugin activation and deactivation on multisites entirely. What do you think?

JanJakes:
> We're installing a better lock in a gate, but behind the gate there's still a cliff. How about we weld the gate until it's safe to go through it? In here, that would mean removing plugin activation and deactivation on multisites entirely. What do you think?

@adamziel We can do that, but considering the number of reports, we'd probably need to do it similarly as it is done in this PR. That is—add multisite checks in all the places, just never allow any of them to go through. So the skeleton would largely remain the same, I guess; we'd only add a better error message.

We're currently getting all these multisite reports even though multisite flow is not even working.

Alternatively, we could just straight refuse to boot the plugin very early if it's a multisite install, but this PR has some useful changes, so it might be better not to throw it away. WDYT?

PR #465, opened 2026-07-29
Honor NO_BACKSLASH_ESCAPES consistently

JanJakes:
## What changed

- Match mysqlnd string escaping in both default and `NO_BACKSLASH_ESCAPES` modes.
- Make `LIKE BINARY` pattern translation respect the active SQL mode.
- Preserve stored `DEFAULT` and `CHECK` expression values across SQL mode changes and table rebuilds.
- Quote reconstructed schema literals according to the active SQL mode.
- Expose `SERVER_STATUS_NO_BACKSLASH_ESCAPES` through the MySQL protocol so clients such as mysqli escape strings correctly.
- Implement mode-aware `PDO::quote()` behavior.

## Why

String parsing already understood `NO_BACKSLASH_ESCAPES`, but several quoting, reconstruction, pattern-matching, persistence, and protocol paths still assumed backslash escaping. That could change values during schema rebuilds, produce incorrect patterns or literals, and cause connected clients to escape strings for the wrong server mode.

This makes those paths consistent with MySQL/mysqlnd while retaining the existing default-mode behavior.

## Testing

- Full SQLite driver suite: 774 tests, 1,428,369 assertions
- Full MySQL proxy suite: 12 tests, 28 assertions
- Targeted WordPress escaping and SQL injection regression tests
- Coding standards
- PHP syntax checks
- PHP 7.2 compatibility checks for the added code
- Composer validation and diff whitespace checks

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
@@ -1193,10 +1193,21 @@ class WP_MySQL_On_SQLite extends PDO {
 	 * @return WP_MySQL_Parser        A parser initialized for the MySQL query.
 	 */
 	public function create_parser( string $query ): WP_MySQL_Parser {
+		return $this->create_parser_for_sql_modes( $query, $this->active_sql_modes );
+	}
+
+	/**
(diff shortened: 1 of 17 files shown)


JanJakes:
Closing this in favor of https://github.com/WordPress/sqlite-database-integration/pull/466 and deferring it until it's needed.

PR #485, opened 2026-08-11
Remove INET_ATON WordPress dependency

JanJakes:
## Summary

Removes the driver's dependency on WordPress's `absint()` helper in `INET_ATON()`. The replacement uses native PHP logic, allowing the Composer package to run in standalone PHP applications.

A regression test covers the function through the public PDO-compatible API. The package test suite and coding-standards check pass.

## Why

The standalone Composer package autoloads the driver outside WordPress. Calling `INET_ATON()` previously failed because `absint()` is not available in standalone PHP applications.


JanJakes:
This literally inlines [what WP does](https://github.com/WordPress/wordpress-develop/blob/83570840777b037c2df030f933443ca41c663450/src/wp-includes/load.php#L1469), so I'll go ahead and merge.

PR #487, opened 2026-08-12
Reject invalid BIT column defaults

JanJakes:
## Summary

Reject invalid `BIT` column defaults instead of preserving unnormalized input in the emulated information schema. Unsupported values now return MySQL-compatible error 1067.

## Why

The `BIT` default accepted any value beginning with a valid prefix and stored the original value. `SHOW CREATE TABLE` later emitted that value without quoting, allowing untrusted text to become executable SQL in a generated MySQL dump.

This is a hotfix that requires valid literal forms and rejects anything that cannot be safely normalized. Broader default validation based on the AST and full column definition should be done as a follow-up, also for other types.

This prevents untrusted SQL from being embedded in generated MySQL DDL during migration.


JanJakes:
Pushed the `DEFAULT ''` fix and going ahead and merging this for the release.

PR #493, opened 2026-08-21

JanJakes:
## Summary

## Managed storage

```text
wp-content/
└── database/
    ├── .htaccess
    ├── index.php
    ├── .ht.sqlite.lock
    ├── db-path.php
        ├── .htaccess
        ├── index.php
        └── .ht.sqlite
```

The `db-path.php` file returns the database location using `__DIR__`, so copying or moving the complete `database` directory keeps the reference valid. The storage manager also restores missing directories, protection files, and the database file when needed.

## Legacy migration

## Why

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,7 +66,7 @@
 		],
 		"wp-test-php": [
 			"@wp-test-ensure-env @no_additional_args",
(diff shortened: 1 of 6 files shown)


JanJakes:
Superseded by #494 so the pull request uses the database-access branch in this repository.

PR #494, opened 2026-08-21

JanJakes:
## Summary

## Managed storage

```text
wp-content/
└── database/
    ├── .htaccess
    ├── index.php
    ├── .ht.sqlite.lock
    ├── db-path.php
        ├── .htaccess
        ├── index.php
        └── .ht.sqlite
```

The `db-path.php` file returns the database location using `__DIR__`, so copying or moving the complete `database` directory keeps the reference valid. The storage manager also restores missing directories, protection files, and the database file when needed.

## Legacy migration

## Why

## Summary by CodeRabbit

* **New Features**
  * Added managed SQLite database storage with randomized paths when no database file is specified.
  * Added support for explicit database paths and in-memory databases.
  * Improved database initialization with protected directories, restricted permissions, locking, and safe recovery.

* **Bug Fixes**
  * Improved migration of existing SQLite databases and handling of interrupted initialization.
  * Enhanced validation and error handling while avoiding exposure of sensitive filesystem details.

* **Tests**
  * Added comprehensive coverage for storage creation, reuse, migration, permissions, locking, and failure scenarios.

### DIFF
--- a/packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
+++ b/packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
@@ -1,0 +1,260 @@
+<?php
+
+/**
+ * Manages the storage layout for the WordPress SQLite database.
+ *
+ * Filesystem warnings are suppressed to avoid exposing database paths:
+ * phpcs:disable WordPress.PHP.NoSilencedErrors.Discouraged
+ */
+class WP_SQLite_Storage {
+	/**
+	 * Filename of a script that returns the path to the database file.
+	 */
+	private const DATABASE_PATH_FILENAME = 'db-path.php';
+
+	/**
+	 * Filename of the database file.
+	 */
+	private const DATABASE_FILENAME = '.ht.sqlite';
+
+	/**
+	 * Filename of a lock file used for database storage setup.
+	 */
+	private const LOCK_FILENAME = '.ht.sqlite.lock';
+
+	/**
+	 * Number of bytes of a random token used in generated database paths.
+	 */
+	private const RANDOM_TOKEN_BYTE_LENGTH = 16;
+
+	/**
+	 * Managed database root directory.
+	 *
+	 * @var string
+	 */
+	private $database_root;
+
+	/**
+	 * Create a SQLite storage manager.
+	 *
+	 * @param string|null $database_root Managed database root. Defaults to FQDBDIR.
+	 */
+	public function __construct( ?string $database_root = null ) {
+		$this->database_root = trailingslashit( $database_root ?? FQDBDIR );
+	}
+
+	/**
+	 * Initialize the SQLite database storage.
+	 *
+	 * Uses an explicit file path or ":memory:" as provided. Otherwise, initializes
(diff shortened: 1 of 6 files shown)


adamziel:
The idea looks great. I don't have a good understanding why is the `.ht.sqlite.lock` file outside of the `.ht.<hash>` directory, but that also doesn't matter too much. I won't be able to review this deeply before my afk.

JanJakes:
> I don't have a good understanding why is the .ht.sqlite.lock file outside of the .ht.<hash> directory

@adamziel The lock needs to be outside because it guards creating `db-path.php` and the hashed directory itself. That is, it locks even before the directory exists. It needs to be at a fixed, predictable path.

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
```diff
@@ -0,0 +1,325 @@
...
+		if ( @is_file( $db_path_file ) ) {
+			$database_path = $this->read_database_path( $db_path_file );
+			if ( @is_file( $database_path ) ) {
+				return $database_path;
+			}
+		}
+
+		// Initialize or repair the managed database under a lock.
+		$this->ensure_protected_directory( $this->database_root );
+		$lock = $this->acquire_lock();
```
adamziel, line 91: The concurrent requests will fail, I'm not sure about the consequences of that. At best, the user will see some missing assets or failed XHR. At worst, it will paralyze the automated wp-cron plugin update job. Can we make this more lenient? If we can't acquire the lock, then maybe we can wait a second and see if the concurrent handler created the database file already?

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
```diff
@@ -0,0 +1,325 @@
...
+	/**
+	 * Ensure that a database and its protected directory exist.
+	 *
+	 * @param string $database_path Absolute database path.
+	 */
+	private function ensure_database( string $database_path ) {
+		$this->ensure_protected_directory( dirname( $database_path ) );
+
+		if ( ! @is_file( $database_path ) ) {
+			// Create an empty database file with restricted permissions.
```
adamziel, line 144: fun, SQLite3 can actually open an empty file as a database. TIL!

adamziel (approved):
So many good ideas in here @JanJakes! I left a note about error handling and, other than that, this seems ripe for shipping. Thank you! 🚢

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
JanJakes: Good point. When locking works, `flock` will block and wait as needed. But if we can't acquire a lock, we should handle it more gracefully indeed. I'll look into it.

JanJakes:
@adamziel I addressed the locking issue in 78e7e17d by handling `flock()` failures as you've suggested, and added safer database path publishing in 7c48da6e.

JanJakes:
Closing in favor of https://github.com/WordPress/sqlite-database-integration/pull/502.

PR #496, opened 2026-08-23
Align savepoint handling with MySQL semantics

### LINKED ISSUE #495: Write inside SAVEPOINT attempts nested BEGIN IMMEDIATE
## Summary

A translated write executed after a MySQL-compatible `SAVEPOINT` is active attempts to open the adapter's wrapper transaction with `BEGIN IMMEDIATE`. SQLite rejects that with `cannot start a transaction within a transaction`, so `wpdb->update()` returns `false` even though the savepoint itself was accepted.

## Exact behavior

The failing sequence is:

```sql
SAVEPOINT datamachine_transaction_1;
UPDATE wp_datamachine_jobs
SET status = 'processing', engine_data = '...'
WHERE job_id = 1 AND status = 'pending';
```

The update result observed immediately at the caller boundary is:

```json
{
  "return": false,
  "return_type": "boolean",
  "rows_affected": 0,
  "last_error_has_nested_transaction": true
}
```

The adapter trace for the `UPDATE` contains only:

```text
Executing: BEGIN IMMEDIATE | (no parameters)
SQLSTATE[HY000]: General error: 1 cannot start a transaction within a transaction
```

The translated `UPDATE` itself is therefore never executed.

## Root behavior

`WP_PDO_MySQL_On_SQLite::query()` wraps translated writes in its own transaction. A preceding SQL `SAVEPOINT` has already opened an SQLite transaction, but that state is not preventing the write wrapper from issuing `BEGIN IMMEDIATE`. This may be because the transaction was opened through SQL rather than `PDO::beginTransaction()`, so PDO transaction state alone cannot identify it.

## Reproduction evidence

This was reproduced through the exact deterministic Extra-Chill/data-machine#2578 agent bundle pinned from `chubes4/wp-site-generator` revision `b2669b1e59681e58997f2bf6bacfd18eb8bf230d`, with deterministic OpenAI and GitHub doubles and no live external mutation.

- WordPress: `6.9.2`
- PHP: `8.3.32`
- WP Codebox CLI/runtime packages: `0.21.0`
- Playground CLI: `3.1.46`
- retained runtime: `runtime-mt6bqbcc-dfymmz`
- retained artifact bundle: `artifact-bundle-sha256-38b3a05a217740e7b6801de40989b68ad518c5fa789c97c5a683ea9ae0df5161`
- artifact path: `/mnt/extrachill-workspace/tmp/opencode/verify-2578-after-3346-filesystem-trace-wp692/runtime-mt6bqbcc-dfymmz`

The runtime proves the surrounding effect: one Action Scheduler action starts and completes, but the compare-and-swap update fails, the job remains `pending`, and the downstream deterministic handler runs zero times.

## Expected

A write inside an active SQLite savepoint should execute under that transaction without attempting another top-level `BEGIN IMMEDIATE`. The adapter should preserve the write's real affected-row result and allow the caller to release or roll back its savepoint.

Refs Extra-Chill/data-machine#3348
Refs Extra-Chill/data-machine#2578

chubes4:
## What?

Align SQL savepoint handling with MySQL semantics.

A standalone `SAVEPOINT` in the default autocommit mode no longer opens an SQLite transaction. It is discarded immediately, so following writes use the normal `BEGIN IMMEDIATE` wrapper and commit independently.

Inside explicit transactions, track active savepoint names to emulate MySQL behavior for nested savepoints, reused and case-insensitive names, `ROLLBACK TO`, `RELEASE`, and missing-savepoint error 1305. Missing-savepoint errors leave the surrounding transaction active.

Add regression coverage across the MySQL-on-SQLite, concurrency, PDO API, and WordPress integration test suites.

## Why?

SQLite treats a standalone `SAVEPOINT` as opening a transaction, but MySQL does not. Passing it through to SQLite created a hidden transaction on PHP versions before 8.4. Subsequent writes then either attempted a nested `BEGIN IMMEDIATE` or bypassed the normal write-locking path, depending on how the transaction state was tracked.

The original Data Machine reproduction relied on SQLite-only behavior: issuing a bare `SAVEPOINT`, performing a write, and committing with `RELEASE SAVEPOINT`. Real MySQL and MariaDB discard the standalone savepoint, so the later release reports error 1305. That downstream issue is tracked in Extra-Chill/data-machine#3436.

> [!NOTE]
> Consumers that relied on a bare `SAVEPOINT` opening a transaction on SQLite must open an explicit transaction instead, matching MySQL behavior.

Fixes #495

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
@@ -712,18 +712,18 @@ class WP_MySQL_On_SQLite extends PDO {
 	private $in_transaction = false;
 
 	/**
-	 * User savepoints active in the current transaction, from outermost to innermost.
+	 * User savepoints in a transaction opened by a SAVEPOINT statement.
+	 *
+	 * On PHP < 8.4, PDO SQLite cannot detect transactions opened with raw SQL.
+	 * Tracking the savepoint stack keeps the inTransaction() polyfill accurate
+	 * when the outermost savepoint is released.
+	 *
+	 * Savepoints inside a transaction opened by BEGIN are not tracked because
+	 * releasing them cannot end the outer transaction.
 	 *
 	 * @var string[]
 	 */
-	private $transaction_savepoints = array();
-
-	/**
-	 * Whether the current transaction was opened by the outermost user savepoint.
-	 *
-	 * @var bool
-	 */
-	private $transaction_started_by_savepoint = false;
+	private $savepoint_transaction_stack = array();
 
 	/**
 	 * Whether a MySQL table lock is active.
@@ -2106,9 +2106,8 @@ class WP_MySQL_On_SQLite extends PDO {
 		 * @see self::begin_wrapper_transaction()
 		 */
 		$this->connection->query( 'BEGIN IMMEDIATE' );
-		$this->in_transaction                   = true;
-		$this->transaction_savepoints           = array();
-		$this->transaction_started_by_savepoint = false;
+		$this->in_transaction              = true;
+		$this->savepoint_transaction_stack = array();
 	}
 
 	/**
@@ -2120,9 +2119,8 @@ class WP_MySQL_On_SQLite extends PDO {
 			return;
 		}
 		$this->connection->query( 'COMMIT' );
-		$this->in_transaction                   = false;
-		$this->transaction_savepoints           = array();
-		$this->transaction_started_by_savepoint = false;
+		$this->in_transaction              = false;
+		$this->savepoint_transaction_stack = array();
 	}
 
 	/**
@@ -2134,9 +2132,8 @@ class WP_MySQL_On_SQLite extends PDO {
 			return;
 		}
 		$this->connection->query( 'ROLLBACK' );
-		$this->in_transaction                   = false;
-		$this->transaction_savepoints           = array();
-		$this->transaction_started_by_savepoint = false;
+		$this->in_transaction              = false;
+		$this->savepoint_transaction_stack = array();
 	}
 
 	/**
@@ -2178,7 +2175,7 @@ class WP_MySQL_On_SQLite extends PDO {
 						$this->execute_sqlite_query( sprintf( 'ROLLBACK TO SAVEPOINT %s', $savepoint_name ) );
 						$savepoint_index = $this->find_transaction_savepoint_index( $savepoint_key );
 						if ( null !== $savepoint_index ) {
-							$this->transaction_savepoints = array_slice( $this->transaction_savepoints, 0, $savepoint_index + 1 );
+							$this->savepoint_transaction_stack = array_slice( $this->savepoint_transaction_stack, 0, $savepoint_index + 1 );
 						}
 					}
 					return;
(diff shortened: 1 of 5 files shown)


packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
```diff
@@ -711,6 +711,20 @@ class WP_MySQL_On_SQLite extends PDO {
...
+	 */
+	private $transaction_savepoints = array();
+
+	/**
+	 * Whether the current transaction was opened by the outermost user savepoint.
+	 *
+	 * @var bool
+	 */
+	private $transaction_started_by_savepoint = false;
+
```
JanJakes, line 724: Could we use a single property by tracking only savepoints in a transaction opened by `SAVEPOINT`? Savepoints inside a transaction opened by `BEGIN` do not need tracking because releasing them cannot end the outer transaction. This would remove the need for `$transaction_started_by_savepoint`.

Let’s also document why this state is needed. Something like:

```php
	/**
	 * User savepoints in a transaction opened by a SAVEPOINT statement.
	 *
	 * On PHP < 8.4, PDO SQLite cannot detect transactions opened with raw SQL.
	 * Tracking the savepoint stack keeps the inTransaction() polyfill accurate
	 * when the outermost savepoint is released.
	 *
	 * Savepoints inside a transaction opened by BEGIN are not tracked because
	 * releasing them cannot end the outer transaction.
	 *
	 * @var string[]
	 */
	private $savepoint_transaction_stack = array();
```

JanJakes (commented):
@chubes4 Thanks for the PR! This looks good overall. I left one small simplification suggestion.

packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
chubes4: Done. Thank you!

JanJakes:
@chubes4 I had Fable review it thoroughly, and it caught that the implementation was diverging from MySQL behavior. With autocommit enabled (MySQL default), a standalone `SAVEPOINT` is discarded instead of opening a transaction. Tracking it as an active transaction also caused later writes to skip normal `BEGIN IMMEDIATE` locking, while a failed statement could roll back the entire savepoint transaction.

I pushed a follow-up commit to your branch that aligns savepoint handling with MySQL while preserving nested savepoints inside explicit transactions. If this resolves the issue you've encountered, I think we can proceed with this.

chubes4:
I will take a look and make sure this still resolves the problem. The error was thrown during headless testing on my VPS.

chubes4:
Thanks for taking this over — your approach is right, and I wanted to confirm the premise rather than assume it.

I verified against real MariaDB 10.11 with autocommit enabled:

```
SELECT @@autocommit, @@in_transaction;   -- 1, 0
SAVEPOINT sp1;
SELECT @@in_transaction;                 -- 0
```

A bare `SAVEPOINT` does not open a transaction, exactly as your commit message describes. And the follow-up matches your new error path:

```
SAVEPOINT sp1;
UPDATE ...;
RELEASE SAVEPOINT sp1;
-- ERROR 1305 (42000): SAVEPOINT sp1 does not exist
```

So skipping the SQLite statement and reporting 1305 reproduces MySQL faithfully. My original patch preserved behavior MySQL never had; yours is the correct fix.

One piece of context worth recording, since it validates the change. The reproduction behind #495 came from Data Machine, which issues a bare `SAVEPOINT` on SQLite without an enclosing transaction and then commits with `RELEASE SAVEPOINT`. Against this PR head that sequence now returns 1305 on release, and on the rollback path both `ROLLBACK TO SAVEPOINT` and `RELEASE SAVEPOINT` fail while the write stays committed.

That is a latent bug on our side, not a regression here — the same sequence already fails on real MySQL, and it only appeared to work because the old adapter let a bare savepoint open a transaction. I have filed it against Data Machine (Extra-Chill/data-machine#3436); the fix is to open a real transaction instead of relying on a bare savepoint.

Flagging it only so the behavior change is understood as intentional: consumers relying on a bare `SAVEPOINT` opening a transaction on SQLite will now see 1305, which is what MySQL does. No changes requested from me.

PR #500, opened 2026-08-31
Preserve index prefix lengths and order in SHOW CREATE TABLE primary keys

### LINKED ISSUE #501: SHOW CREATE TABLE drops index prefix lengths and DESC from primary keys
## Bug

`SHOW CREATE TABLE` loses index prefix lengths and descending order in the `PRIMARY KEY` clause. A table created with `PRIMARY KEY (session_id(100))` on a `MEDIUMTEXT` column is reported as `PRIMARY KEY (session_id)`, and `PRIMARY KEY (a DESC, b)` comes back as `PRIMARY KEY (a, b)`.

The information schema records the data correctly (`SUB_PART = 100`, `COLLATION = 'D'` in `_wp_sqlite_mysql_information_schema_statistics`); only the `PRIMARY KEY` branch of the statement reconstruction drops it, while the branch handling all other index types preserves both.

## Impact

Any consumer that rebuilds a schema from `SHOW CREATE TABLE` produces DDL that MySQL rejects: `ERROR 1170 (42000): BLOB/TEXT column used in key specification without a key length`. This is how the bug surfaces in WordPress Studio's push feature (`wp sqlite export` relays the statement as is), reported in Automattic/studio#4739.

## Reproduction

```sql
CREATE TABLE t (session_id MEDIUMTEXT NOT NULL, PRIMARY KEY (session_id(100)));
SHOW CREATE TABLE t;
-- reports: PRIMARY KEY (`session_id`)
-- expected, as in MySQL 8: PRIMARY KEY (`session_id`(100))
```

Verified against a real MySQL 8 server: its `SHOW CREATE TABLE` emits `PRIMARY KEY (`session_id`(100))` for the same table, and re-importing the driver's current output fails there with `ERROR 1170`.

samuelmbabhazi:
## Summary

`SHOW CREATE TABLE` reconstructs the `PRIMARY KEY` clause from the information schema, but unlike the branch that handles all other indexes, it emitted only the quoted column names. Index prefix lengths (`SUB_PART`) and descending order (`COLLATION = 'D'`) were dropped, so a table created with `PRIMARY KEY (session_id(100))` on a `MEDIUMTEXT` column came back as `PRIMARY KEY (session_id)`.

The practical impact: exports built on top of `SHOW CREATE TABLE` (for example `wp sqlite export`, which WordPress Studio uses for its push feature) produce a dump that MySQL rejects with `ERROR 1170 (42000): BLOB/TEXT column used in key specification without a key length`. The prefix is recorded correctly in `_wp_sqlite_mysql_information_schema_statistics` (`SUB_PART = 100`); the bug was emission only. Discovered through Automattic/studio#4739.

Fixes #501

## Fix

The column formatting closure that already handled `SUB_PART` and `DESC` for regular keys is hoisted and shared by the `PRIMARY KEY` branch, so both paths emit identical column definitions.

## Testing

- New regression test: a `PRIMARY KEY (session_id(100))` on `MEDIUMTEXT` round-trips through `SHOW CREATE TABLE` with its prefix, a composite key `PRIMARY KEY (a, b(50))` keeps the prefix only where defined, and `PRIMARY KEY (a DESC, b)` keeps its descending key part. The test fails on the previous code.
- Verified against a real MySQL 8 server: its `SHOW CREATE TABLE` emits `PRIMARY KEY (`session_id`(100))` and `PRIMARY KEY (`a` DESC,`b`)` for the same tables, matching the fixed output; the previous output fails to import there with `ERROR 1170`, the fixed output imports cleanly.
- Full `mysql-on-sqlite` unit suite: 875 tests, 1428750 assertions, no failures (the same 17 skipped and 2 incomplete as on trunk, PHP 8.5).
- `composer run check-cs` is clean on both changed files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected `SHOW CREATE TABLE` output to preserve primary-key index prefix lengths for single-column and composite keys.
  - Ensured descending key parts and column definitions are represented accurately in generated table-creation statements.
  - Improved consistency between primary-key and secondary-index formatting.

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-mysql-on-sqlite.php
@@ -7320,21 +7320,24 @@ class WP_MySQL_On_SQLite extends PDO {
 		}
 
 		// 7. Generate CREATE TABLE statement constraints, collect indexes.
+		$format_index_column = function ( $column ) {
+			$definition = $this->quote_mysql_identifier( $column['COLUMN_NAME'] );
+			if ( null !== $column['SUB_PART'] ) {
+				$definition .= sprintf( '(%d)', $column['SUB_PART'] );
+			}
+			if ( 'D' === $column['COLLATION'] ) {
+				$definition .= ' DESC';
+			}
+			return $definition;
+		};
+
 		foreach ( $grouped_constraints as $constraint ) {
 			ksort( $constraint );
 			$info = $constraint[1];
 
 			if ( 'PRIMARY' === $info['INDEX_NAME'] ) {
 				$sql  = '  PRIMARY KEY (';
-				$sql .= implode(
-					', ',
-					array_map(
-						function ( $column ) {
-							return $this->quote_mysql_identifier( $column['COLUMN_NAME'] );
-						},
-						$constraint
-					)
-				);
+				$sql .= implode( ', ', array_map( $format_index_column, $constraint ) );
 				$sql .= ')';
 			} else {
 				$is_unique = '0' === $info['NON_UNIQUE'];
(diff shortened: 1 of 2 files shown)


JanJakes:
Thanks for the fix, looks good!

PR #502, opened 2026-09-04

JanJakes:
## Summary

> [!NOTE]
> This is an alternative approach to #494. It keeps the same randomized storage design, but delegates coordination to SQLite's VFS and locking protocol through a dedicated SQLite database.

- **SQLite-backed locking:** Setup and maintenance are coordinated through a dedicated SQLite locking database.
- **Database isolation:** Maintenance also acquires an exclusive lock on an existing database before changing it.
- **Retryable setup:** Another request can complete setup after an interruption.

## Managed storage

```text
wp-content/
└── database/
    ├── .htaccess
    ├── index.php
    ├── .ht.sqlite.lock
    ├── db-path.php
        ├── .htaccess
        ├── index.php
        └── .ht.sqlite
```

The `db-path.php` file returns the database location using `__DIR__`, so copying or moving the complete `database` directory keeps the reference valid. If setup is interrupted after the path is published, another request can finish creating the protected directory and database.

## Storage locking
The storage implements a locking mechanism for initialization and maintenance. Locking uses a dedicated empty SQLite database and a maintenance file:

1. An exclusive lock is acquired on the locking database at ".ht.sqlite.lock".
2. A maintenance marker file is created at ".ht.sqlite.maintenance":
    - When requests see the file, they use the locking database.
    - Its absence provides a fast path for normal traffic.
3. When the lock is released, the maintenance file is automatically removed.

## Legacy migration

## Why

## Summary by CodeRabbit

* **New Features**
  * Added managed SQLite storage with randomized and in-memory database locations.
  * Added support for migrating existing SQLite databases.
  * Improved database protection and concurrent-access handling, including stale-lock recovery.

* **Bug Fixes**
  * Improved handling of database initialization failures with clearer failure responses.
  * Added safeguards for invalid database paths and storage-operation failures.
  * Improved cleanup of WordPress database files during test environment resets.

* **Tests**
  * Expanded coverage for initialization, migration, permissions, locking, concurrency, recovery, and in-memory databases.

### DIFF
--- a/packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
+++ b/packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
@@ -91,7 +91,14 @@ class WP_SQLite_Storage {
 	private $storage_lock_connection;
 
 	/**
-	 * Time to wait for the SQLite storage lock, in milliseconds.
+	 * Connection holding the exclusive transaction on the WordPress SQLite database.
+	 *
+	 * @var PDO|null
+	 */
+	private $database_lock_connection;
+
+	/**
+	 * Time to wait for existing database connections, in milliseconds.
 	 *
 	 * @var int
 	 */
@@ -133,9 +140,16 @@ class WP_SQLite_Storage {
 		// Never resolve a database path while storage maintenance is in progress.
 		$this->wait_until_unlocked();
 
+		// Check if we're already holding a lock.
+		$was_locked = null !== $this->storage_lock_connection;
+
 		// Explicitly configured database path.
 		if ( null !== $this->database_path ) {
 			$this->ensure_database( $this->database_path );
+			if ( $was_locked ) {
+				// If the storage was locked, ensure the database is locked as well.
+				$this->lock();
+			}
 			return $this->database_path;
 		}
 
@@ -156,7 +170,6 @@ class WP_SQLite_Storage {
 
 		// Initialize or repair the managed storage under the storage lock.
 		// Preserve a lock already held by this instance for a larger operation.
-		$keep_lock = null !== $this->storage_lock_connection;
 		$this->lock();
 		try {
 			// Another process may have completed the initialization meanwhile.
@@ -170,9 +183,10 @@ class WP_SQLite_Storage {
 			}
 
 			$this->ensure_database( $database_path );
+			$this->lock();
 			return $database_path;
 		} finally {
-			if ( ! $keep_lock ) {
+			if ( ! $was_locked ) {
 				$this->unlock();
 			}
 		}
@@ -184,27 +198,29 @@ class WP_SQLite_Storage {
 	 * The locking mechanism does the following:
 	 *   - Waits for a lock held by another process, up to the database lock timeout.
 	 *   - Marks maintenance as active so new requests wait.
+	 *   - When a database exists, also acquires an exclusive database lock.
+	 *
+	 * Existing database connections must be closed first to avoid blocking the lock.
 	 *
 	 * The lock is released when unlock() is called or the process ends.
 	 *
 	 * @throws RuntimeException When the lock cannot be acquired.
 	 */
 	public function lock(): void {
-		// Reuse a lock already held by this instance.
-		if ( null !== $this->storage_lock_connection ) {
-			return;
-		}
+		$was_locked = null !== $this->storage_lock_connection;
 
 		// Serialize maintenance through the dedicated locking database.
-		try {
-			$this->ensure_database( $this->lock_path );
-			$connection = new PDO( 'sqlite:' . $this->lock_path );
-			$connection->setAttribute( PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION );
-			$connection->exec( 'PRAGMA busy_timeout = ' . $this->database_lock_timeout );
-			$connection->exec( 'BEGIN EXCLUSIVE' );
-			$this->storage_lock_connection = $connection;
-		} catch ( Throwable $exception ) {
-			throw new RuntimeException( 'Failed to acquire the SQLite storage lock.', 0, $exception );
+		if ( ! $was_locked ) {
+			try {
+				$this->ensure_database( $this->lock_path );
+				$connection = new PDO( 'sqlite:' . $this->lock_path );
+				$connection->setAttribute( PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION );
+				$connection->exec( 'PRAGMA busy_timeout = ' . $this->database_lock_timeout );
+				$connection->exec( 'BEGIN EXCLUSIVE' );
+				$this->storage_lock_connection = $connection;
+			} catch ( Throwable $exception ) {
+				throw new RuntimeException( 'Failed to acquire the SQLite storage lock.', 0, $exception );
+			}
 		}
 
 		try {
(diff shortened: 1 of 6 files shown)


packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
```diff
@@ -0,0 +1,524 @@
...
+						// Connection lock successfully acquired.
+						return $connection;
+					}
+				}
+			} catch ( PDOException $exception ) {
+				$error_info    = $connection->errorInfo();
+				$sqlite_busy   = 5;
+				$database_busy = isset( $error_info[1] ) && ( (int) $error_info[1] & 0xff ) === $sqlite_busy;
+				if ( ! $database_busy ) {
+					throw $exception;
```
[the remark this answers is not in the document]
JanJakes: The PDO constructor is outside the try block, so a constructor failure never reaches this catch block. The original `PDOException` propagates unchanged.

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
```diff
@@ -0,0 +1,524 @@
...
+	 * @return PDO Connection holding the transaction.
+	 */
+	private function lock_database( string $database_path ): PDO {
+		// Do not silently create an empty database if the expected file is missing.
+		$pdo_options = array();
+		if ( defined( 'Pdo\Sqlite::ATTR_OPEN_FLAGS' ) ) {
+			$pdo_options[ Pdo\Sqlite::ATTR_OPEN_FLAGS ] = Pdo\Sqlite::OPEN_READWRITE;
+		} elseif ( defined( 'PDO::SQLITE_ATTR_OPEN_FLAGS' ) ) {
+			$pdo_options[ PDO::SQLITE_ATTR_OPEN_FLAGS ] = PDO::SQLITE_OPEN_READWRITE;
+		}
```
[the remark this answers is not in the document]
JanJakes: This is a valid edge case on PHP 7.2. There is no simple fix, and I think it's OK to accept this.

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
```diff
@@ -0,0 +1,524 @@
...
+	 * Ensure that a database and its protected directory exist.
+	 *
+	 * @param string $database_path Absolute database path.
+	 */
+	private function ensure_database( string $database_path ): void {
+		$this->ensure_protected_directory( dirname( $database_path ) );
+
+		if ( ! @is_file( $database_path ) ) {
+			// Create an empty database file with restricted permissions.
+			$database_handle = @fopen( $database_path, 'c' );
```
adamziel, line 414: This isn't atomic, right? Could we get an exclusive lock to avoid multiple requests getting here at the same time?

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
```diff
@@ -0,0 +1,524 @@
...
+		}
+
+		// This runs before wp_opcache_invalidate() is available.
+		$opcache_restrict_api = ini_get( 'opcache.restrict_api' );
+		$script_filename      = isset( $_SERVER['SCRIPT_FILENAME'] ) ? realpath( $_SERVER['SCRIPT_FILENAME'] ) : false;
+		if (
+			function_exists( 'opcache_invalidate' )
+			&& ( ! $opcache_restrict_api || ( $script_filename && 0 === stripos( $script_filename, $opcache_restrict_api ) ) )
+		) {
+			opcache_invalidate( $this->database_path_file, true );
```
adamziel, line 369: nice

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
adamziel: Well, actually it doesn't seem to matter here. `'c'` won't damage the file.

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
```diff
@@ -0,0 +1,524 @@
...
+			// Migrate legacy database paths.
+			$legacy_path = $this->database_root . self::DATABASE_FILENAME;
+			if ( ! @is_file( $legacy_path ) ) {
+				$legacy_path = $this->database_root . self::DATABASE_FILENAME . '.php';
+			}
+			if ( @is_file( $legacy_path ) ) {
+				$this->move_legacy_database( $legacy_path, $database_path );
+			} else {
+				$this->ensure_database( $database_path );
+				$this->lock();
```
adamziel, line 188: wait, aren't we already holding the lock? there's a `lock()` call a bit earlier in this method

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
JanJakes: @adamziel Right, 'c' won’t truncate an existing file. Multiple requests can reach this step, but the later `BEGIN EXCLUSIVE` call allows only one process to hold the storage lock at a time.

adamziel:
Releasing the database lock before `rename()` leaves a window where an older WordPress request can save a post into `.ht.sqlite-wal`. We then move only `.ht.sqlite`, leaving that saved change behind.

The maintenance marker does not protect us from requests already running the old plugin code.

I reproduced this by pausing between unlocking and renaming. Another process committed a row during that pause. Migration failed to lock the moved database, but the next initialization succeeded with the row missing. The old WAL file was still at the original location.

Can we keep the legacy database in place until old requests have finished and can no longer open that path? I don’t think we should automatically move it while this can leave successfully saved changes behind.

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
JanJakes: @adamziel This is only needed in a special case when the caller locked the storage before calling `initialize()` and expects it to remain locked afterwards. If the WordPress database doesn't exist yet, we'll only hold the storage lock (the special `.ht.sqlite.lock` database), but not the actual WordPress database lock.

This is the scenario:
```php
$storage->lock();       // Database doesn't exist yet; locks storage only.
$storage->initialize(); // Creates the database and must lock it too (the second lock() call).
// Continue maintenance with both locks held.
$storage->unlock();
```


I can make the second call conditional on `$was_locked` and add a comment explaining this edge case.

JanJakes:
> Releasing the database lock before `rename()` leaves a window where an older WordPress request can save a post into `.ht.sqlite-wal`. We then move only `.ht.sqlite`, leaving that saved change behind.

@adamziel Yes, in the migration path, there is a very narrow edge case where this could happen, but I'm not sure how likely it is in real-world scenarios. It seems to me there is no straightforward way to address it, because the connection must be closed so that the rename works in all environments.

> Can we keep the legacy database in place until old requests have finished and can no longer open that path? I don’t think we should automatically move it while this can leave successfully saved changes behind.

I’m not sure how we can establish that all old requests have finished and can no longer open the legacy path. Do you have a solution in mind? Alternatively, should we first release the randomized storage just for new databases (in 3.1) and only later handle the migration path (e.g., in 3.2)? That would at least handle the portion of users who update regularly.

adamziel:
> I’m not sure how we can establish that all old requests have finished and can no longer open the legacy path

Maybe a shared/exclusive `flock()`? In any case I think you right, maybe we don't need to address that.

packages/plugin-sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-storage.php
adamziel: Yeah, let's add a comment. Thank you!

adamziel (approved):

PR #505, opened 2026-09-08
Fix lexer edge cases and string handling

JanJakes:
## Summary

This PR fixes lexer edge cases and related string handling:

- **String values:** Preserve binary and single-byte charset payloads and correctly decode escaped newlines.
- **Function names:** Exclude trailing whitespace from function-name token ranges under `IGNORE_SPACE`.
- **Version comments:** Support six-digit versions from MySQL 8.1, preserve five-digit fallback parsing and unversioned numeric content, and recognize vertical tabs as whitespace.
- **WordPress validation:** Mirror the parent `wpdb::query()` charset checks, retaining its binary-data and prevalidated-query exemptions.
- **`LIKE BINARY` patterns:** Preserve raw bytes instead of producing `NULL` during GLOB conversion, and handle escaped newlines.

The lexer fixes cover both PHP packages and the native extension.

## Why

The token decoder's UTF-8 regex modifier caused an accidental `TypeError` for legitimate binary data as well as malformed text. Decoding must preserve bytes; charset enforcement needs column types and SQL mode. Mirroring WordPress's checks restores its expected validation behavior.

Regression tests cover decoding, token ranges, version comments, and pattern conversion. CI expectations remove the `test_strip_invalid_text` cases fixed by decoding and the two invalid-query tests fixed by WordPress validation.

## Remaining limitations

- Direct MySQL-on-SQLite writes can still accept invalid text under strict mode; full charset enforcement remains driver work. WordPress's checks retain their exemptions, including tables containing binary columns.
- SQLite GLOB can still treat distinct invalid UTF-8 sequences as equal. The `LIKE BINARY` fix preserves pattern bytes but does not provide fully byte-correct matching.
- Legacy multibyte charsets such as Big5 and GBK remain unsupported by the decoder.


packages/mysql-parser/tests/WP_MySQL_Server_Suite_Parser_Tests.php
```diff
@@ -34,7 +34,7 @@ public function test_corpus_acceptance_rate(): void {
 				continue;
 			}
 			++$total;
-			$tokens = ( new WP_MySQL_Lexer( $query ) )->remaining_tokens();
+			$tokens = ( new WP_MySQL_Lexer( $query, 80038 ) )->remaining_tokens();
```
mho22, line 39: The only thing I found worth mentionning, that could maybe be removed. Since [WP_MySQL_Lexer](https://github.com/WordPress/sqlite-database-integration/blob/bf181a286470be4de550112ce1f1d05322502c0e/packages/mysql-parser/src/class-wp-mysql-lexer.php#L330) seems to default to `80038` when it is not defined.

mho22 (approved):
Looks good to me!

packages/mysql-parser/tests/WP_MySQL_Server_Suite_Parser_Tests.php
JanJakes: It does currently default to 80038, but this test checks an exact set of expected failures that depends on the MySQL version. Passing it explicitly keeps the test stable if the lexer’s default changes. I’ve added a comment explaining why it is pinned.

mho22: Gotcha! 👍

mho22 (approved):

Issue #422, opened 2026-06-02
DB issue when using mailpoet


plus8:
hi guys/gals/thembies/furries, 

please excuse my ignorance, but I've been running sqlite dbs on my sites for a while now and it's been fantastic. Measurable performance improvements and overall top marks 😄 

however, recently I encountered an issue with mailpoet (mailer sending plugin) which seems to throw a sqlite error.

Initially I reported it to mailpoet staff who have taken note but pointed out this might be worth reporting here too, so here I am.

I'm not expecting anything, just reporting this in the hope that it's useful information to someone.

As of version 5.28 of mailpoet i have been seeing the following error at the top of the admin dashboard (below).

Hopefully it makes more sense to y'all than it does me.

Thank you for your work so far. The SQLite integration has been awesome and hopefully this helps in improving compatibility. 

best wishes,

Plus8 

----------------------------------------------------
Migration "MailPoet\Migrations\Db\Migration_20260504_120000_Db" failed. Details: An exception occurred while executing a query:
 
MySQL query:

SELECT COLUMN_TYPE
FROM INFORMATION_SCHEMA.COLUMNS
WHERE TABLE_SCHEMA = DATABASE()
AND TABLE_NAME = 'wp_mailpoet_subscribers'
AND COLUMN_NAME = 'source'

Queries made or created this session were:

    Executing: BEGIN | (no parameters)

    Executing: SELECT COLUMN_TYPE
    FROM INFORMATION_SCHEMA.COLUMNS
    WHERE TABLE_SCHEMA = DATABASE()
    AND TABLE_NAME = :param0
    AND COLUMN_NAME = :param1 | parameters: wp_mailpoet_subscribers, source

    Executing: ROLLBACK | (no parameters)

Error occurred at line 4229 in Function handle_error. Error message was: SQLSTATE[HY000]: General error: 1 no such table: INFORMATION_SCHEMA.COLUMNS.

Backtrace:

#0 /var/www/plus8/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php(549): WP_SQLite_Translator->get_error_message()

#1 /var/www/plus8/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php(455): WP_SQLite_DB->_do_query()
#2 /var/www/plus8/wp-content/plugins/mailpoet/lib/Doctrine/WPDB/Connection.php(115): WP_SQLite_DB->query()
[...]

PR #506, opened 2026-09-09
Fix schema reconstruction with native numeric results

JanJakes:
## Summary

Make **schema reconstruction independent of `PDO::ATTR_STRINGIFY_FETCHES`**. Interpret SQLite's primary-key, nullability, and uniqueness flags as integers so reconstruction preserves auto-increment columns, `NOT NULL`, defaults, and unique indexes in either fetch mode.

Add a regression test that creates an existing SQLite table before driver initialization and verifies its reconstructed column and index metadata with string fetching enabled and disabled.

## Why

Reconstruction compared SQLite's numeric flags strictly against strings. Since connection initialization stopped forcing string results in #291, reconstruction can receive integers before WordPress enables string fetching. This could mark every column in an auto-increment table as auto-incrementing, causing a later `ALTER TABLE` to fail with "has more than one primary key."

Related to #422.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when SQLite returns column and index metadata as either strings or integers.
  * Information schema reconstruction now consistently handles numeric metadata values regardless of fetch settings or PHP version.
  * Schema and index details are reconstructed more reliably across different SQLite metadata formats.

* **Tests**
  * Added coverage for schema reconstruction with stringified and native SQLite fetch values.
  * Verified reconstructed column metadata and index statistics across supported configurations.

### DIFF
--- a/packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-information-schema-reconstructor.php
+++ b/packages/mysql-on-sqlite/src/sqlite/class-wp-sqlite-information-schema-reconstructor.php
@@ -377,13 +377,13 @@ class WP_SQLite_Information_Schema_Reconstructor {
 		$definition[] = $mysql_type;
 
 		// NULL/NOT NULL.
-		if ( '1' === $column_info['notnull'] ) {
+		if ( 1 === (int) $column_info['notnull'] ) {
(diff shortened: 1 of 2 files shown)


JanJakes:
This is a straightforward fix. I'll go ahead and merge this one.

